PR #28, opened 2019-11-22
Support source, sink and function extension

ngjaying:


jinfahua:
1. The log sink is still not implement as ``sink interface``?
2. The source & sink configuration is not consistent... Do we need to make them consistent?

PR #63, opened 2019-12-27
Performance changes

ngjaying:


jinfahua:
1) Do we need to set the limit of dynamic buff size to avoid OOM issue? For example, expose an item in the configuration file - by default, the max buff size is 10M? 
2) Update the document for setting the concurrency value of operators for rules.

Issue #117, opened 2020-02-21
A generic function for getting data from metadata


jinfahua:
Need a generic function that can read data from record metadata.

PR #143, opened 2020-03-11
Support for generic meta function

ngjaying:
#117


jinfahua:
1) To add docs
2) Need to rewrite the mqtt function?

PR #275, opened 2020-05-20
fix float32/64 decoding from edgex bug

worldmaomao:


jinfahua:
Merged, thanks @worldmaomao

PR #282, opened 2020-05-26
Add support for http pull source

jinfahua:
Add support for http pull source


xstream/nodes/source_node.go
```diff
@@ -190,18 +192,25 @@ func (m *SourceNode) getConf(ctx api.StreamContext) map[string]interface{} {
 	conf, err := common.LoadConf(confPath)
 	props := make(map[string]interface{})
 	if err == nil {
-		cfg := make(map[string]map[string]interface{})
+		cfg := make(map[interface{}]interface{})
```
ngjaying, line 195: The original design just return the map[interface{}]interface{} prop value to the source. And the source config can do the conversion there

The root cfg seems always have a string key "default", do we need to change the map key type here?

xstream/nodes/source_node.go
```diff
@@ -190,18 +192,25 @@ func (m *SourceNode) getConf(ctx api.StreamContext) map[string]interface{} {
...
-			var ok bool
-			props, ok = cfg["default"]
+			def, ok := cfg["default"]
 			if !ok {
 				logger.Warnf("default conf is not found", confkey)
-			}
-			if c, ok := cfg[confkey]; ok {
-				for k, v := range c {
-					props[k] = v
+			} else {
```
ngjaying, line 202: common.convertMap(i) can convert map[interface{}]interface{} to map[string]interface{} recursively.

The conversion here only deal with one level

xstream/nodes/source_node.go
jinfahua: I changed the definition is that it has a bug when trying to load with hierarchy configurations. As in below, the `ClientId` under `optional` cannot be loaded.

```yaml
default:
  protocol: tcp
  server: localhost
  type: mqtt
  optional:
    ClientId: client1
    Username: user1
```

jinfahua:
@ngjaying Code updated, please review it, thanks.

PR #289, opened 2020-06-02
Support reserved words

ngjaying:


jinfahua:
@ngjaying How to handle below problem?

image

ngjaying:
> @ngjaying How to handle below problem?
> 
> image

Add backtick for the reserved words. meta(`Light-Dimming`->Device)

docs/en_US/sqls/query_language_elements.md
```diff
@@ -13,7 +13,12 @@ Kuiper provides a variety of elements for building queries. They are summarized
 | [ORDER BY](#ORDER BY) | Order the rows by values of one or more columns.             |
 | [HAVING](#HAVING)     | HAVING specifies a search condition for a group or an aggregate. HAVING can be used only with the SELECT expression.             |
 |                       |                                                              |
+### Use reserved words
+These elements and other predefined operators, functions are reserved words. To use the reserved words as the column name and the table name etc., they need to be quoted by backtick.
```
jinfahua, line 17: If the field name includes Chinese, or blank character, is it supported? If we support that, we should write the doc as well.

xsql/parser_test.go
```diff
@@ -26,7 +26,18 @@ func TestParser_ParseStatement(t *testing.T) {
 				Sources: []Source{&Table{Name: "tbl"}},
 			},
 		},
-
+		{
+			s: "SELECT `select` FROM tbl",
```
jinfahua, line 30: 1. Add parse error unit case - such as 'SELECT `select FROM tbl'
2. Add Chinese, and field name with blank character etc.

xsql/processors/xsql_processor_test.go
```diff
@@ -123,7 +138,8 @@ func createStreams(t *testing.T) {
 	}
 	demo1 := `CREATE STREAM demo1 (
 					temp FLOAT,
-					hum BIGINT,
+					hum BIGINT,` +
+		"`from`" + ` STRING,
```
jinfahua, line 142: Is it to add a field that is record type, and the record type also uses special key-words or characters. Such as,

- select STRUCT(`字段` STRING, `file name` BIGINT),

PR #293, opened 2020-06-04
Json path support

ngjaying:


docs/en_US/json_expr.md
```diff
@@ -146,6 +146,7 @@ So for example, when applied to the previous JSON data sample we can reach the f
 - `$.friends.first` refers to “dale”.
 - `$.friends` refers to the full array of friends.
 - `$.friends[0]` refers to the first friend listed in the previous array (contrary to arrays members are zero-based).
+- `$.friends[0][lastname]` refers to the lastname of the first friend listed. Use bracket if there are reserved words in the field key.
```
jinfahua, line 149: reserved words or special characters (such as  space ' ',  '.' and Chinese chars etc).

I will add a doc for our reserved keyword, and add a link here

PR #300, opened 2020-06-10
feat(server): Support https for rest service

ngjaying:


jinfahua:
@zhanghongtong Qiguai, we should also change the script and docs for adding SSL feature in Kuiper server, thanks.

PR #311, opened 2020-06-16
add plugin-sink with influxDB 

smart33690:
add a file influxDB.go


jinfahua:
@smart33690 
Thanks for contributing influxdb sink sample plugin! Could you please make following changes?
- Rebase commits in your PR.
- Update `docs/zh_CN/plugins/overview.md' & `docs/en_US/plugins/overview.md', and add your influxdb plugin.
- Add a doc named 'influxdb.md' in both `docs/zh_CN/plugins/sinks` & `docs/en_US/plugins/sinks`, which describes how to use your plugin.

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+	"encoding/json"
+	api "github.com/emqx/kuiper/xstream/api"
+	_ "github.com/influxdata/influxdb1-client/v2"
+	client "github.com/influxdata/influxdb1-client/v2"
+	"log"
+	"time"
+)
+
+type influxSink struct {
+	url       string
```
ngjaying, line 13: These properties are not used

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (m *influxSink) Open(ctx api.StreamContext) (err error) {
+	logger := ctx.GetLogger()
+	logger.Debug("Opening influx sink")
+	return
+}
+
+func (m *influxSink) Collect(ctx api.StreamContext, data interface{}) error {
+	cli := connInflux();
```
ngjaying, line 30: Why do we need to setup a new connection everytime? The collect function will be called for each message, is there a problem to reset connection each time like this? And it seems the cli(connection) does not close, will it be a problem? For other sink/source like mqttSink, we only hold a connection throughout the liftcycle. Can we do the same for influxdb?

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+	logger := ctx.GetLogger()
+	if v, ok := data.([]byte); ok {
+		var out ListMap
+		json.Unmarshal([]byte(v), &out)
```
ngjaying, line 34: The data is not always []map[string]float64. Other types need to be handled also. And the error of Unmarshall is not handled here.

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+		bp, err := client.NewBatchPoints(client.BatchPointsConfig{
+			Database:  "databasename",
```
ngjaying, line 36: It'd be better to export these properties as sink config properties instead of hard coded here.

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+			Precision: "ns", //default is ns
+		})
+		if err != nil {
+			logger.Debug(err)
```
ngjaying, line 40: should return err instead of just print it out otherwise bp will be nil in line 52

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		tags := map[string]string{"tagkey": "tagvalue"}
```
ngjaying, line 42: Are these "tagkey", "field1" etc. all hard coded? Can they be parameterize?

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+		fields := map[string]interface{}{
+			"filed1": out[0]["filed1"],
+			"filed2": out[0]["filed2"],
+		}
+
+		pt, err := client.NewPoint("measurement", tags, fields, time.Now())
```
ngjaying, line 48: Another hard coded string "measurement"

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+		if err != nil {
+			logger.Debug(err)
```
ngjaying, line 50: Again, error is not handled. The same applies for the next several error handlings.

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+		logger.Debug("insert success")
+	} else {
+		logger.Debug("insert faild")
+	}
+
+	return nil
+}
+
+func (m *influxSink) Close(ctx api.StreamContext) error {
+	return nil
```
ngjaying, line 66: The client is not closed

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+
+func connInflux() client.Client {
+	cli, err := client.NewHTTPClient(client.HTTPConfig{
+		Addr:     "addrIp",
```
ngjaying, line 72: Another hardcoded strings

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+		Username: "",
+		Password: "",
+	})
+	if err != nil {
+		log.Fatal(err)
```
ngjaying, line 77: Do not use Fatal, it will exit Kuiper instead of just this plugin.

plugins/sinks/influxdb.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	return cli
+}
+
+
+var Influx influxSink
```
ngjaying, line 83: Should export construction function here.
```
func Influx() api.Sink {
	return &influxSink{}
}
```

plugins/sinks/influxdb.go
smart33690: Thank you for your suggestion，I did some optimization of Influxdb.go 
Currently,only one type of data supported(Due to data format problems,is "[{}]" not "{}")
I will continue to optimize and solve this problem.

plugins/sinks/influxdb.go
smart33690: I see the example of the mysql sink like this
"var Mysql mysqlSink"

ngjaying: The tutorial needs to be modified. 
When exporting the struct, the plugin will run in singleton mode which means that all rules share one instance. On the contrary, export a "constructor" will have one instance for each rule. I think this sink is required to be one instance for each rule.

plugins/sinks/influxdb.go
ngjaying: ok. If other types are not supported currently, we'd better handle the error of unmarshall. For example, print out the error and return. Otherwise, it will crash the whole rule.

jinfahua:
@smart33690 Could you please use `gofmt` to format code, thanks.

Issue #313, opened 2020-06-16
The state of rule is not persistent


worldmaomao:
Steps:
1. Stop some rule by invoking rest api
2.Restart kuiper
3.After restart kuiper, the rule which was stopped before become to be running.

Expect：
After restart kuiper, the rule which was stopped before still be stopped.

jinfahua:
Done & close the issue.

PR #331, opened 2020-06-23
“fix:The state of rule is not persistent #313”

EMQmyd:


xstream/server/server/ruleManager.go
```diff
@@ -185,10 +186,27 @@ func stopRule(name string) (result string) {
...
+	rule, err := ruleProcessor.GetRuleByName(name)
+	if nil != err {
+		return fmt.Sprintf("%v", err)
+	}
+
+	err = startRule(name)
+	if nil != err {
+		return fmt.Sprintf("%v", err)
+	}
+	if 1 == rule.State {
```
ngjaying, line 208: Can we check the rule state before starting it? The stopped rule should not be started at all

xstream/server/server/rpc.go
```diff
@@ -116,7 +116,10 @@ func (t *Server) StopRule(name string, reply *string) error {
 	*reply = stopRule(name)
 	return nil
 }
-
+func (t *Server) RecoverRule(name string, reply *string) error {
```
ngjaying, line 119: This seems no effect. The caller can call the recoverRule directly. And this is not an exported RPC API. I suggest this function can be removed.

xstream/api/stream.go
```diff
@@ -79,6 +79,7 @@ type TopNode interface {
 }
 
 type Rule struct {
+	State   int                      `json:"state"` //0:started 1:stoped
```
ngjaying, line 82: Any reason not to use bool type here? How about name it started or stopped?

xsql/processors/xsql_processor.go
```diff
@@ -229,6 +229,33 @@ func (p *RuleProcessor) ExecCreate(name, ruleJson string) (*api.Rule, error) {
...
+	if err != nil {
+		return err
+	}
+	defer p.db.Close()
+
+	err = p.db.Replace(name, string(ruleJson))
+	if err != nil {
+		return err
+	} else {
+		log.Infof("Rule %s is replaceed.", name)
```
ngjaying, line 254: typo here, should be "replaced"

xsql/processors/xsql_processor.go
```diff
@@ -229,6 +229,33 @@ func (p *RuleProcessor) ExecCreate(name, ruleJson string) (*api.Rule, error) {
 	return rule, nil
 }
 
+func (p *RuleProcessor) ExecReplaceRuleState(name string, state int) (err error) {
+	rule, err := p.GetRuleByName(name)
+	if nil != err {
```
ngjaying, line 234: let's be consistent with the err check style  `err != nil`

xsql/processors/xsql_processor.go
```diff
@@ -229,6 +229,33 @@ func (p *RuleProcessor) ExecCreate(name, ruleJson string) (*api.Rule, error) {
...
+		return err
+	}
+
+	rule.State = state
+	ruleJson, err := json.Marshal(rule)
+	if nil != err {
```
ngjaying, line 240: The same as above

PR #339, opened 2020-06-24
Feat/edgex array

jinfahua:
EdgeX array support


xsql/ast.go
```diff
@@ -1164,14 +1164,21 @@ func (v *ValuerEval) evalBinaryExpr(expr *BinaryExpr) interface{} {
 	case error:
 		return val
 	}
-
+	if isSliceOrArray(lhs) {
```
ngjaying, line 1167: The case of []interface{} can be combined with isSliceOrArray. Remove the case and leave isSliceOrArray only

xsql/ast.go
```diff
@@ -1188,34 +1195,46 @@ func (v *ValuerEval) evalJsonExpr(result interface{}, op Token, expr Expr) inter
...
-				if berVal.isIndex() {
-					if berVal.Start >= val.Len() {
-						return fmt.Errorf("out of index: %d of %d", berVal.Start, val.Len())
-					}
-					return val.Index(berVal.Start).Interface()
-				} else {
-					if berVal.Start >= val.Len() {
-						return fmt.Errorf("start value is out of index: %d of %d", berVal.Start, val.Len())
-					}
+		return v.subset(result, op, expr)
```
ngjaying, line 1198: The same as above, remove the case []interface{}, []map[string]interface{}:

Issue #334, opened 2020-06-24
Return a JSON message for root REST API address


jinfahua:
Currently, the API only returns `OK`, more information with JSON format should be returned.

- Kuiper version
- Kuiper up time
- Operator system 

The impacted modules
- A unit test
- Helmchart?
- API documents
- Kuiper management UI

jinfahua:
@ysfscream Could you please update Kuiper management UI accordingly?

ysfscream:
@jinfahua OK, I will update it recently.

PR #340, opened 2020-06-24
feat:Return a JSON message for root REST API address #334

EMQmyd:


xstream/server/server/rest.go
```diff
@@ -97,12 +98,26 @@ func createRestServer(port int) *http.Server {
...
 	case http.MethodGet, http.MethodPost:
 		w.WriteHeader(http.StatusOK)
+		info := new(information)
+		info.Version = version
+		info.UpTimeSeconds = time.Now().Unix() - startTimeStamp
+		info.Os = runtime.GOOS
+		if byteInfo, err := json.Marshal(info); err == nil {
+			w.Write(byteInfo)
+			return
+		}
```
jinfahua, line 120: Print an error message in log if having any error. Or if it could not happen, just ignore the error check?

Issue #229, opened 2020-04-17
Failed to use the absolute path to start kuiper


zhanghongtong:
Kuiper can only be started with relative paths in the correct directory, not with absolute paths

zhanghongtong:
Chang path:
```
./bin -> /usr/lib/kuiper/bin
./etc -> /etc/kuiper
./log -> /var/log/kuiper
./data -> /var/lib/kuiper/data
./plugins -> /var/lib/kuiper/plugins
```

jinfahua:
It's fixed, and close the issue.

PR #343, opened 2020-06-29
feat:Failed to use the absolute path to start kuiper #229

EMQmyd:


common/util.go
```diff
@@ -159,8 +163,33 @@ func GetDataLoc() (string, error) {
 	return GetLoc(data_dir)
 }
 
+func getAbsPath(subdir string) (dir string, err error) {
+	switch true {
```
ngjaying, line 167: Seems like changing the previous semantic with contains.  "aetc" will now also return etc path, I don't think it is expected.

common/util.go
```diff
@@ -18,13 +18,17 @@ import (
...
-	data_dir      = "/data/"
-	log_dir       = "/log/"
-	StreamConf    = "kuiper.yaml"
-	KuiperBaseKey = "KuiperBaseKey"
-	MetaKey       = "__meta"
+	logFileName    = "stream.log"
+	etc_dir        = "/etc/"
+	data_dir       = "/data/"
+	log_dir        = "/log/"
+	abs_etc_dir    = "/etc/kuiper/"
```
ngjaying, line 25: How about using map for these constant abs path

common/util.go
```diff
@@ -159,8 +163,33 @@ func GetDataLoc() (string, error) {
...
+	}
+	if _, err = os.Stat(dir); os.IsExist(err) {
+		return dir, nil
+	}
+	return "", err
+}
+
 func GetLoc(subdir string) (string, error) {
-	dir, err := os.Getwd()
+	dir, err := getAbsPath(subdir)
```
ngjaying, line 188: Need to check with @zhanghongtong which is the highest priority: absolute or relative path

jinfahua:
2 tests are failed @EMQmyd , it seems related to your fix.

PR #348, opened 2020-07-01
MQTT sink dynamic configuration field "retained" & "qos"

smart33690:
**add field "retained" dynamic configuration**

**modify field "qos" dynamic configuration**


xstream/sinks/mqtt_sink.go
```diff
@@ -56,6 +58,19 @@ func (ms *MQTTSink) Configure(ps map[string]interface{}) error {
...
+			qos = v
+		} else if v, ok := qos.(string); ok {
+			if v == "1" {
+				qos = 1
+			} else if v == "2" {
+				qos = 2
+			}
+		}
+	}
+
```
jinfahua, line 73: 1. I think `qos` specified in rule should be `int` type, it should not be `string` or `byte`.
2. Check the value of `qos`, if it's not 0, 1 or 2, then it should return an error `not valid qos value %d`
3. Convert to `byte` type when setting `MQTTSink` struct instance.

xstream/sinks/mqtt_sink.go
```diff
@@ -95,15 +110,24 @@ func (ms *MQTTSink) Configure(ps map[string]interface{}) error {
 		}
 	}
 
+	retained := false
+	if pk, ok := ps["retained"]; ok {
```
jinfahua, line 114: Update the document `docs/en_US/rules/sinks/mqtt.md` and `docs/zh_CN/rules/sinks/mqtt.md`.

jinfahua:
@smart33690 Thanks for your PR, 2 comments... Could you please take a look?

xstream/sinks/mqtt_sink.go
smart33690: 1.This is c.Publish method “Publish(topic string, qos byte, retained bool, payload interface{}) Token”

smart33690: > 1. I think `qos` specified in rule should be `int` type, it should not be `string` or `byte`.
> 2. Check the value of `qos`, if it's not 0, 1 or 2, then it should return an error `not valid qos value %d`
> 3. Convert to `byte` type when setting `MQTTSink` struct instance.

1.This is c.Publish method “Publish(topic string, qos byte, retained bool, payload interface{}) Token”

jinfahua: Yes, but I don't think the type passed `Configure(ps map[string]interface{})` will be `byte`. It should be `int`.

jinfahua:
Thank you, merged the code.

PR #370, opened 2020-07-20
skip if info not end with .json

chensheng0:


jinfahua:
@chensheng0 Thanks for your PR :-)

PR #377, opened 2020-07-21
feat(main): gracefully shut down restful  rpc  prometheus services

GZJ:
Http server shutdown can close the service after all connections are executed.


ngjaying:
Merged, thanks for contributing.

PR #425, opened 2020-08-18
fix(sinks): fix metadata format

ysfscream:


jinfahua:
@EMQmyd Please follow the name convention as corrected by Shifan, thanks.

Issue #396, opened 2020-08-03
Filter for every nth element


dmocek:
Does kuiper support filtering for every nth element?  I don't see a built-in function for it.  I thought maybe I could select every nth element of a group, but I don't see support for that either.  Any way to do this without having to create a custom function?

jinfahua:
Do you mean function such as 95% percentile data? If it's true, now it's under planning. Hopefully it would be available at the end of Sep.

dmocek:
What I want is if I have these temperature readings coming into kuiper: 75, 79, 85, 86, 91, 92 and I want to filter for every 2nd reading, only the values 79, 86, and 92 will pass through.  If I want to filter for every 3rd reading, only the values 85, ad 92 will pass through.  If I want to filter for every 6th reading, only the value 92 will pass through.  So I want to be able to specify a number (n), and only the nth value will pass through the filter.

jinfahua:
Currently no such functions are supported, I will set the feature as enhancement.

Below could be a workaround, but it's cumbersome.
- Use `countwindow` to [trigger data](https://github.com/emqx/kuiper/blob/develop/docs/en_US/sqls/windows.md#count-window) with specified count interval.
- Use [data template](https://github.com/emqx/kuiper/blob/develop/docs/en_US/rules/data_template.md#data-traversal) in sink to filter your expected data.

ngjaying:
We added a new function `collect()` which will return an array of the elements of the current window. Thus, you can use index to access any of the elements in that array. For example, to get the temperature of every 3rd element, the sql will be like `SELECT Collect(*)[2]->temperature AS temp FROM demo GROUP BY COUNTWINDOW(3)`  or `SELECT Collect(temperature)[2] AS temp FROM demo GROUP BY COUNTWINDOW(3)`

dmocek:
Great!  Quick turn-around time.  Thanks.

PR #442, opened 2020-08-27
feat(func): add collect function

ngjaying:
#396


xsql/parser_test.go
```diff
@@ -1205,21 +1205,9 @@ func TestParser_ParseStatement(t *testing.T) {
...
-							Args: []Expr{
-								&NumberLiteral{Val: -0.3},
-							},
-						},
-						Name:  "sample",
-						AName: ""},
-				},
-				Sources: []Source{&Table{Name: "tbl"}},
-			},
+			s:    `SELECT sample(-.3,) FROM tbl`,
```
jinfahua, line 1208: Why this statement is returned with nil value?

jinfahua (commented):
A testcase problem...

xsql/parser_test.go
ngjaying: "sample" is not a valid function name. There is no function validation previously for functions with no arguments so that it passed the parser. The validation for no argument functions is added in this PR.

PR #443, opened 2020-08-27
feat: update the sink metada format

jinfahua:


plugins/sinks/taos.json
```diff
@@ -54,12 +54,12 @@
 		"control": "text",
 		"type": "string",
 		"hint": {
-			"en_US": "User of Taosdb",
-			"zh_CN": "Taosdb 用户名"
+			"en_US": "User name",
```
ysfscream, line 57: `Username` should be a complete word

plugins/sinks/taos.json
```diff
@@ -54,12 +54,12 @@
...
+			"zh_CN": "用户名"
 		},
 		"label": {
-			"en_US": "DB user",
-			"zh_CN": "数据库用户名"
+			"en_US": "User name",
```
ysfscream, line 61: `Username` should be a complete word

plugins/sinks/taos.json
jinfahua: Updated.

plugins/sinks/taos.json
jinfahua: Updated.

ysfscream (approved):

PR #469, opened 2020-09-01
feat():Delete alarm

EMQmyd:


plugins/manager.go
```diff
@@ -389,14 +383,7 @@ func (m *Manager) install(t PluginType, name string, src string) ([]string, stri
...
 		} else if fileName == name+".json" {
-			confDir, err := common.GetConfLoc()
-			if nil != err {
-				return filenames, "", err
-			}
-			err = unzipTo(file, path.Join(confDir, PluginTypes[t], fileName))
-			if err != nil {
-				return filenames, "", err
-			}
+			unzipTo(file, path.Join(m.etcDir, PluginTypes[t], fileName))
```
ngjaying, line 386: Do we need to handle the error here? Shall we fail the installation if error happens?

ngjaying (approved):

PR #486, opened 2020-09-03
feat():add api of sources yaml

EMQmyd:


plugins/sinkMeta.go
```diff
@@ -128,6 +128,7 @@ func newAbout(fi *fileAbout) *about {
 	}
 	ui := new(about)
 	ui.Trial = fi.Trial
+	ui.Installed = fi.Installed
```
jinfahua, line 131: Where do you initial the `Installed` variable?

PR #504, opened 2020-09-14
refactor: add debug info

jinfahua:


xstream/nodes/sink_node.go
```diff
@@ -159,18 +159,22 @@ func (m *SinkNode) Open(ctx api.StreamContext, result chan<- error) {
 				var sink api.Sink
 				var err error
 				if !m.isMock {
+					logger.Debugf(fmt.Sprintf("Trying to get sink for rule %s with options %v\n", ctx.GetRuleId(), m.options))
```
ngjaying, line 162: Debugf already allow format symbol. Do not need to nest fmt.Sprintf

ngjaying (changes requested):

xstream/nodes/sink_node.go
jinfahua: Updated

ngjaying (approved):

PR #508, opened 2020-09-15
Update conf.go

soyoo:
the last return will never reach.

### DIFF
--- a/common/http_util.go
+++ b/common/http_util.go
@@ -46,7 +46,7 @@ func Send(logger api.Logger, client *http.Client, bodyType string, method string
 			switch value.(type) {
 			case []interface{}, map[string]interface{}:
 				if temp, err := json.Marshal(value); err != nil {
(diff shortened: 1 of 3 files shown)


jinfahua:
Thanks for your patch, below 3 suggestions. 

- The unit testcase is failed. Could you please use `gofmt` to format the code before creating PR? It shows that the line of change has code format issue.
- Create PR against `dev/0.9.1` branch, not against `master` directly.
- It seems that just delete the last `return false` statement is better?

soyoo:
> Thanks for your patch, below 3 suggestions.
> 
> * The unit testcase is failed. Could you please use `gofmt` to format the code before creating PR? It shows that the line of change has code format issue.
> * Create PR against `dev/0.9.1` branch, not against `master` directly.
> * It seems that just delete the last `return false` statement is better?

hi, i had changed it.

PR #539, opened 2020-09-24
docs: update the edge-manager usage doc

jinfahua:
Updated the doc.


jinfahua:
@EMQmyd Please move screen-shot files into `resource` folder, and rename those file names of `en_US` directory according to the updates made in `zh_CN` folder.

PR #547, opened 2020-09-28
taos install

EMQmyd:


jinfahua:
@zhanghongtong Did you delete `.ci/Dockerfile-plugins-alpine` intentionally?

zhanghongtong:
> @zhanghongtong Did you delete `.ci/Dockerfile-plugins-alpine` intentionally?

Yes, I deleted it, because we don't need the alpine plugins

PR #566, opened 2020-10-20
chore(CI): change aws and docker account

zhanghongtong:


jinfahua:
It has problem with the Mac builds...

PR #580, opened 2020-10-30
feat():Change to a smaller time unit for elapsed time (rule status)

EMQmyd:


ngjaying (changes requested):
The tests are still needed

ngjaying (approved):

PR #581, opened 2020-10-30
feat(rule):support for update streams

EMQmyd:


xstream/server/server/rest.go
```diff
@@ -191,6 +191,20 @@ func streamHandler(w http.ResponseWriter, r *http.Request) {
 		}
 		w.WriteHeader(http.StatusOK)
 		w.Write([]byte(content))
+	case http.MethodPut:
+		streamProcessor.DropStream(name)
```
ngjaying, line 195: Suggest to put this after syntax check so that if the syntax is error, the stream won't be dropped.

ngjaying (changes requested):
One small thing to change

ngjaying (approved):

PR #586, opened 2020-11-03
Adaptation log: syslog

EMQmyd:


jinfahua:
Is is possible to read from a program environment variable which tell us to turn on the logrus_syslog. I suggest that the default behavior should not turn on logrus_syslog.

common/util.go
```diff
@@ -79,21 +82,19 @@ type KuiperConf struct {
 }
 
 func init() {
+	Log = logrus.New()
+	if "" != os.Getenv(KuiperSyslogKey) {
```
jinfahua, line 86: I suggest to let user specify a specified value as following (case insensitive),
`"true" == os.Getenv(KuiperSyslogKey)`

jinfahua (changes requested):

PR #610, opened 2020-11-19
feat(log):Logrotate support

EMQmyd:


jinfahua:
- Does it have an option for control the log file size? For example, if the log file size reaches 100MB, then create another log file.
- Update the related docs as well.

go.mod
```diff
@@ -14,9 +14,12 @@ require (
 	github.com/gorilla/handlers v1.4.2
 	github.com/gorilla/mux v1.7.3
 	github.com/keepeye/logrus-filename v0.0.0-20190711075016-ce01a4391dd1
+	github.com/lestrrat-go/file-rotatelogs v2.4.0+incompatible
+	github.com/lestrrat-go/strftime v1.0.3 // indirect
```
jinfahua, line 18: Did u use this module?

go.mod
```diff
@@ -14,9 +14,12 @@ require (
...
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/pebbe/zmq4 v1.0.0
 	github.com/prometheus/client_golang v1.2.1
+	github.com/rifflock/lfshook v0.0.0-20180920164130-b9218ef580f5
```
jinfahua, line 22: Did u use this module?

jinfahua (changes requested):

EMQmyd:
> * 它是否具有控制日志文件大小的选项？例如，如果日志文件大小达到100MB，则创建另一个日志文件。
> * 同时更新相关文档。

Can be divided by interval time or file size

go.mod
EMQmyd: deleted

go.mod
EMQmyd: yes

PR #611, opened 2020-11-19
fix(window): hopping window consumes too much memory

ngjaying:
Unnecessary memory are holden when hopping window interval is bigger than length.


ngjaying:
Close as it is not a normal usage of hopping window. Usually, the window length is bigger than hop size.

PR #620, opened 2020-11-23
feat(addr):No way to limit client and REST APIs to localhost

EMQmyd:


etc/kuiper.yaml
```diff
@@ -9,9 +9,13 @@ basic:
   rotateTime: 24
   # Maximum file storage hours
   maxAge: 72
+  # CLI ip
+  ip: "0.0.0.0"
   # CLI port
   port: 20498
   # REST service port
```
jinfahua, line 16: Rest service bind address

etc/kuiper.yaml
```diff
@@ -9,9 +9,13 @@ basic:
   rotateTime: 24
   # Maximum file storage hours
   maxAge: 72
+  # CLI ip
```
jinfahua, line 12: CLI service bind address

etc/kuiper.yaml
```diff
@@ -9,9 +9,13 @@ basic:
...
+  ip: "0.0.0.0"
```
jinfahua, line 13: I don't think double quotations are necessary, could you please verify it?

jinfahua (changes requested):

ngjaying:
Is it better to follow docker compose's rule for [ports](https://docs.docker.com/compose/compose-file/#ports). 
```
port: 9081                     -> if no ip specified, defaults to 0.0.0.0; It is also backward compatible
port: 127.0.0.1:9081
```

Issue #613, opened 2020-11-20
[1.0.1] File logging error message when server fileLog=false


tonyespy:
For the Hanoi (1.3.0) release of EdgeX, we include Kuiper 1.0.1 in the [edgexfoundry snap](https://snapcraft.io/edgexfoundry) package published in the Snap Store.

When we build the snap, we modify `kuiper.yaml` and ensure that `consoleLog` is set to `true`, and `fileLog` is set to `false`, however Kuiper still tries to open the log file resulting in the following log message:

> level=info msg="Failed to log to file, using default stderr." file="common/util.go:161"

Note, since the snap was installed on Ubuntu 18.04, we additional see AppArmor messages preceding the above message which show that Kuiper is actually still trying to open the log file.

> audit: type=1400 audit(1605901473.210:6644): apparmor="DENIED" operation="mknod" profile="snap.edgexfoundry.kuiper" name="/var/log/stream.log" pid=13362 comm="kuiperd" requested_mask="c" denied_mask="c" fsuid=0 ouid=0

tonyespy:
@EMQmyd why did you close this, do you not consider it a bug?

EMQmyd:
This bug has been fixed.I forgot to associate pr.

EMQmyd:
bug(log):File logging error message when server fileLog=false #623

PR #623, opened 2020-11-24
bug(log):File logging error message when server fileLog=false

EMQmyd:


ngjaying (approved):

EMQmyd:
[1.0.1] File logging error message when server fileLog=false #613

Issue #615, opened 2020-11-21
Use window to filter the last n data


DaveZLB:
There is a scenario where I need to filter the last n pieces of data in the window every 1 hour.

Rui-Gan:
```
SELECT collect(*)[-n:-1] from demo GROUP BY TUMBLINGWINDOW(hh, 1);
```

Rui-Gan:
> Which version of eKuiper supports this ?

Versions 1.5.0 and above are supported, but we recommend using the latest version.

PR #625, opened 2020-11-25
Window limit

ngjaying:
#615


ngjaying:
We will optimize the window limit implicitly by analyzing the SQL, so close this one.

PR #635, opened 2020-12-03
feat(kvStore):Replace go-cache with sqlite

EMQmyd:


common/migratio.go
```diff
@@ -0,0 +1,108 @@
+package common
```
ngjaying, line 1: typo in file name. Should be migration

common/migratio_test.go
```diff
@@ -0,0 +1,66 @@
+package common
```
ngjaying, line 1: typo in file name

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/keepeye/logrus-filename v0.0.0-20190711075016-ce01a4391dd1
 	github.com/lestrrat-go/file-rotatelogs v2.4.0+incompatible
 	github.com/lestrrat-go/strftime v1.0.3 // indirect
+	github.com/mattn/go-sqlite3 v1.14.5
 	github.com/patrickmn/go-cache v2.1.0+incompatible
```
ngjaying, line 20: Can we remove go cache?

xstream/server/server/ruleManager.go
```diff
@@ -69,6 +69,7 @@ func doStartRule(rs *RuleState) error {
 				tp.GetContext().SetError(err)
 				logger.Printf("closing rule %s for error: %v", rs.Name, err)
 				tp.Cancel()
+				rs.Triggered = false
```
ngjaying, line 72: Is this done in another PR?

ngjaying (changes requested):

go.mod
EMQmyd: Need to use gocache to migrate data

xstream/server/server/ruleManager.go
EMQmyd: no The state of the rule needs to be changed when the rule is stopped

common/migratio_test.go
EMQmyd: ok

common/migratio.go
EMQmyd: ok

ngjaying (approved):

PR #645, opened 2020-12-11
feat(plugin):Add image plugin.

EMQmyd:


plugins/sinks/image/image.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	return nil
+}
+
+func (m *imageSink) Collect(ctx api.StreamContext, item interface{}) error {
+	logger := ctx.GetLogger()
+	if v, ok := item.([]byte); ok {
+		var msg []map[string][]byte
+		if err := json.Unmarshal(v, &msg); nil != err {
+			return err
```
ngjaying, line 159: Suggest to give a clearer error message for the user. The sink only accepts bytea field, other types are not supported.

ngjaying (changes requested):

ngjaying (approved):

Issue #647, opened 2020-12-13
Hope the system can provide GeoHash function


cmdares:
ref
https://github.com/sunng87/node-geohash

jinfahua:
Thanks for your proposal, we'll implement in future releases.

EMQmyd:
The geohash function plugin is located at: Add geohash function plugin #698

cmdares:
good job!

PR #698, opened 2021-01-28
Add geohash function plugin

EMQmyd:
Hope the system can provide GeoHash function #647

### DIFF
--- a/etc/functions/image.json
+++ b/etc/functions/image.json
@@ -8,17 +8,16 @@
 			"website": "https://www.emqx.io"
 		},
 		"helpUrl": {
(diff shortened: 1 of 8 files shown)


ngjaying (changes requested):
The plugin json file is not updated. Please also check if more files need changes by comparing with the image function plugin.

etc/functions/image.json
```diff
@@ -19,6 +19,7 @@
 	"libs": [
 	  "github.com/nfnt/resize@master"
 	],
+  "name":"image",
```
ngjaying, line 22: Indention problem

ngjaying (changes requested):

ngjaying (approved):

PR #706, opened 2021-02-05
feat(index):support SELECT sli[-1] FROM

EMQmyd:


ngjaying (commented):
It looks good. Just add more detail to the doc about the rule of index range: [start:end)

ngjaying (approved):

PR #719, opened 2021-02-19
feat(functions):cardinality(anyarray) function that returns the total…

EMQmyd:
… number of elements in the array, or 0 if the array is empty


ngjaying (changes requested):
This function is not an aggregate function. The argument is an array type field, not an aggregate rows.

ngjaying (approved):

PR #744, opened 2021-03-15
Fix links to window types in windows.md

jrtitus:
While reading through the docs, I noticed that the markdown was broken in this paragraph:

image

So I decided to make a quick PR to fix it.


ngjaying (approved):

ngjaying:
Thank you very much

PR #750, opened 2021-03-26
feat(table): support table for batch static data as data source

ngjaying:
Modify parser, planner, operator, topo, rest, cli and tests


xsql/parser.go
```diff
@@ -884,6 +891,9 @@ func validateStream(stmt *StreamStmt) error {
 	case common.FORMAT_JSON:
 		//do nothing
 	case common.FORMAT_BINARY:
+		if stmt.StreamType == TypeTable {
```
jinfahua, line 894: Would it be better to validate in the runtime instead of lexical phase? It could be probably complex to make this extensible?

xsql/parser.go
```diff
@@ -901,78 +911,124 @@ func validateStream(stmt *StreamStmt) error {
 	default:
 		return fmt.Errorf("option 'format=%s' is invalid", f)
 	}
+
+	if stmt.StreamType == TypeTable {
+		if t, ok := stmt.Options["TYPE"]; ok {
+			if strings.ToLower(t) != "file" {
```
jinfahua, line 917: Would it be better to validate in the runtime instead of lexical phase? It could be probably complex to make this extensible?

xsql/parser.go
```diff
@@ -901,78 +911,124 @@ func validateStream(stmt *StreamStmt) error {
...
 			}
-		} else {
-			return nil, fmt.Errorf("found %q, expected keyword stream.", lit1)
+		case TABLE:
+			ess := &DropTableStatement{}
+			if tok2, lit2 := p.scanIgnoreWhitespace(); tok2 == IDENT {
+				ess.Name = lit2
+				return ess, nil
+			} else {
+				return nil, fmt.Errorf("found %q, expected stream name.", lit2)
```
jinfahua, line 1028: `stream` should be `table`?

jinfahua (changes requested):

xsql/parser.go
ngjaying: extensible not supported yet so add the check here for a quick failure

xsql/parser.go
ngjaying: The same as above. Extensible to be supported later

jinfahua (approved):

PR #770, opened 2021-04-30
fix: do not print out the password in config & meta fields problem

ngjaying:
1. Fix the problem: multiple meta fields with only case differences may be retained in message. Should only have one as it is case insensitive.
2. Avoid printing out password in the log in EdgexSource.


jinfahua:
This file `https://github.com/emqx/kuiper/blob/master/deploy/docker/conf_util.go` also has password print issues.

jinfahua (approved):

Issue #772, opened 2021-05-05
Question about meta(origin) and compiling extended functions


hjsj4r:
Hi team, I'm using kuiper as edgex rules-engine running on docker, I'd appreciate your help with a couple issues I'm having.

My original implementation was using emqx/Kuiper: 1. 0. 0-alpine, Kuiper was configured to consume events from the edgex message bus. Those events have some "meta" values like the origin timestamp that I need to extract and export to an http endpoint, following the documentation I could extract and export the timestamps using a rule, the meta function and a template. The problem comes with the formatting of this field, it is being sent as a float number represented in scientific notation. I'm aware that Kuiper uses edgex's value descriptors to format the reading values, but, how can I format the timestamp?

This is the sql for the rule I want to implement:

   `SELECT reading, meta(created) AS time FROM demo WHERE meta(device) = "device1" `

so time is being exported as something like : 1.62000273e+09

I also tried to write an extended function to format the timestamp, i changed the docker image to emqx/kuiper:1.0.0 and tried to compile the function in the same running instance of kuiper:

    `go build -trimpath --buildmode=plugin -o /go/kuiper/_build/kuiper-1.0.0-linux-aarch64/plugins/functions $PWD/plugins/functions/format_timestamp.go`

after restarting kuiper and trying to use the function I get:

cannot open /go/kuiper/_build/kuiper-1.0.0-linux-aarch64/plugins/functions/format_timestamp.so: plugin.Open("/go/kuiper/_build/kuiper-1.0.0-linux-aarch64/plugins/functions/format_timestamp"): plugin was built with a different version of package internal/unsafeheader.

What's the correct way of doing this? 
Thank you for your time.

ngjaying:
Thanks for the very detail explanation.

We have a built-in function to format datetime, but currently it only supports datetime type field as the parameter. We will extend the support for bigint and float. Once implemented, your sql could be written as below. Notice that, I multiply your time by 1000 becasue you mentioned your timestamp value is 1.62000273e+09 which seems like not including the milliseconds so convert it to millisecond.

```sql
SELECT reading, format_time(meta(created) * 1000, "yyyy-MM-dd T HH:mm:ss") AS time FROM demo WHERE meta(device) = "device1"
```

Regarding the plugin problem, it could be caused by build parameter `-trimpath` which is only needed for v1.1.2 and later version. It seems you were using v1.1.0, so don't add that in the build command.

ngjaying:
Release in v1.2.0, so closed

PR #773, opened 2021-05-06
fix(func): support other types for datetime format funcs #772

ngjaying:


jinfahua:
Hi Jiyong, 

Following 2 docs need to be updated as well...

https://github.com/emqx/kuiper/blob/master/docs/en_US/sqls/built-in_functions.md#string-functions
` Add sample for how to use format time. `

https://github.com/emqx/kuiper/blob/master/docs/en_US/sqls/built-in_functions.md#conversion-functions
` Convert to datetime should be supported.`

jinfahua (approved):

Issue #776, opened 2021-05-11
Filter join error if the stream name is not specified


ngjaying:
Given a stream and a table. If filter with table field such as `fieldFromTable`, the field is parsed as a field of stream.
```sql
SELECT * from aStream INNER JOIN aTable ON fieldFromTable > 0
```

If explicitly specify the stream name of the field, it works fine:
```sql
SELECT * from aStream INNER JOIN aTable ON aTable.fieldFromTable > 0
```

PR #777, opened 2021-05-12
fix(join): problem when joining without stream name in fieldRef

ngjaying:
#776


jinfahua (changes requested):
Any doc that need to be updated as well?

ngjaying:
> Any doc that need to be updated as well?

No, these are bug fixes and refactors

jinfahua (approved):

PR #784, opened 2021-05-17
Update influx.md

tmactao:
Update the SQL statement in the rule configuration file influxRule.txt

### DIFF
--- a/docs/zh_CN/plugins/sinks/influx.md
+++ b/docs/zh_CN/plugins/sinks/influx.md
@@ -39,7 +39,7 @@
 ```json
 {
   "id": "influx",
-  "sql": "SELECT * from  demo_stream where temperature < 50",
+  "sql": "SELECT * from  demo_stream where temperature > 50",
   "actions": [
     {
       "log": {},


docs/zh_CN/plugins/sinks/influx.md
```diff
@@ -39,7 +39,7 @@
 ```json
 {
   "id": "influx",
-  "sql": "SELECT * from  demo_stream where temperature < 50",
+  "sql": "SELECT * from  demo_stream where temperature > 50",
```
ngjaying, line 42: Could you please explain why making this change?

tmactao: You can check the conditions in the example usage, the configuration file (influxRule.txt) is "an example for selecting a temperature greater than 50 degrees", so I change it

ngjaying: OK，thank you. Could you please also change the same file in the en_US? Thanks

tmactao: Ok, I will modify it

ngjaying (approved):

Issue #812, opened 2021-06-03
Support multiple sinks with the same type for ui meta


ngjaying:
Currently, if a rule have two sinks with the same type, such as two log sinks, the UI will only show one. That's because the meta api only returns one sink

PR #813, opened 2021-06-03
Refactor for sink meta API

ngjaying:
Delete the logic for rule value and property setting
#812


jinfahua:
[This doc](https://github.com/emqx/kuiper/blob/master/docs/en_US/plugins/overview.md) should be updated as well?

ngjaying:
> [This doc](https://github.com/emqx/kuiper/blob/master/docs/en_US/plugins/overview.md) should be updated as well?

No, the sink json only has the custom properties which are not changed, so no need to update. We just delete the option and base properties from the server side and let the client side store them directly.

jinfahua (approved):

PR #819, opened 2021-06-07
Window time function support

ngjaying:
#816 

Changes:
1. Change data structure for aggregate types to add WindowRange info. Change the corresponding operators and tests to accomodate the change.
2. Change call eval to handle window_start() and window_end()
3. Change aggregate operator to modify the alias calculation to return a clone instead of  the same instance because some data may have assigned to multiple windows.
4. Add various unit tests and docs for window_start and window_end


jinfahua:
Please update the docs to clarify,
- User specifies the event time
- User does not specify the event time

jinfahua (approved):

PR #822, opened 2021-06-08
Support more template functions

ngjaying:


docs/zh_CN/rules/overview.md
```diff
@@ -158,8 +158,14 @@
...
+
+1. Go 语言内置[模板函数](https://golang.org/pkg/text/template/#hdr-Functions)。
+2. 来自 [sprig library](http://masterminds.github.io/sprig/) 的丰富的扩展函数集。
+3. Kuiper 扩展的函数。
+
 Kuiper 扩展了几个可以在模版中使用的函数。
 
-- `json para1`: `json` 函数用于将 map 内容转换为 JSON 字符串
-- `base64 para1`: `base64` 函数用于将参数值编码为 base64 字符串
-- `add para1 para2`: `add` 函数用于将两个数值类型的参数相加
```
jinfahua, line 165: Where is this func? Is it deprecated?

jinfahua (changes requested):

docs/zh_CN/rules/overview.md
ngjaying: sprig library provides a func with the same name. So the user can continue to use it. Internally, we just delete our implementation

PR #860, opened 2021-06-28

smart33690:


ngjaying (changes requested):
Please rebase on master and run go fmt to the modified files in order to pass the test case static check.

PR #864, opened 2021-06-29

yongxingMa:


ngjaying (commented):
Please add the signoff, check https://github.com/lf-edge/ekuiper/pull/864/checks?check_run_id=2938051301 for detail

PR #869, opened 2021-06-29

yongxingMa:
Signed-off-by: 726518972@qq.com <Smart33690>


yongxingMa:
Why does DCO check not successful，Which step is wrong？

ngjaying:
> Why does DCO check not successful，Which step is wrong？

Check the error message "Committer: 726518972@qq.com; Smart33690 is not a valid email address. ". Use `git config --global user.email "726518972@qq.com"` to configure the email address.

PR #878, opened 2021-07-06
Fix resubscription failure when reconnecting

tixff:
When the MQTT client reconnects, the subscription function may be executed before the connection, resulting in the following error:
**Found error: not currently connected and ResumeSubs not set**
In order to avoid subscription failure, it is best to initiate a subscription after the connection is successful

Of course there might be a better way to do this, but I'm just offering an idea


ngjaying (commented):
Shall we remove the subscirbe in opts.SetConnectionLostHandler?

tixff:
> Shall we remove the subscirbe in opts.SetConnectionLostHandler?



> Shall we remove the subscirbe in opts.SetConnectionLostHandler?

It's best not to remove it

ngjaying:
> 
> It's best not to remove it

It looks like subscribe will run twice for reconnection. Is it idempotent? Is it possible that the subscription will happen twice thus receive duplicate messages?

ngjaying (approved):
Thank you, approve.

Issue #905, opened 2021-07-22
Use eKuiper-kubernets-tool, execute command “docker logs” no log output


yongxingMa:
### When I use eKuiper-kubernets-tool，execute command “docker logs” no log output，I have to enter the container to view the log。

```
[root@ecs-s3-small-1-linux-20200611135428 ~]# docker logs aeea82a4a4bf
Kuiper kubernetes tool is started successfully!

[root@ecs-s3-small-1-linux-20200611135428 ~]# docker exec -it aeea82a4a4bf /bin/sh
/kuiper-kubernetes-tool # cat log/kubeedge.log 
time="2021-07-22T11:28:29Z" level=info msg="http://127.0.0.1:9081/streams:post resp:Stream stream1 is created." file="logger.go:192"
time="2021-07-22T11:28:29Z" level=info msg="http://127.0.0.1:9081/streams:post resp:Stream stream2 is created." file="logger.go:192"
time="2021-07-22T11:28:29Z" level=info msg="http://127.0.0.1:9081/streams:get resp:[\"kubeEdgeStream\",\"stream1\",\"stream2\"]\n" file="logger.go:192"
time="2021-07-22T11:28:29Z" level=info msg="http://127.0.0.1:9081/streams/stream1:get resp:{\"Name\":\"stream1\",\"StreamFields\":[{\"FieldType\":\"bigint\",\"Name\":\"id\"},{\"FieldType\":\"string\",\"Name\":\"name\"},{\"FieldType\":\"float\",\"Name\":\"score\"}],\"Options\":{\"DATASOURCE\":\"topic/temperature\",\"FORMAT\":\"json\",\"KEY\":\"id\"}}\n" file="logger.go:192"
time="2021-07-22T11:28:29Z" level=info msg="http://127.0.0.1:9081/streams/stream2:delete resp:Stream stream2 is dropped." file="logger.go:192"
time="2021-07-22T11:28:29Z" level=info msg="http://127.0.0.1:9081/rules:post resp:Rule rule1 was created successfully." file="logger.go:192"
time="2021-07-22T11:28:29Z" level=info msg="http://127.0.0.1:9081/rules/rule1:put resp:Rule rule1 was updated successfully." file="logger.go:192"
time="2021-07-22T11:28:29Z" level=info msg="http://127.0.0.1:9081/rules:post resp:Rule rule2 was created successfully." file="logger.go:192"
[...]

PR #915, opened 2021-08-02
feat(tool): for ekuiper kubernetes tool, add options for log print location

SuperRxan:
Add two options in etc/cf.yaml file for log print
    * consoleLog: true/false
    * fileLog: true/false
Closes: #905 

### DIFF
--- a/tools/kubernetes/conf/conf.go
+++ b/tools/kubernetes/conf/conf.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"gopkg.in/yaml.v3"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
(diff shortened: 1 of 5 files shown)


tools/kubernetes/conf/conf.go
```diff
@@ -132,15 +135,22 @@ func (c *config) initLog() bool {
...
+			fmt.Println("Failed to init log file settings..." + err.Error())
+			Log.Infof("Failed to log to file, using default stderr.")
+			return false
+		} else if c.ConsoleLog {
+			mw := io.MultiWriter(os.Stdout, logFile)
+			Log.SetOutput(mw)
+		} else if !c.ConsoleLog {
+			Log.SetOutput(logFile)
+		}
+	} else if c.ConsoleLog {
```
ngjaying, line 150: fileLog & consoleLog seems mutual exclusive. But they should can be existed together

ngjaying (approved):

PR #921, opened 2021-08-04
fix,refactor(sqlite): Fix (and refactoring) related to the too many files open issue

rwadowski:
Fix (and refactoring) related to the too many files open issue - since sqlite does not work well with concurrent writes - kv store was rewritten to use single connection for access to the database,
* introduced database structure which will take away sql database specifics (sqlite in this case)
* fix in topo.go - there was possible scenario that during Open execution of underlying goroutine Cancel was called and store property was set to nil which was leading to null pointer exception


ngjaying:
@rwadowski Awesome!
You need to signoff. Please click into the detail of DCO check and follow the instruction to pass the check.

rwadowski:
@ngjaying  what are fvt tests ? How can I run them locally ?

ngjaying:
@rwadowski Thanks again for your great work! I append my commit to yours：
1. Leave only kv interface in pkg and move other sqlkv implementation to internal. So that sqlite dependency is not exported.
2. Refactor some API to move error to the last return value to comply with go convention.
3. Extract kvstore initialization as Setup function so that both server startup and testings can init it by one method call.
4. Fix all unit test and fvt test failures.

Please review and feel free to add any comments.

rwadowski:
@ngjaying I understand that PR will undergo a review process.

ngjaying (approved):

ngjaying:
Merged, thanks @rwadowski

Issue #954, opened 2021-08-18
The TIMESTAMP_FORMAT property of the stream is not implemented


wangzh1024:
**Environment**:

- eKuiper version: 
1.2.1
- OS :
Raspbian GNU/Linux 10 (buster)

**What happened and what you expected to happen**:
When I create the stream, I customize the time type through the TIMESTAMP_FORMAT property. But when I query the time, the following error occurs：
` [{"error":"error in preprocessor: invalid data type for StationDealTime, cannot convert to datetime: parsing time "2021-08-18 15:16:23" as "2006-01-02T15:04:05.000Z07:00": cannot parse " 15:16:23" as "T""}]   `
The reason is that the TIMESTAMP_FORMAT property is not implemented. I hope this property can be implemented as soon as possible.

PR #955, opened 2021-08-18
fix(planner): set timestamp format property in processing time mode

ngjaying:
The timestampFormat property is also useful in processing time mode when defining a datetime type and receive a string value and vise versa. So the property must be complied in processing time mode.

Closes: #954


docs/en_US/sqls/streams.md
```diff
@@ -37,6 +37,8 @@ CREATE STREAM
 | StrictValidation     | true | To control validation behavior of message field against stream schema. See [Strict Validation](#Strict Validation) for more info. |
 | CONF_KEY | true | If additional configuration items are requied to be configured, then specify the config key here. See [MQTT stream](../rules/sources/mqtt.md) for more info. |
 | SHARED | true | Whether the source instance will be shared across all rules using this stream |
+| TIMESTAMP | true | The field to represent the event's timestamp. If specified, the rule will run with event time. Otherwise, it will run with processing time. Please refer to [timestamp management](./windows.md#timestamp-management) for details. |
```
jinfahua, line 40: Change `TIMESTAMP` to `TIMESTAMP_FIELD` to make it more specific?

jinfahua (changes requested):

docs/en_US/sqls/streams.md
ngjaying: I suggest no because it will cause breaking API changes

jinfahua (approved):

Issue #937, opened 2021-08-13
doc: sql language elements join/group by must working in window


ngjaying:
Update the doc to clarify the usage and example of join/group by https://github.com/lf-edge/ekuiper/blob/master/docs/en_US/sqls/query_language_elements.md. If joining two streams, it must be used within window.

PR #958, opened 2021-08-19
doc(sql): clear description for join/group by

ngjaying:
Closes: #937

### DIFF
--- a/internal/topo/planner/planner.go
+++ b/internal/topo/planner/planner.go
@@ -266,7 +266,7 @@ func createLogicalPlan(stmt *ast.SelectStatement, opt *api.RuleOption, store kv.
 			p.SetChildren(append(children, tableChildren...))
 			children = []LogicalPlan{p}
 		} else if w == nil {
-			return nil, errors.New("need to run stream join in windows")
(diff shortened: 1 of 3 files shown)


internal/topo/planner/planner.go
```diff
@@ -266,7 +266,7 @@ func createLogicalPlan(stmt *ast.SelectStatement, opt *api.RuleOption, store kv.
 			p.SetChildren(append(children, tableChildren...))
 			children = []LogicalPlan{p}
 		} else if w == nil {
-			return nil, errors.New("need to run stream join in windows")
+			return nil, errors.New("need to run stream join within a window by using the GROUP BY WINDOW clause")
```
jinfahua, line 269: need to run JOIN statement within a window by using the GROUP BY WINDOW clause

jinfahua (changes requested):

internal/topo/planner/planner.go
ngjaying: yes. otherwise, the rule will receive one message from a single stream which has no meaning of join. A window is needed to hold multiple messages so that they could be from multiple streams to join.

jinfahua (approved):

PR #965, opened 2021-08-20
conf(env) Kuiper can be configured via env variables

rwadowski:
@ngjaying it turns out that there is a ready library to use for this purpose


ngjaying (commented):
Great work. Just several concerns from me:
1. We actually only need a small protion of the features of viper. Could you check how much does the binary size increase after adding this dependency.
2. Besides kuiper.yaml, we also have other yaml configs in etc/sources and maybe other places. We may need to support all of them as well. It will be better to eliminate the direct dependenct to gopkg.in/yaml.v3 after all changes.

rwadowski:
> Great work. Just several concerns from me:
> 
> 1. We actually only need a small protion of the features of viper. Could you check how much does the binary size increase after adding this dependency.

`-rwxrwxr-x 1 rwadowski rwadowski 23841760 Aug 23 10:35 no_viper`
`-rwxrwxr-x 1 rwadowski rwadowski 25699552 Aug 23 10:34 with_viper`

It is more less 7% increase of memory size

rwadowski:
@ngjaying 
> 2\. Besides kuiper.yaml, we also have other yaml configs in etc/sources and maybe other places. We may need to support all of them as well. It will be better to eliminate the direct dependenct to gopkg.in/yaml.v3 after all changes.
I guess you are writing for example about conf_util.go which seems to do similar thing

ngjaying:
> @ngjaying
> 
> > 2. Besides kuiper.yaml, we also have other yaml configs in etc/sources and maybe other places. We may need to support all of them as well. It will be better to eliminate the direct dependenct to gopkg.in/yaml.v3 after all changes.
> > I guess you are writing for example about conf_util.go which seems to do similar thing

Not just conf_util.go. There are many other places like internal/topo/node/node.go that is using "gopkg.in/yaml.v3" to parse config like etc/sources/edgex.yaml.

rwadowski:
> > @ngjaying
> > > 1. Besides kuiper.yaml, we also have other yaml configs in etc/sources and maybe other places. We may need to support all of them as well. It will be better to eliminate the direct dependenct to gopkg.in/yaml.v3 after all changes.
> > >    I guess you are writing for example about conf_util.go which seems to do similar thing
> 
> Not just conf_util.go. There are many other places like internal/topo/node/node.go that is using "gopkg.in/yaml.v3" to parse config like etc/sources/edgex.yaml.

I guess it will be faster for me to remove dependency and see what will crash

ngjaying:
> > Great work. Just several concerns from me:
> > 
> > 1. We actually only need a small protion of the features of viper. Could you check how much does the binary size increase after adding this dependency.
> 
> `-rwxrwxr-x 1 rwadowski rwadowski 23841760 Aug 23 10:35 no_viper`
> `-rwxrwxr-x 1 rwadowski rwadowski 25699552 Aug 23 10:34 with_viper`
> 
> It is more less 7% increase of memory size

It looks like an obvious increase for more than 1MB. I don't know if increase can be much smaller if we can remove all yaml dependencies.

rwadowski:
@ngjaying what's more it looks like viper depends on the gopkg.in/yaml.v3

rwadowski:
> It looks like an obvious increase for more than 1MB. I don't know if increase can be much smaller if we can remove all yaml dependencies.

It seems that not all yaml dependencies will be removed as I mentioned in other post due to vipers dependency on the yaml ones.

ngjaying:
> > It looks like an obvious increase for more than 1MB. I don't know if increase can be much smaller if we can remove all yaml dependencies.
> 
> @ngjaying It seems that not all yaml dependencies will be removed as I mentioned in other post due to vipers dependency on the yaml ones. So the question whether increase of footprint will be acceptable at all ?

If not very hard to implement by ourselves, I would prefer to not introduce vipers dependency.

rwadowski:
> > > It looks like an obvious increase for more than 1MB. I don't know if increase can be much smaller if we can remove all yaml dependencies.
> > 
> > 
> > @ngjaying It seems that not all yaml dependencies will be removed as I mentioned in other post due to vipers dependency on the yaml ones. So the question whether increase of footprint will be acceptable at all ?
> 
> If not very hard to implement by ourselves, I would prefer to not introduce vipers dependency.

Well I am far from being a golang expert but I guess it is possible to iterate over config struct and look for environment variables built on path in the tree - this would be a relatively generic way of loading envs. Is it something you think about ?

ngjaying:
> > > > It looks like an obvious increase for more than 1MB. I don't know if increase can be much smaller if we can remove all yaml dependencies.
> > > 
> > > 
> > > @ngjaying It seems that not all yaml dependencies will be removed as I mentioned in other post due to vipers dependency on the yaml ones. So the question whether increase of footprint will be acceptable at all ?
> > 
> > 
> > If not very hard to implement by ourselves, I would prefer to not introduce vipers dependency.
> 
> Well I am far from being a golang expert but I guess it is possible to iterate over config struct and look for environment variables built on path in the tree - this would be a relatively generic way of loading envs. Is it something you think about ?

It could be a refactor of conf_util.go, but instead of write to file, just rewrite the conf values in memory.

rwadowski:
> > > > > It looks like an obvious increase for more than 1MB. I don't know if increase can be much smaller if we can remove all yaml dependencies.
> > > > 
> > > > 
> > > > @ngjaying It seems that not all yaml dependencies will be removed as I mentioned in other post due to vipers dependency on the yaml ones. So the question whether increase of footprint will be acceptable at all ?
> > > 
> > > 
> > > If not very hard to implement by ourselves, I would prefer to not introduce vipers dependency.
> > 
> > 
> > Well I am far from being a golang expert but I guess it is possible to iterate over config struct and look for environment variables built on path in the tree - this would be a relatively generic way of loading envs. Is it something you think about ?
> 
> It could be a refactor of conf_util.go, but instead of write to file, just rewrite the conf values in memory.

@ngjaying will refactor of the conf_util.go will apply for normal app runtime ? As I see it is in /deploy/docker package which probably applies for docker. I think we need something that will apply env variable functionality to any type of deployment. I am not sure how this precisely works though ...

ngjaying:
> @ngjaying will refactor of the conf_util.go will apply for normal app runtime ? As I see it is in /deploy/docker package which probably applies for docker. I think we need something that will apply env variable functionality to any type of deployment. I am not sure how this precisely works though ...

Yes, exactly
1. Merge conf_util to the "app runtime" instead of outside of the main package
2. Refactor it to onveride the env variable to memory instead of changing the conf file
3. Maybe a more graceful way to deal with the config key case as the env are upper case but the conf key varies.

rwadowski:
> Yes, exactly
> 
> 1. Merge conf_util to the "app runtime" instead of outside of the main package
> 2. Refactor it to onveride the env variable to memory instead of changing the conf file
> 3. Maybe a more graceful way to deal with the config key case as the env are upper case but the conf key varies.

So keeping simple hardcoded map of env keys is fine ? We will have to add new keys there if there will be a new feature but after all new config will have to be added as well so env handling can be done in this case too.

ngjaying:
> > Yes, exactly
> > 
> > 1. Merge conf_util to the "app runtime" instead of outside of the main package
> > 2. Refactor it to onveride the env variable to memory instead of changing the conf file
> > 3. Maybe a more graceful way to deal with the config key case as the env are upper case but the conf key varies.
> 
> So keeping simple hardcoded map of env keys is fine ? We will have to add new keys there if there will be a new feature but after all new config will have to be added as well so env handling can be done in this case too.

Do you mean `file_keys_map`? That is not hardcoded map of all keys. It is actually a map of upper case key to the camel case which cannot be calculated. For keys like `debug` are not in the map but can still be used as they are single word and can be deduced by just turn the env to lowercase. And this is needed to be refactored as my point 3.

rwadowski:
@ngjaying shouldn't we take assumption that env variable coresponds to name in config for eample
```
basic:
   debug: true
```
should have env variable like `KUIPER_BASIC_DEBUG` ?
This will give us a very easy naming convention for env variables.

ngjaying:
> @ngjaying shouldn't we take assumption that env variable coresponds to name in config for eample
> 
> ```
> basic:
>    debug: true
> ```
> 
> should have env variable like `KUIPER_BASIC_DEBUG` ?
> This will give us a very easy naming convention for env variables.

We do have the convention. This is the current convention `KUIPER__BASIC__DEBUG` which will map to kuper.yaml/basic/debug. And this does NOT need any hardcoded map. However, `KUIPER__BASIC__CONSOLELOG` currently needs a map, as we cannot know which alphabet need to be uppercase as the config key is consoleLog. The problem may be that we should support case insensitive config key.

rwadowski:
> > @ngjaying shouldn't we take assumption that env variable coresponds to name in config for eample
> > ```
> > basic:
> >    debug: true
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > should have env variable like `KUIPER_BASIC_DEBUG` ?
> > This will give us a very easy naming convention for env variables.
> 
> We do have the convention. This is the current convention `KUIPER__BASIC__DEBUG` which will map to kuper.yaml/basic/debug. And this does NOT need any hardcoded map. However, `KUIPER__BASIC__CONSOLELOG` currently needs a map, as we cannot know which alphabet need to be uppercase as the config key is consoleLog. The problem may be that we should support case insensitive config key.

Yeah, I was going into this direction that make all upper case envs like in linuks. As for now I extracted single file parsing into the internal/conf/load.go

rwadowski:
I am looking for places in the code where edgex.yaml, random.yaml, zmq.yaml, httppull.yaml, client.yaml are loaded

rwadowski:
@ngjaying I have applied all changes with difference of the env naming. Can you look at it ? Without mapping code is simpler I think

rwadowski:
@ngjaying another thing about config stuff
instead of 
```
sink:
  # The cache persistence threshold size. If the message in sink cache is larger than 10, then it triggers persistence. If you find
  # the remote system is slow to response, or sink throughput is small, then it's recommend to increase below 2 configurations.
  # More memory is required with the increase of below 2 configurations.
  # If the message count reaches below value, then it triggers persistence.
  cacheThreshold: 10
  # The message persistence is triggered by a ticker, and cacheTriggerCount is for using configure the count to trigger the persistence procedure
  # regardless if the message number reaches cacheThreshold or not. This is to prevent the data won't be saved as the cache never pass the threshold.
  cacheTriggerCount: 15
```
shouldn't be 
```
sink:
  cache:
    # The cache persistence threshold size. If the message in sink cache is larger than 10, then it triggers persistence. If you find
    # the remote system is slow to response, or sink throughput is small, then it's recommend to increase below 2 configurations.
    # More memory is required with the increase of below 2 configurations.
    # If the message count reaches below value, then it triggers persistence.
    threshold: 10
    # The message persistence is triggered by a ticker, and cacheTriggerCount is for using configure the count to trigger the 
    #persistence procedure
    # regardless if the message number reaches cacheThreshold or not. This is to prevent the data won't be saved as the cache   
    # never pass the threshold.
  trigger:
    count: 15
```

ngjaying (changes requested):

ngjaying:
> @ngjaying another thing about config stuff
> instead of
> 
> ```
> sink:
>   # The cache persistence threshold size. If the message in sink cache is larger than 10, then it triggers persistence. If you find
>   # the remote system is slow to response, or sink throughput is small, then it's recommend to increase below 2 configurations.
>   # More memory is required with the increase of below 2 configurations.
>   # If the message count reaches below value, then it triggers persistence.
>   cacheThreshold: 10
>   # The message persistence is triggered by a ticker, and cacheTriggerCount is for using configure the count to trigger the persistence procedure
>   # regardless if the message number reaches cacheThreshold or not. This is to prevent the data won't be saved as the cache never pass the threshold.
>   cacheTriggerCount: 15
> ```
> 
> shouldn't be
> 
> ```
> sink:
>   cache:
>     # The cache persistence threshold size. If the message in sink cache is larger than 10, then it triggers persistence. If you find
>     # the remote system is slow to response, or sink throughput is small, then it's recommend to increase below 2 configurations.
>     # More memory is required with the increase of below 2 configurations.
>     # If the message count reaches below value, then it triggers persistence.
>     threshold: 10
>     # The message persistence is triggered by a ticker, and cacheTriggerCount is for using configure the count to trigger the 
>     #persistence procedure
>     # regardless if the message number reaches cacheThreshold or not. This is to prevent the data won't be saved as the cache   
>     # never pass the threshold.
>     trigger:
>       count: 15
> ```
> 
> ?

There are not many configurations under sink, so we just flat them to avoid introducing too many hierarchies.

internal/conf/load.go
```diff
@@ -0,0 +1,137 @@
...
+	err = yaml.Unmarshal(b, &configMap)
+	if err != nil {
+		return err
+	}
+	err = process(configMap, os.Environ())
+	if err != nil {
+		return err
+	}
+	var bs []byte
+	bs, err = yaml.Marshal(configMap)
```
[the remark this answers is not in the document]
rwadowski: @ngjaying I am not sure how to make it better, I need to convert map[string]interface{} to struct / interface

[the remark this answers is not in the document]
rwadowski: @ngjaying in case of camelCase. I have not handled it properly - I will need to add mapping but I have feeling that proper hierarchy in they yaml config would solve the problem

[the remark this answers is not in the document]
ngjaying: ah ok. I think current approach is ok. It is just an one time action. Does this also solve the problem of lowercase -> camelCase?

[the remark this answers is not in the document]
rwadowski: nope, I am about commit fix about camel case handling,I have changed Unmatshalling/Marshalling to sth that looks better at a cost of additional dependency.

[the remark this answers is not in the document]
rwadowski: I have commited it right now, I have feeling that it looks better right now

extensions/go.mod
```diff
@@ -3,14 +3,14 @@ module github.com/lf-edge/ekuiper/extensions
 go 1.16
 
 require (
+	github.com/go-redis/redis/v8 v8.11.3
```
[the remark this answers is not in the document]
rwadowski: Fixed

rwadowski:
@ngjaying what check logs step in run_fvt_tests does ?

rwadowski:
> @rwadowski I would like to get rid of the mapping. I think it is possible like adding a json tag in addition to yaml tag in the config strutct fields. Let the json tag be case insensitive like all lowercase and try to leverage it to do the decode.

@ngjaying can you elaborate more ? I don't know golang ecosystem well. I am trying to find something how to make it feasible but I can't see anything that would be handy.

internal/conf/load.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func LoadConfigFromPath(path string, c interface{}, mapping EnvNameMapping) error {
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		return err
+	}
+	configMap := make(map[string]interface{})
+	err = yaml.Unmarshal(b, &configMap)
+	if err != nil {
```
ngjaying, line 109: I suppose the keys in the configMap are camelCase as they specified in the yaml.  Is the yaml tag in the config struct like  ```ConsoleLog   bool   `yaml:"consoleLog"` ``` is still used in the decode of line 116? For envs, if we let the key to be all lowercase and just set into the configMap, is it possible to let the decode map to the correct case. Maybe it is possible to add a new tag in the config struct field like ```ConsoleLog   bool   `yaml:"consoleLog",json:"consolelog"` ```  in which the json tag name is all lowercase.

rwadowski: Is something I committed fine ? I leveraged [this](https://pkg.go.dev/github.com/mitchellh/mapstructure#hdr-Field_Tags)

ngjaying:
We are releasing 1.3.1 today. After then, we will merge back dev-1.4.0 to master and try to merge this. Thank you!

rwadowski:
> We are releasing 1.3.1 today. After then, we will merge back dev-1.4.0 to master and try to merge this. Thank you!

I assume that review can be closed ?

ngjaying:
> > We are releasing 1.3.1 today. After then, we will merge back dev-1.4.0 to master and try to merge this. Thank you!
> 
> I assume that review can be closed ?

We can keep this. Just rebase it on the new master later.

ngjaying:
Force push to this after rebasing the new master. Will do some smoke testing and merge this if all are fine. Thanks.

cmd/kuiper/main.go
```diff
@@ -64,22 +63,20 @@ func main() {
 	//		Usage: "the name of stream",
 	//	}}
 
-	b, err := conf.LoadConf(clientYaml)
+	var cfg map[string]clientConf
+	err := conf.LoadConfig(&cfg)
```
ngjaying, line 67: This is wrong. Here we should read clientYaml: client.yaml but the LoadConfig read from kuiper.yaml

internal/conf/load.go
```diff
@@ -0,0 +1,155 @@
...
+		return err
+	}
+	var configs map[string]interface{}
+	if _, success := c.(*map[string]interface{}); success {
+		configs = configMap
```
ngjaying, line 56: I suppose here is the case of source config reading. Even read into map, we still need to deal with case. For example, edgex.yaml has a key "messageType" in yaml and after reading the map key is also camel case "messageType"; However, the env var will be parsed to "messagetype". How do we deal with this right now?

ngjaying (changes requested):

internal/conf/load.go
ngjaying: maybe we can get the key case from its json file, for example edgex.json has a edgex.json file with properties section that listed all the available properties.

cmd/kuiper/main.go
rwadowski: Fixed

internal/conf/load.go
rwadowski: @ngjaying I disabled normalization for map type config due to one of failing tests. Maybe we should fix the test instead of creating an edge case for the map type ?

ngjaying: There is actually a big difference here. For struct, we can implicitly define the case mapping by the yaml tag. But for map, there is no tag. How do you know what an env var MESSAGETYPE should be converted to?

rwadowski: I think it would good to make the same behavior - make everything case insensitive. Somehow it does not convince me that for yaml we should be case insensitive while for map case sensitive.

ngjaying: yaml is case sensitive now as we read it from `yaml.Unmarshal(b, &output)`. We did not want to change this behavior. When the output is a struct, the case sensitive yaml key are map to the struct field with the help of yaml tag of the struct. Othwise, if the outpus is a map, the case sensitive yaml key directly become the map key. For example, yaml key `messageType` will become `MessageType` in a struct and become `messageType` in a map. 

The problem is that the env is case insensitive and you normalize key to handle it. If the output is a struct, we can normalize the key and changes back to the case sensitive struct key with the help of mapstructure decode. But for map, if we normalize the key, we cannot revert back. We seems to miss the key case information. And that's my concern.

ngjaying: Moreover, the struct always define all available keys(fields), but map we only read from yaml. If the yaml does not define all keys, we have no way to know if a key from ENV is valid or not. I think the root cause is the missing information of property meta data for source configure. We can actually read that meta data from an optional json file like edgex.json.

rwadowski: > Moreover, the struct always define all available keys(fields), but map we only read from yaml. If the yaml does not define all keys, we have no way to know if a key from ENV is valid or not. I think the root cause is the missing information of property meta data for source configure. We can actually read that meta data from an optional json file like edgex.json.

You mean add such functionality - json to map[string]interface{} ?

ngjaying: No. There is an optional json file for each plugin. For example, etc/sources/edgex.json defines the meta data of edgex source.  Inside it, there is a section properties which define all the available properties with the correct case. When parsing yaml to map we can use the properties information to validate the keys. When convert ENV to conf, we can also use the properties information to map the key from upper case to the correct case.

rwadowski: It is an array - how does this refer to config ?

ngjaying: Let me take edgex as an example. From etc/sources/edgex.json, we can get a list of correct case properties(keys) like ["protocol", "messageType"...]. In internal/topo/node/node.go line 182, you parse yaml to a map. At there, we can loop the map key and validate against the key array of json. Then in this file, when handling env, we can still normalize the key and when decoding back to map, use the key array of json to restore the case of the keys.

rwadowski: ok I think I finally understood what you are writing about, where I am supposed to look for these files. One file at particular ocassion will suffice or particular yaml requires particular json ?

ngjaying: Great! Each yaml should have a json file alongside for all built-in sources. It is optional for extended plugin, so if json is not found, we can just skip this check.

rwadowski: Awesome, are these json locations the same as yaml counterparts or they have specific location ?

ngjaying: yes, they are at the same location.

rwadowski: hi, I made the changes but I have problems [here](https://github.com/lf-edge/ekuiper/runs/3549255802?check_suite_focus=true), can you help me ?

ngjaying: Hi Robert, that should be a network problem of the tdengine client download. We are aware of it and just check with the tdengine team. Let's just leave it there.

rwadowski: What about changes, is this you were writing about ?

ngjaying: Yes, thanks

ngjaying (dismissed):

ngjaying:
I'll need to rebase and merge it.

internal/conf/load.go
```diff
@@ -0,0 +1,243 @@
...
+	err = process(configs, os.Environ())
+	if err != nil {
+		return err
+	}
+	return mapstructure.Decode(configs, c)
+}
+
+func process(configMap map[string]interface{}, variables []string) error {
+	for _, e := range variables {
+		if !strings.HasPrefix(e, Prefix) {
```
ngjaying, line 74: The Prefix should not be a constant. It should be a variable. When parsing a source config, the prefix is the source name. For example, EDGEX__DEFAULT__PROTOCOL.

ngjaying (changes requested):
@rwadowski I have merged this and add a log to help debugging if env has applied successfully. I also find another problem when trying locally, I have added it to the review. Notice that, I have force push my changes to your branch. When you continue working with this, please fetch and reset to the latest. `git fetch origin` `git reset --hard ISSUE-962`

internal/conf/load_test.go
```diff
@@ -0,0 +1,122 @@
...
+		t.Error(err)
+	}
+
+	if c.Basic.ConsoleLog != true {
+		t.Errorf("env variable should set it to true")
+	}
+}
+
+func TestJsonCamelCase(t *testing.T) {
+	key := "KUIPER__DEFAULT__BODYTYPE"
```
ngjaying, line 45: Here the key prefix should be the same as the yaml name. HTTPPULL_DEFAULT_BODYTYPE

internal/conf/load.go
rwadowski: ok so how can I determine prefix ?

internal/conf/load_test.go
rwadowski: I have pushed fixes

internal/conf/load.go
```diff
@@ -44,8 +43,10 @@ func LoadConfigByName(name string, c interface{}) error {
 	return LoadConfigFromPath(p, c)
 }
 
-func LoadConfigFromPath(path string, c interface{}) error {
-	b, err := ioutil.ReadFile(path)
+func LoadConfigFromPath(p string, c interface{}) error {
+	prefix := getPrefix(p)
+	fmt.Println(prefix)
```
ngjaying, line 48: Remove this

ngjaying (changes requested):
It looks good now except one unnecessary print. Also could you use rebase to get rid of the merge commits to keep a linear history.

internal/conf/load.go
rwadowski: After applying fixes this test started to fail - [here](https://github.com/lf-edge/ekuiper/actions/runs/1216670656) which is pretty surprising

ngjaying: It is a timeout, usually rerun would solve it.

ngjaying (approved):
Merging

PR #969, opened 2021-08-23
fix(redis sink plugin): use redis v7 lib in redis sink to consistent with edgeX

SuperRxan:
fix(redis sink plugin): use redis v7 lib in redis sink to consistent with edgeX


SuperRxan:
need update the redis plugin in 1.3.0 release

ngjaying (approved):

PR #1001, opened 2021-09-08
fix(store): the db path must be fixed

ngjaying:
Due to CI and other tools' requirement


ngjaying:
@rwadowski I am removing the path configuration for sqlite because it may violate some installation combinations. We still want the path default to the data dir. Do you have any concerns about this change?

rwadowski:
> @rwadowski I am removing the path configuration for sqlite because it may violate some installation combinations. We still want the path default to the data dir. Do you have any concerns about this change?

ok

SuperRxan (approved):

PR #1009, opened 2021-09-15
Initial implementation of shared node between streams

rwadowski:
Signed-off-by: Robert Wadowski <robert.wadowski1@gmail.com>


ngjaying (commented):
I was thinking to use channels to connect between memory sink & source. Basically, we can have a global channel map like `map[string]map[string]chan` where key is the topic and the value is a list of channels for sources like {source1: ch1, source2: ch2}. When opening a memory sink, get or create an entry in that pool. When opening a source, we create a channel for it and attach it to one or multiple entries in the channel map (because the topic may be like wildcard like devices/*). We can take advantage of the go channel which is very suitable for this kind of data pipeline task.

rwadowski:
> I was thinking to use channels to connect between memory sink & source. Basically, we can have a global channel map like `map[string]map[string]chan` where key is the topic and the value is a list of channels for sources like {source1: ch1, source2: ch2}. When opening a memory sink, get or create an entry in that pool. When opening a source, we create a channel for it and attach it to one or multiple entries in the channel map (because the topic may be like wildcard like devices/*). We can take advantage of the go channel which is very suitable for this kind of data pipeline task.

Essentially it is such kind of structure. The pool is shared nodes where key is node id, every structure there has list of channels to sources...

ngjaying:
> > I was thinking to use channels to connect between memory sink & source. Basically, we can have a global channel map like `map[string]map[string]chan` where key is the topic and the value is a list of channels for sources like {source1: ch1, source2: ch2}. When opening a memory sink, get or create an entry in that pool. When opening a source, we create a channel for it and attach it to one or multiple entries in the channel map (because the topic may be like wildcard like devices/*). We can take advantage of the go channel which is very suitable for this kind of data pipeline task.
> 
> Essentially it is such kind of structure. The pool is shared nodes where key is node id, every structure there has list of channels used by sources... . It is more nicely packed.

The pool actually contains a list of source (may not be garbage collected when the source is holding by this global pool) instead of channels which means the sink and source are coupled. Is it possible here to let multiple sinks send messages to multiple sources? When using golang, I suggest to take advantage of the CSP model provided by channels. And the source here looks very different from the common pattern, in which open is a long running go routine. We can still use the same pattern by assigning a channel to it by the pool. In configure, it can request a channel from the pool; in the open, it can listen to that channel like a subscription; and in the close, it can release the channel from the pool.

ngjaying:
Basically, the pool is a message queue. And the memory sink just publish to it and the source sub from it. With that, the pool may also be possible to switch to an external message broker implementation.

rwadowski:
@ngjaying hi, can you help me with this build [problem](https://github.com/lf-edge/ekuiper/runs/3609653807?check_suite_focus=true) ? I am not sure how it is related to my changes

ngjaying:
> @ngjaying hi, can you help me with this build [problem](https://github.com/lf-edge/ekuiper/runs/3609653807?check_suite_focus=true) ? I am not sure how it is related to my changes

Sure, I may look into this later this week or next. Sorry that I am busy with other things now.

rwadowski:
> > @ngjaying hi, can you help me with this build [problem](https://github.com/lf-edge/ekuiper/runs/3609653807?check_suite_focus=true) ? I am not sure how it is related to my changes
> 
> Sure, I may look into this later this week or next. Sorry that I am busy with other things now.

Never mind, I fixed itself, as I told I am not expert in golang :D

internal/topo/node/shared/manager.go
```diff
@@ -0,0 +1,125 @@
...
+	if sinkConsumerChannels, exists := sinkChannels[sink]; exists {
+		if sourceChannel, exists := sinkConsumerChannels.consumers[source]; exists {
+			close(sourceChannel)
+			delete(sinkConsumerChannels.consumers, source)
+		}
+	}
+	return nil
+}
+
+func closeSink(sink string) error {
```
ngjaying, line 106: Is it supported to create source prior to sink? Additionally, if there are still sources listen to the mem topic, is it a good idea to just delete all these consumer channels?

internal/topo/node/shared/manager.go
```diff
@@ -0,0 +1,125 @@
...
+		ch: ch,
+	}
+	return s, nil
+}
+
+func GetSource() *source {
+	return &source{}
+}
+
+func getOrCreateSinkChannels(sink string) (*channels, error) {
```
ngjaying, line 52: I assume the channel maps to a "topic", not a sink. It looks like multiple sinks can share the same channel here which is expected, please double confirm. The argument name "sink" here is misleading.

internal/topo/node/sink_node.go
```diff
@@ -421,6 +422,11 @@ func doGetSink(name string, action map[string]interface{}) (api.Sink, error) {
 		s = &sink.RestSink{}
 	case "nop":
 		s = &sink.NopSink{}
+	case "shared":
```
ngjaying, line 425: Can we name it memory? Because it can be used standalone without sharing.

internal/topo/node/shared/instance.go
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package shared
+
+import "github.com/lf-edge/ekuiper/pkg/api"
+
+type instance struct {
```
ngjaying, line 19: It seems this whole file is not used.

ngjaying (commented):
Thanks Robert for another great pieces of work. Have you tested this manually because the UT does not cover much? I recommend to add some end to end tests in the test folders. Also, we need some documentations.

rwadowski:
> Thanks Robert for another great pieces of work. Have you tested this manually because the UT does not cover much? I recommend to add some end to end tests in the test folders. Also, we need some documentations.

Hi played with quiper as much as I know the platform. I just wrote python script to deploy stream, produce mqtt load and just ran ekuiper against it. I was wondering how to test id precisely because it boils down to the mapping. Are you talking about jmx files in the `test` dir ?
I will do the documentation once review will be close to finish since there are few points that needs to be figured out.

ngjaying:
> hat needs to be f



> > Thanks Robert for another great pieces of work. Have you tested this manually because the UT does not cover much? I recommend to add some end to end tests in the test folders. Also, we need some documentations.
> 
> Hi played with quiper as much as I know the platform. I just wrote python script to deploy stream, produce mqtt load and just ran ekuiper against it. I was wondering how to test id precisely because it boils down to the mapping. Are you talking about jmx files in the `test` dir ?
> I will do the documentation once review will be close to finish since there are few points that needs to be figured out.

Yes, I mean jmx in the test dir.

rwadowski:
> > hat needs to be f
> 
> > > Thanks Robert for another great pieces of work. Have you tested this manually because the UT does not cover much? I recommend to add some end to end tests in the test folders. Also, we need some documentations.
> > 
> > 
> > Hi played with quiper as much as I know the platform. I just wrote python script to deploy stream, produce mqtt load and just ran ekuiper against it. I was wondering how to test id precisely because it boils down to the mapping. Are you talking about jmx files in the `test` dir ?
> > I will do the documentation once review will be close to finish since there are few points that needs to be figured out.
> 
> Yes, I mean jmx in the test dir.

Is there any description how to tackle these files ?

ngjaying:
> Is there any description how to tackle these files ?

Yes, there is a readme.md in that `test` dir. Please check if that helps

ngjaying:
I record a video about running these fvt tests locally. Please check
https://zoom.us/rec/share/2xMYTXR5alDFwt2muH1BDDejU4Es95CvGgzVdBuuLVao7RNLxjQJ1QF54La1uA.l2uz8llMCSasBVEU

ngjaying:
> > Thanks Robert for another great pieces of work. Have you tested this manually because the UT does not cover much? I recommend to add some end to end tests in the test folders. Also, we need some documentations.
> 
> Hi played with quiper as much as I know the platform. I just wrote python script to deploy stream, produce mqtt load and just ran ekuiper against it. I was wondering how to test id precisely because it boils down to the mapping. Are you talking about jmx files in the `test` dir ?
> I will do the documentation once review will be close to finish since there are few points that needs to be figured out.

Hi Robert, it's ok if you have difficulties to create fvt test case in `test` dir, we can take it later.

rwadowski:
> > > Thanks Robert for another great pieces of work. Have you tested this manually because the UT does not cover much? I recommend to add some end to end tests in the test folders. Also, we need some documentations.
> > 
> > 
> > Hi played with quiper as much as I know the platform. I just wrote python script to deploy stream, produce mqtt load and just ran ekuiper against it. I was wondering how to test id precisely because it boils down to the mapping. Are you talking about jmx files in the `test` dir ?
> > I will do the documentation once review will be close to finish since there are few points that needs to be figured out.
> 
> Hi Robert, it's ok if you have difficulties to create fvt test case in `test` dir, we can take it later.



> > > Thanks Robert for another great pieces of work. Have you tested this manually because the UT does not cover much? I recommend to add some end to end tests in the test folders. Also, we need some documentations.
> > 
> > 
> > Hi played with quiper as much as I know the platform. I just wrote python script to deploy stream, produce mqtt load and just ran ekuiper against it. I was wondering how to test id precisely because it boils down to the mapping. Are you talking about jmx files in the `test` dir ?
> > I will do the documentation once review will be close to finish since there are few points that needs to be figured out.
> 
> Hi Robert, it's ok if you have difficulties to create fvt test case in `test` dir, we can take it later.

Hi, As for now I have been assigned to do internal Intech stuff, it will take me longer time to handle assigned tasks. Once I will be done I will go back to this ticket.

ngjaying:
> will go back to this tic

Sure Robert. Thanks very much

PR #1020, opened 2021-09-30
feat(io): Support source/sink share connection

SuperRxan:


etc/mqtt_source.yaml
```diff
@@ -1,7 +1,8 @@
 #Global MQTT configurations
 default:
   qos: 1
-  servers: [tcp://127.0.0.1:1883]
+  connectionSelector: mqtt_conf1
```
ngjaying, line 4: Need to discuss if this could be changed. This my conflict with the docker doc about setting mqtt server with env var

internal/topo/connection/client_edgex.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"fmt"
+	"github.com/edgexfoundry/go-mod-messaging/v2/messaging"
+	"github.com/edgexfoundry/go-mod-messaging/v2/pkg/types"
+	"github.com/lf-edge/ekuiper/pkg/cast"
+)
+
+func init() {
+	registerClientFactory("mqtt", func(super Client) ClientCreator {
+		return &MQTTClientCreator{super: super}
```
ngjaying, line 28: why putting mqtt ins edgex file?

internal/topo/connection/manager.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+type clientManager struct {
+	lock          sync.Mutex
+	clientFactory map[string]ClientFactoryFunc
+	clientMap     map[string]clientWrapper
+}
+
+func registerClientFactory(clientType string, creatorFunc ClientFactoryFunc) {
+	m.clientFactory[clientType] = creatorFunc
```
ngjaying, line 23: Is lock needed?

internal/topo/connection/client_factory.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+type ClientFactoryFunc func(super Client) ClientCreator
+
+type ClientCreator interface {
+	CfgValidate(props map[string]interface{}) error
+	GetClient() (interface{}, error)
+}
+
+type Client struct {
```
ngjaying, line 15: This name seems misleading, because it is just some configuration information instead of a client.

ngjaying (changes requested):

PR #1061, opened 2021-11-09
fix(plugin): install plugins with shellParas, plugin zip package must…

### LINKED ISSUE #1017: if user install plugins with shellParas,   plugin zip package must have install.sh file
**Environment**:

- eKuiper version (e.g. `1.3.0`):  1.3.1
- Hardware configuration (e.g. `lscpu`):
- OS (e.g. `cat /etc/os-release`):
- Others:

found missing install.sh in plugin package,
while user install the plugins with shellParas, the plugin install successful

expect: should return error when missing install.sh and  shellParas not nil

SuperRxan:
Close #1017  

### DIFF
--- a/internal/plugin/native/manager.go
+++ b/internal/plugin/native/manager.go
@@ -410,6 +410,17 @@ func (rr *Manager) install(t plugin2.PluginType, name, src string, shellParas []
 	}
 	defer r.Close()
 
+	haveInstallFile := false
+	for _, file := range r.File {
+		fileName := file.Name
+		if fileName == "install.sh" {
+			haveInstallFile = true
+		}
+	}
+	if len(shellParas) != 0 && !haveInstallFile {
+		return filenames, "", fmt.Errorf("have shell parameters : %s but no install.sh file", shellParas)
+	}
+
 	soPrefix := regexp.MustCompile(fmt.Sprintf(`^((%s)|(%s))(@.*)?\.so$`, name, ucFirst(name)))
 	var yamlFile, yamlPath, version string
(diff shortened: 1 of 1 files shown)


internal/plugin/native/manager.go
```diff
@@ -486,6 +490,11 @@ func (rr *Manager) install(t plugin2.PluginType, name, src string, shellParas []
 			conf.Log.Infof(`run install script:%s`, outb.String())
 			conf.Log.Infof("install %s plugin %s", plugin2.PluginTypes[t], name)
 		}
+	} else if haveInstallParas && !haveInstallFile {
```
ngjaying, line 493: Suggest to do this earlier to avoid copy then deleting files

SuperRxan: ack

internal/plugin/native/manager.go
```diff
@@ -408,6 +408,17 @@ func (rr *Manager) install(t plugin2.PluginType, name, src string, shellParas []
...
+		fileName := file.Name
+		if fileName == "install.sh" {
+			haveInstallFile = true
+		}
+	}
+	if len(shellParas) != 0 && !haveInstallFile {
+		return filenames, "", fmt.Errorf("have shell parameters : %s but no install.sh file", shellParas)
+	}
+
 	defer r.Close()
```
ngjaying, line 422: must add this before line 411

ngjaying (changes requested):

ngjaying (approved):

PR #1092, opened 2021-12-01
doc(ui): update screenshot 

lileiemq:


ngjaying (changes requested):
Please also update the CN version

ngjaying (approved):

PR #1188, opened 2022-02-24
fix(portable): fix rule hang problem when python plugin has exited

ngjaying:
Signed-off-by: Jiyong Huang <huangjy@emqx.io>

### DIFF
--- a/internal/server/rest.go
+++ b/internal/server/rest.go
@@ -1,4 +1,4 @@
-// Copyright 2021 EMQ Technologies Co., Ltd.
+// Copyright 2021-2022 EMQ Technologies Co., Ltd.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


internal/server/rest.go
```diff
@@ -283,17 +283,18 @@ func rulesHandler(w http.ResponseWriter, r *http.Request) {
...
+				err = doStartRule(rs)
+				if err != nil {
+					result = err.Error()
+				}
 			}
-		}
-
+		}()
 		w.WriteHeader(http.StatusCreated)
 		w.Write([]byte(result))
```
SuperRxan, line 299: now move these two function in go routine, if createRuleState(r) or doStartRule(rs) has error, result can not get the error message

SuperRxan (approved):

internal/server/rest.go
ngjaying: Logically the rule meta has created even when the rule cannot start correctly, so it is correct to return created. The start rule error is a runtime error which can be found by getStatus or log.

PR #1250, opened 2022-04-27
feat(native plugin): add sql sink plugin

SuperRxan:


ngjaying (changes requested):
etc/functions *.json should also update link

ngjaying (approved):

PR #1265, opened 2022-05-05
fix(integration test): add integration test for tdengine/sql

SuperRxan:


.github/workflows/run_fvt_tests.yaml
```diff
@@ -66,6 +66,14 @@ jobs:
         echo "jmeter.save.saveservice.response_data.on_error=true" >> /tmp/apache-jmeter-$JMETER_VERSION/user.properties
         wget -O /tmp/apache-jmeter-$JMETER_VERSION/lib/ext/mqtt-xmeter-1.13-jar-with-dependencies.jar https://github.com/emqx/mqtt-jmeter/raw/master/Download/v1.13.0/mqtt-xmeter-1.13-jar-with-dependencies.jar
         ln -s /tmp/apache-jmeter-$JMETER_VERSION /opt/jmeter
+    - name: install tdengine client
+      timeout-minutes: 10
+      env:
+        TD_VERSION: 2.2.0.5
```
ngjaying, line 72: Can we use newer version?

SuperRxan: how about 2.4.0.0 ? latest is 2.4.0.18 ?

SuperRxan: now using 2.4.0.18

ngjaying (approved):

PR #1299, opened 2022-06-13
fix(sql): fix sql order by alias bug

### LINKED ISSUE #1269: Not getting corect result while using the alias name in order by clause.
**Environment**:

- eKuiper version (e.g. `1.4.0`): 

I am trying to write a SQL to get the result with the desc order of count.
`Select count(*) as count, collect(Bool) as R1 from STREAM_RAND_BOOL GROUP BY Bool, TumblingWindow(ss, 10) ORDER BY count DESC
`

To reproduce we can write any query with order by in alias name.

My scenario:
CONF_KEY:
```
CONF_KEY_RAND_BOOL:
  topic: edgex/events/#/#/randomBool/Bool
  messageType: request
```
Stream:
`"create stream STREAM_RAND_BOOL() WITH (FORMAT=\"JSON\", TYPE=\"edgex\", CONF_KEY=\"CONF_KEY_RAND_BOOL\")"
`

Rule SQL:
`Select count(*) as count, collect(Bool) as R1 from STREAM_RAND_BOOL GROUP BY Bool, TumblingWindow(ss, 100) ORDER BY count DESC`

the result of this query is not correct.

SuperRxan:
Close #1269

### DIFF
--- a/internal/topo/operator/order_operator.go
+++ b/internal/topo/operator/order_operator.go
@@ -29,10 +29,10 @@ type OrderOp struct {
  *  input: *xsql.Tuple from preprocessor | xsql.WindowTuplesSet from windowOp | xsql.JoinTupleSets from joinOp
  *  output: *xsql.Tuple | xsql.WindowTuplesSet | xsql.JoinTupleSets
  */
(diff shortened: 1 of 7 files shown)


ngjaying:
Great!

PR #1303, opened 2022-06-17
Bugfix

SuperRxan:


internal/xsql/valuer.go
```diff
@@ -592,6 +596,10 @@ func (v *ValuerEval) evalBinaryExpr(expr *ast.BinaryExpr) interface{} {
 		return v.evalJsonExpr(lhs, expr.OP, expr.RHS)
 	}
 	rhs := v.Eval(expr.RHS)
+	if isSliceOrArray(rhs) {
```
ngjaying, line 599: Is it better to use the OP IN as the entry point?

ngjaying (changes requested):

ngjaying (approved):

PR #1306, opened 2022-06-23
fix(native plugin): load native plugin when set up

superrxan:


internal/plugin/native/manager.go
```diff
@@ -437,7 +442,7 @@ func (rr *Manager) install(t plugin2.PluginType, name, src string, shellParas []
 		if yamlFile == fileName {
 			err = filex.UnzipTo(file, yamlPath)
 			if err != nil {
-				return filenames, "", err
+				goto errout
```
ngjaying, line 445: Can we use defer instead of goto

PR #1311, opened 2022-06-29
fix(ui): fix default array properties is null

SuperRxan:

### DIFF
--- a/internal/meta/sinkMeta.go
+++ b/internal/meta/sinkMeta.go
@@ -118,9 +118,8 @@ func newField(fis []*fileField) (uis []field, err error) {
 			Hint:              newLanguage(fi.Hint),
 			Label:             newLanguage(fi.Label),
 		}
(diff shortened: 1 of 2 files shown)


internal/meta/sinkMeta.go
```diff
@@ -118,9 +118,8 @@ func newField(fis []*fileField) (uis []field, err error) {
 			Hint:              newLanguage(fi.Hint),
 			Label:             newLanguage(fi.Label),
 		}
-		uis = append(uis, ui)
 		switch t := fi.Default.(type) {
-		case []map[string]interface{}:
+		case []interface{}:
```
ngjaying, line 122: Is it possible not this type

SuperRxan: not possible, the fi is processed result of json

ngjaying (approved):

PR #1357, opened 2022-08-10
fix(ci): use uniform base image to build ekuiper and plugins

SuperRxan:


deploy/docker/Dockerfile-slim-python
```diff
@@ -12,13 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.18 AS builder
+ARG GO_VERSION=1.18.5
+
+FROM ghcr.io/lf-edge/ekuiper/base:$GO_VERSION-debian AS builder
```
ngjaying, line 17: 1. Is there already a kuiperd inside this base image? Can we just copy it?
2. If the external user want to build a docker image, can they get this dependant base image?

SuperRxan: 1. The base image do not build kuiperd but have source code for kuiper, I think we can remove these lines
```
COPY . /go/kuiper

WORKDIR /go/kuiper
```

ci should copy the most new ekuiper source code into build environment, copy the source code for base image do not make sense.

```

FROM debian:stretch

ARG GO_VERSION=1.18.5

RUN apt update \
    && apt install -y wget git zip upx pkg-config libczmq-dev build-essential debhelper jq zip

RUN case $(dpkg --print-architecture) in \
        arm|armhf|armv7) \
            ARCH=armv6l \
        ;; \
        *) \
            ARCH=$(dpkg --print-architecture)\
        ;; \
    esac \
    && wget --no-verbose https://golang.org/dl/go$GO_VERSION.linux-$ARCH.tar.gz \
    && tar -C /usr/local -xzf go$GO_VERSION.linux-$ARCH.tar.gz

ENV PATH=$PATH:/usr/local/go/bin

COPY . /go/kuiper

WORKDIR /go/kuiper

ENV KUIPER_SOURCE=/go/kuiper

```

2. yes, they can get that base image

ngjaying (approved):

Rory-Z (approved):

PR #1392, opened 2022-09-05
fix(plugin): move plugin meta data out of etc folder

superrxan:


ngjaying (commented):
Do we need to update the build CI?

ngjaying (approved):

PR #1393, opened 2022-09-06
Graph API updates

ngjaying:


superrxan (commented):
do we need add rest api to get these meta data ?

ngjaying:
> do we need add rest api to get these meta data ?

Yes, but I'll just leave it in a next PR

superrxan (approved):

PR #1394, opened 2022-09-06
fix(plugin): check whether dataTemplate is an empty field

elpsyr:
Signed-off-by: crui <crui@relaper.com>

### DIFF
--- a/extensions/sinks/influx/influx.go
+++ b/extensions/sinks/influx/influx.go
@@ -95,7 +95,7 @@ func (m *influxSink) Configure(props map[string]interface{}) error {
 		}
 	}
 	if i, ok := props["dataTemplate"]; ok {
-		if i.(string) != "" {
+		if i, ok := i.(string); ok && i != "" {
 			m.hasTransform = true
 		}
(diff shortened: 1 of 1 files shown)


extensions/sinks/influx/influx.go
```diff
@@ -76,26 +76,28 @@ func (m *influxSink) Configure(props map[string]interface{}) error {
...
 	}
 	if i, ok := props["fields"]; ok {
 		if i, ok := i.(string); ok {
 			m.fields = i
 		}
 	}
-	if _, ok := props["dataTemplate"]; ok {
-		m.hasTransform = true
+	if i, ok := props["dataTemplate"]; ok {
+		if i.(string) != "" {
```
ngjaying, line 98: i.(string) may panic if it is not a string, shall we add the check also?

elpsyr: @ngjaying You are right. We need to judge whether this is a string first.I will perfect it.

ngjaying (approved):
Thank you!

PR #1421, opened 2022-09-29
Make redis sink updatable

ngjaying:


superrxan (commented):
now redis is built-in plugin, redis.json should move to ect/sinks folder

superrxan (approved):

PR #1427, opened 2022-10-08
fix(etc): move etc files into data

superrxan:


internal/plugin/native/manager.go
```diff
@@ -69,15 +69,15 @@ func InitManager() (*Manager, error) {
 	if err != nil {
 		return nil, fmt.Errorf("cannot find plugins folder: %s", err)
 	}
-	etcDir, err := conf.GetConfLoc()
+	dataDir, err := conf.GetDataLoc()
 	if err != nil {
 		return nil, fmt.Errorf("cannot find etc folder: %s", err)
```
ngjaying, line 74: Not etc now, update the error message?

internal/plugin/portable/manager.go
```diff
@@ -49,7 +49,7 @@ func InitManager() (*Manager, error) {
 	if err != nil {
 		return nil, fmt.Errorf("cannot find plugins folder: %s", err)
 	}
-	etcDir, err := conf.GetConfLoc()
+	dataDir, err := conf.GetDataLoc()
 	if err != nil {
 		return nil, fmt.Errorf("cannot find etc folder: %s", err)
```
ngjaying, line 54: The same problem in the error message

ngjaying (approved):

PR #1442, opened 2022-10-18
fix(fea): support sink config by resource id

superrxan:

### DIFF
--- a/internal/conf/yaml_config_ops.go
+++ b/internal/conf/yaml_config_ops.go
@@ -230,11 +230,10 @@ func (c *ConfigKeys) AddConfKey(confKey string, reqField map[string]interface{})
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	if _, ok := c.etcCfg[confKey]; ok {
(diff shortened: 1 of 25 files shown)


internal/conf/yaml_config_ops.go
```diff
@@ -230,12 +230,11 @@ func (c *ConfigKeys) AddConfKey(confKey string, reqField map[string]interface{})
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
+	c.dataCfg[confKey] = reqField
 	if _, ok := c.etcCfg[confKey]; ok {
-		return fmt.Errorf("duplicate key %s, already exist in etc folder", confKey)
+		delete(c.etcCfg, confKey)
```
ngjaying, line 235: Assume dataCfg have higher priority to override etcCfg. Is it better to leave it, since the etc file is not changed?

superrxan: etcCfg just load from disk and save in memory, this delete operation just happen in memory and do not save to disk

main use case is user edit the config key in etc folder and save it. we save the configs in dataCfg and save to disk.

ngjaying (approved):

PR #1454, opened 2022-10-25
fix(rule): do not recover rule that has error

superrxan:


internal/plugin/native/manager.go
```diff
@@ -628,7 +628,6 @@ func (rr *Manager) loadRuntime(t plugin2.PluginType, soName, soFilepath, symbolN
 		} else {
 			mod, err := rr.getSoFilePath(t, soName, false)
 			if err != nil {
-				conf.Log.Warnf(fmt.Sprintf("cannot find the native plugin %s in path: %v", soName, err))
```
ngjaying, line 631: Will this generate a lot of logs especially for function plugins? I suggest to use Debugf here

ngjaying (approved):

PR #1478, opened 2022-11-03
fix(rest): fix restapi plugins/sources/prebuild and plugins/sources/{name} conflict issue

superrxan:


ngjaying (commented):
If plugin build tag is not enabled, the meta APIs will all fail?

ngjaying (approved):

PR #1509, opened 2022-11-18
fix(rest): let rule name empty if not set

superrxan:


ngjaying (commented):
This is an intended behavior to avoid no name rule. Suggest not to change this.

PR #1513, opened 2022-11-18
fea(plug): add general tensorflow function

superrxan:


extensions/functions/tensorFlowGeneral/resize.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2022 EMQ Technologies Co., Ltd.
```
ngjaying, line 1: Try to merge this to the original resize function or remove from this PR. Let the resize function to have args of (x,y,z, channels, format)

extensions/functions/tensorFlowGeneral/install.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
```
ngjaying, line 1: In the etc tflite example models, can we add a license file from the original location?

extensions/functions/tensorFlowGeneral/tflite.go
```diff
@@ -0,0 +1,259 @@
...
+				}
+				err = input.SetUint8s(v)
+				if err != nil {
+					return nil, false
+				}
+			default:
+				return fmt.Errorf("invalid %d parameter, unsupported type %v in the model", i, t), false
+			}
+		default:
+			// TODO support multiple dimensions. Here assume user passes a 1D array.
```
ngjaying, line 156: Here add a validation for the input argument like the first case

ngjaying (changes requested):

extensions/functions/tensorFlowGeneral/resize.go
superrxan: ack

extensions/functions/tensorFlowGeneral/tflite.go
ngjaying: wrong validation, len(arg) == dim(1) * dim(2) * dim(n)

ngjaying (changes requested):

ngjaying (approved):

PR #1563, opened 2022-12-16
fix(build): add git for python runtime image

superrxan:


internal/server/rule_manager.go
```diff
@@ -135,7 +135,7 @@ func getAllRulesWithStatus() ([]map[string]interface{}, error) {
 	sort.Strings(ruleIds)
 	result := make([]map[string]interface{}, len(ruleIds))
 	for i, id := range ruleIds {
-		ruleName := id
+		ruleName := ""
```
ngjaying, line 138: why put this here?

ngjaying (changes requested):

internal/server/rule_manager.go
superrxan: removed

ngjaying (approved):

PR #1581, opened 2022-12-27
fix(doc): add full data export/import docs

superrxan:

### DIFF
--- a/docs/directory.json
+++ b/docs/directory.json
@@ -511,6 +511,10 @@
 						{
 							"title": "规则集管理",
 							"path": "operation/restapi/ruleset"
+						},
+						{
+							"title": "数据导入导出",
+							"path": "operation/restapi/configuration"
 						}
 					]
 				},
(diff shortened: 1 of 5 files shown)


ngjaying (changes requested):
Suggestion: Name it as: Data import/export; 数据导入导出

docs/directory.json
```diff
@@ -511,6 +511,10 @@
 						{
 							"title": "规则集管理",
 							"path": "operation/restapi/ruleset"
+						},
+						{
+							"title": "配置管理",
```
ngjaying, line 516: missing here

docs/directory.json
```diff
@@ -1204,6 +1212,10 @@
 						{
 							"title": "Ruleset",
 							"path": "api/restapi/ruleset"
+						},
+						{
+							"title": "Configuration export/import",
```
ngjaying, line 1217: Still found `Configuration`.  Please use upper case like Data Export/Import

ngjaying (changes requested):
Suggest to rename the files as well, like data.md

ngjaying (approved):

PR #1615, opened 2023-01-18
chore: add go.work

logo306142054:
Signed-off-by: chenhongxin <306142054@qq.com>

### DIFF
--- a/go.work
+++ b/go.work
@@ -1,0 +1,3 @@
+go 1.18
+
+use ./tools/kubernetes/


ngjaying:
It seems the workspace file has conflict with the current plugin modules. We'll need to investigate more.

ngjaying:
Close this as it cannot pass UT. Let's create another PR once we can adapt to this workspace feature

PR #1664, opened 2023-03-10
fix(config): add api to get install script

superrxan:


internal/binder/factory.go
```diff
@@ -14,15 +14,20 @@
...
-import "github.com/lf-edge/ekuiper/pkg/api"
+import (
+	"github.com/lf-edge/ekuiper/internal/plugin"
+	"github.com/lf-edge/ekuiper/pkg/api"
+)
 
 type SourceFactory interface {
 	Source(name string) (api.Source, error)
 	LookupSource(name string) (api.LookupSource, error)
+	GetSourcePlugin(name string) (plugin.EXTENSION_TYPE, string, string)
```
ngjaying, line 25: Add comment for the param and return parameters, I don't know what are they. The same for other interface changes.

ngjaying (commented):
Suggest to add more comments

Issue #1462, opened 2022-10-27
Function to return the first non null value


ngjaying:
We may need a COALESCE function like pgsql to return the first non null value. So COALESCE(col, 0) will return 0 if column col is null.

PR #1669, opened 2023-03-14
feat:  add coalesce function

Lieoxc:
feat(function) :
1. add coalesce function to return the first non null value.
2. add coalesce function instructions

close:  #1462


ngjaying (changes requested):
Awesome! Thanks for the great contribution. However, I suggest to move this function to another category "other" in fike funcs_misc.go. Because we define analytic function to be stateful function which use state API inside and there are some special handling in the parser side. Apparently, this function is not stateful.

Also, please follow the details link in the DCO check to pass it.

PR #1670, opened 2023-03-14
feat: add coalesce function

Lieoxc:
feat(function) :
1. add coalesce function to return the first non null value.
2. add coalesce function instructions

close:  lf-edge#1462

Signed-off-by: Lieoxc <xiaocongli45@gmail.com>


ngjaying (approved):
Perfect!

PR #1690, opened 2023-03-23
fix (function): make aggregate functions return the same value as the SQL for empty Windows

### LINKED ISSUE #1682: Consistent aggregate function return value for empty window
Optimization Function Print Log （max、min）

Rui-Gan:
Closes #1682


superrxan (approved):

internal/binder/function/funcs_agg.go
```diff
@@ -487,7 +509,7 @@ func dedup(r []interface{}, col []interface{}, all bool) (interface{}, error) {
 	}
 	if !all {
 		if len(result) == 0 {
-			return nil, nil
+			return []interface{}{}, nil
```
ngjaying, line 512: The second argument 'all' means whether to reutrn the full array or just a single event. Here, we need to return a single event -> nil

ngjaying (changes requested):

ngjaying (approved):

Issue #1689, opened 2023-03-23
Adding additional compression methods


ngjaying:
We provide gzip compression/decompression ablity by compress/decompress function, MQTT source and sink. 

**What would you like to be added/modified**:
Add more compression method, like gzip and bzip2. Both are actually supported by golang.

**Why is this needed**:
Provide more options when using compression feature.

ngjaying:
Just add the compressor and decompressor struct into internal/compressor package

PR #1697, opened 2023-03-27
feat(func): add gzip, flate compressor/decompressor

carlclone:


internal/compressor/flate.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	return g.buffer.Bytes(), nil
+}
+
+func newFlateDecompressor() (*flateDecompressor, error) {
+	return &flateDecompressor{}, nil
+}
+
+type flateDecompressor struct {
+	reader io.ReadCloser
```
ngjaying, line 45: In the https://pkg.go.dev/compress/flate reset example, it is possible to init a reader with nil data, I suggest to follow the code style there. In zlib, NewReaer(nil) has an error, that's why I coded like this.

ngjaying (commented):
Please check the built binary size to see if it expands too much after importing these new dependencies. Run `Make` and check the bin/kuiperd size.

internal/compressor/flate.go
carlclone: got it , and why is this test failed ? 

```
mock sink receive  map[email:Dddd id:1 name:test]
--- FAIL: TestFormat_Apply (5.51s)
    sink_node_test.go:315: 1 	result mismatch:
        
        exp=[0a047465737410011a0444646464]
        
        got=[]
        
The test bucket size is 6.
```

carlclone: > In the https://pkg.go.dev/compress/flate reset example, it is possible to init a reader with nil data, I suggest to follow the code style there. In zlib, NewReaer(nil) has an error, that's why I coded like this.

code modified , and the size is kuiper 7.6M and kuiperd 28.6M  , same as before

ngjaying: > got it , and why is this test failed ?
> 
> ```
> mock sink receive  map[email:Dddd id:1 name:test]
> --- FAIL: TestFormat_Apply (5.51s)
>     sink_node_test.go:315: 1 	result mismatch:
>         
>         exp=[0a047465737410011a0444646464]
>         
>         got=[]
>         
> The test bucket size is 6.
> ```

Ah, it is a problem of the test case which is synced by time.sleep now.. Not your problem

ngjaying (approved):

carlclone:
will do

ngjaying:
Closes #1689

PR #1701, opened 2023-03-28
feat(source): add ruleid to metadata if ruleid exist

carlclone:
Signed-off-by: carlclone <906561974@qq.com>
(cherry picked from commit 749a2d7766ec5a731c1869eab5b0e63fd0f8643a)

Do you think these modification is acceptable ? if so I will continue to develop in this direction


ngjaying:
This may be possible to meet the requirement, but it needs to update for every source types. I would suggest to do it in the meta function implementation `internal\binder\function\funcs_misc.go`. Currently, the meta function is specially treated in the parser and parsed to MetaRef expression. And the evaluation is in `internal\xsql\valuer.go` line 415. We may need to make them work together for newly defined comman metas.

carlclone:
code updated , grateful for your review

PR #1705, opened 2023-03-29
feat(func): add rule_id func

### LINKED ISSUE #1455: Is it possible to attach rule id to analysis result
**Please describe your problem in detail, if necessary, you can upload the log file through the attachment**:
After rule engine detect the abnormal situation, I want the output message includes which rule it is from, is it possible now?

carlclone:
closes #1455


carlclone:
The code gets a bit ugly when trying to support meta(*) in this feature , maybe just support meta(ruleid) ? or any suggestion ? thanks

ngjaying:
> The code gets a bit ugly when trying to support meta(*) in this feature , maybe just support meta(ruleid) ? or any suggestion ? thanks

Good point. How about just create another function like rule_meta()

carlclone:
Sounds good. I already updated the code, and I think there is no other rule info we can get inside the built-in function , maybe  just call it rule_id() ?

ngjaying:
> Sounds good. I already updated the code, and I think there is no other rule info we can get inside the built-in function , maybe just call it rule_id() ?

Great. No problem. Let's wait for the checks. I may merge this after 1.9.0 release.

etc/functions/internal.json
```diff
@@ -3387,6 +3387,29 @@
...
+			"hint": {
+				"en_US": "ID of currently matched rule",
+				"zh_CN": "当前匹配到的规则的ID"
+			}
+		},
+		"node": {
+			"category": "function",
+			"icon": "iconPath",
+			"label": {
+				"en_US": "ID of currently matched rule",
```
ngjaying, line 3409: Label is the word to show in UI, should be short. Suggest "Rule ID"

ngjaying (changes requested):

etc/functions/internal.json
carlclone: done

ngjaying (approved):

Issue #1703, opened 2023-03-29
Optimization: Return data information for  REST sink


ngjaying:
Currently, when an exception happens in REST sink such as timeout, users don't have the processing data information in the message to diagnose. They may need to open debug mode and rerun to diagnose
![ee6216abcd711500f2f03814f8adb52](https://user-images.githubusercontent.com/792416/228398397-ea8f94ff-d198-42e3-bc7e-87291e228cdf.png)

**What would you like to be added/modified**:
Print out the full http request information when errors happen.
Also, this may apply to other sinks.

**Why is this needed**:
Easier to diagnose

PR #1713, opened 2023-03-31
feat(sink): add http request/response body to rest sink log when exception occurred

carlclone:


internal/io/http/rest_sink.go
```diff
@@ -56,14 +56,33 @@ func (me MultiErrors) Error() string {
...
+	decodedData, _, err := ctx.TransformOutput(item)
 	if err != nil {
-		return fmt.Errorf("rest sink fails to send out the data: %s", err)
+		logger.Warnf("rest sink decode data error: %v", err)
+		return fmt.Errorf("rest sink decode data error: %v", err)
+	}
+
+	resp, err := ms.Send(ctx, decodedData, item, logger)
+	if err != nil {
+		return fmt.Errorf(`%s | method=%s path="%s" status=%d request_body="%s"`,
```
ngjaying, line 67: Sprintf is not needed, Errorf can do the formation

internal/io/http/rest_sink.go
```diff
@@ -56,14 +56,33 @@ func (me MultiErrors) Error() string {
 func (ms *RestSink) Collect(ctx api.StreamContext, item interface{}) error {
 	logger := ctx.GetLogger()
 	logger.Debugf("rest sink receive %s", item)
-	resp, err := ms.Send(ctx, item, logger)
+	decodedData, _, err := ctx.TransformOutput(item)
```
ngjaying, line 59: Move this out may have problem if "Send" is used elsewhere

ngjaying (changes requested):

internal/io/http/rest_sink.go
carlclone: i searched the code and did not see other usage , can we make it a private function ?

ngjaying: That's fine if no one use it

internal/io/http/rest_sink.go
```diff
@@ -56,14 +56,33 @@ func (me MultiErrors) Error() string {
...
+		return fmt.Errorf("rest sink decode data error: %v", err)
+	}
+
+	resp, err := ms.Send(ctx, decodedData, item, logger)
+	if err != nil {
+		return fmt.Errorf(`rest sink fails to send out the data: %s | method=%s path="%s" status=%d request_body="%s"`,
+			err,
+			ms.config.Method,
+			ms.config.Url,
+			-1,
```
ngjaying, line 71: what does "-1" mean

carlclone: sorry ,  i'll deleted it

ngjaying (approved):

ngjaying:
Thank you!

ngjaying:
Closes #1703

PR #1715, opened 2023-03-31
fix(redis): upgrade redis client from v7 to v9

superrxan:


go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/edgexfoundry/go-mod-messaging/v2 v2.3.0
 	github.com/fxamacker/cbor/v2 v2.4.0
 	github.com/gdexlab/go-render v1.0.1
-	github.com/go-redis/redis/v7 v7.4.1
+	github.com/go-redis/redis/v7 v7.3.0
```
ngjaying, line 15: still have v7 dependency

ngjaying (changes requested):

go.mod
superrxan: resolved

ngjaying (approved):

PR #1717, opened 2023-04-03
chore: upgrade go version to 1.20

Rui-Gan:


test/redis/set.go
```diff
@@ -1,35 +1,38 @@
 package main
 
 import (
+	"context"
 	"fmt"
-	"github.com/go-redis/redis/v7"
```
superrxan, line 5: please do not put redis related commit in this PR

internal/topo/node/operations.go
```diff
@@ -128,6 +128,7 @@ func (o *UnaryOperator) doOp(ctx api.StreamContext, errCh chan<- error) {
 			}
 			stats.IncTotalRecordsIn()
 			stats.ProcessTimeStart()
+			//it := deepcopy.Copy(item)
```
superrxan, line 131: remove this

PR #1726, opened 2023-04-04
fix(sink): shadowed variable cause the inner err unhandled

carlclone:


ngjaying (commented):
Do you think it will be better to just return the error explicitly instead of setting the return variable name?

ngjaying:
> > Do you think it will be better to just return the error explicitly instead of setting the return variable name?
> 
> thanks for advice , i push a new version , is this ok ?

You're welcome. Looks good.
Just curious, how do you find this bug

carlclone:
ok

ngjaying:
@carlclone Just to clarify. It is optional to add your name to the license statement. It is okay to put your Github id instead of your name as it is just to give you credit for this project. Thanks for your great contributions

carlclone:
> @carlclone Just to clarify. It is optional to add your name to the license statement. It is okay to put your Github id instead of your name as it is just to give you credit for this project. Thanks for your great contributions

but this file is not created by me , i wil add copyright in my kafka sink file , thanks

ngjaying (approved):

PR #1728, opened 2023-04-04
chore: add check for go tidy

Yisaer:


Rory-Z:
Why we need this ?
I think we can running `go mod tidy && git diff .` is ok

Yisaer:
> Why we need this ? I think we can running `go mod tidy && git diff .` is ok

We can use it to guarantee mod tidy if we integrate it into CI check progress

Rory-Z:
> > Why we need this ? I think we can running `go mod tidy && git diff .` is ok
> 
> We can use it to guarantee mod tidy if we integrate it into CI check progress later

Looks this just show diff of go.sum, so why not use `git diff`

Yisaer:
> > > Why we need this ? I think we can running `go mod tidy && git diff .` is ok
> > 
> > 
> > We can use it to guarantee mod tidy if we integrate it into CI check progress later
> 
> Looks this just show diff of go.sum, so why not use `git diff`

That's also ok to me, the main target is to add tidy check into `Makefile` (then we can aggregate other static check into one command), the implementation doesn't make matter.

Rory-Z:
> That's also ok to me, the main target is to add tidy check into `Makefile` (then we can aggregate other static check into one command), the implementation doesn't make matter.

Maybe we can add `go mod tidy && git diff .` to Makefile, doesn't need script file ?

Yisaer:
> > That's also ok to me, the main target is to add tidy check into `Makefile` (then we can aggregate other static check into one command), the implementation doesn't make matter.
> 
> Maybe we can add `go mod tidy && git diff .` to Makefile, doesn't need script file ?

updated.

Rory-Z (approved):

PR #1735, opened 2023-04-06
ci: introduce revive check into ci

Yisaer:


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v3
+
+      - name: Run Revive Action
+        uses: docker://morphy/revive-action:v2
```
Rory-Z, line 35: Why not use https://github.com/marketplace/actions/revive-action ?

Rory-Z (approved):

PR #1736, opened 2023-04-06
upgrade go-mod-messaging to v3

Rui-Gan:


etc/connections/connection.yaml
```diff
@@ -55,11 +55,11 @@ edgex:
     optional:
       KeepAlive: "50"
 
-  zeroMsgBus: #connection key
-    protocol: tcp
-    server: localhost
-    port: 5571
-    type: zero
+#  zeroMsgBus: #connection key
```
ngjaying, line 58: I prefer to just remove these confs. The same for other confs

test/edgex/benchmark/pub.go
```diff
@@ -22,24 +22,25 @@ import (
...
+
+	v2 "github.com/edgexfoundry/go-mod-core-contracts/v2/common"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/dtos"
+	"github.com/edgexfoundry/go-mod-messaging/v3/messaging"
+	"github.com/edgexfoundry/go-mod-messaging/v3/pkg/types"
 )
 
 var msgConfig1 = types.MessageBusConfig{
-	PublishHost: types.HostInfo{
+	Broker: types.HostInfo{
```
ngjaying, line 38: Is this possible to connect to edgeX v2?

ngjaying (changes requested):

PR #1742, opened 2023-04-07
style: enable superfluous-else check and fix

Yisaer:


internal/topo/node/sink_node.go
```diff
@@ -157,11 +157,12 @@ func (m *SinkNode) Open(ctx api.StreamContext, result chan<- error) {
 							for {
 								select {
 								case data := <-m.input:
-									if temp, processed := m.preprocess(data); processed {
+									var temp interface{}
```
ngjaying, line 160: Actually, I prefer the previous style lol. What's the advantage of this link?

Yisaer: `superfluous-else` removes the redundant `else` in the code.

What about revise code in this way:

```golang
if temp, processed := m.preprocess(data); processed {
    data = temp
} else {
    break
}
```

ngjaying: Looks better.

ngjaying (approved):

PR #1743, opened 2023-04-07
feat(func): change compress implementation to `github.com/klauspost/compress` , add zstd compressor/decompressor

carlclone:
klauspost/compress offers better performance and smoother compress level than go standard library and has some extra compress algorithm . it already used by segmentio/kafka-go and pingcap/tidb  , here is some test result i run 

go std library

compress speed
image

decompress speed
image

ratio
image

klaupost

compress speed
image
decompress speed
image

ratio
image


ngjaying:
It seems compression speed is faster but compression ratio decrease? Will it increase the binary size a lot?

carlclone:
> It seems compression speed is faster but compression ratio decrease? Will it increase the binary size a lot?

kuiper+kuiperd total size from 39MB to 40MB 

ratio maybe cause by the compress level , the author smooth the level from 1 to 9 compare to standard library , maybe we can change it to higher compress level to gain same ratio and better speed , but the ratio seems pretty close for now

ngjaying:
> > It seems compression speed is faster but compression ratio decrease? Will it increase the binary size a lot?
> 
> kuiper+kuiperd total size from 39MB to 40MB
> 
> ratio maybe cause by the compress level , the author smooth the level from 1 to 9 compare to standard library , maybe we can change it to higher compress level to gain same ratio and better speed , but the ratio seems pretty close for now

I only concern about kuiperd. Will it be smaller to remove ztsd? Maybe we should just provide the basic compression method as built-in to keep it small.

carlclone:
@ngjaying  
31.2 -> 32.1  with zstd
31.2 -> 31.5  without zstd

ngjaying:
> @ngjaying 31.2 -> 32.1 with zstd 31.2 -> 31.5 without zstd
> 
> should i remove it ? this algorithm works pretty well thogh .
> 
> image
> 
> image

Ah, we are getting bigger...
Thanks for the great work! I think we can merge this. We can decide to provide this in standard package or not later before releasing. We can control it by build tag.

internal/binder/function/funcs_stateful_test.go
```diff
@@ -49,13 +49,13 @@ func TestCompressExec(t *testing.T) {
 				"hello world",
 				"zlib",
 			},
-			result: []byte{120, 156, 202, 72, 205, 201, 201, 87, 40, 207, 47, 202, 73, 1, 4, 0, 0, 255, 255, 26, 11, 4, 93},
+			result: []byte{120, 156, 0, 11, 0, 244, 255, 104, 101, 108, 108, 111, 32, 119, 111, 114, 108, 100, 3, 0, 26, 11, 4, 93},
```
ngjaying, line 52: Is this decompressible by other lib or even languages like nodejs?

docs/en_US/guide/sinks/builtin/mqtt.md
```diff
@@ -16,7 +16,7 @@ The action is used for publish output message into an MQTT server.
...
 | retained           | true     | If retained is `true`,The broker stores the last retained message and the corresponding QoS for that topic.The default value is `false`.                                                                                                                                                                                                                  |
-| compression        | true     | Compress the payload with the specified compression method. Support `zlib`, `gzip`, `flate` method now.                                                                                                                                                                                                                                                   |
+| compression        | true     | Compress the payload with the specified compression method. Support `zlib`, `gzip`, `flate`, `zstd` method now.                                                                                                                                                                                                                                           |
```
ngjaying, line 19: please search docs and etc for zlib, it is possible we need to add zstd in those places.

ngjaying (changes requested):
Need a general update: Add license statement for newly added files, and update the copyright year for existing files.

internal/binder/function/funcs_stateful_test.go
carlclone: yes , i tested it 
image

docs/en_US/guide/sinks/builtin/mqtt.md
carlclone: done

ngjaying:
Thank you. LGTM.

ngjaying (approved):

carlclone:
> Thank you. LGTM. By the way, you may be interested on this feature #1746 just planned. Feel free to discuss.

got it.

PR #1749, opened 2023-04-10
feat: introduce cgo free sqlite

Yisaer:
Introduce https://github.com/logoove/sqlite as the cgo-free sqlite which based on the https://gitlab.com/cznic/sqlite with several convenient fixing.


internal/pkg/store/sql/sqlite/cgo_free_sqllite.go
```diff
@@ -0,0 +1,85 @@
+//go:build !cgo
```
ngjaying, line 1: I was trying to replace the old driver. Are there any reason to keep the cgo version?

internal/pkg/store/sql/sqlite/cgo_free_sqllite.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"os"
+	"path"
+	"sync"
+
+	"github.com/lf-edge/ekuiper/internal/conf"
+	"github.com/lf-edge/ekuiper/internal/pkg/store/definition"
+
+	// introduce sqlite
+	_ "github.com/logoove/sqlite"
```
ngjaying, line 32: Why using this?

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/keepeye/logrus-filename v0.0.0-20190711075016-ce01a4391dd1
 	github.com/klauspost/compress v1.15.11
 	github.com/lestrrat-go/file-rotatelogs v2.4.0+incompatible
+	github.com/logoove/sqlite v1.15.3
 	github.com/mattn/go-sqlite3 v1.14.16
```
ngjaying, line 26: Aim to remove mattn/go-sqlite dependency

internal/pkg/store/sql/sqlite/cgo_free_sqllite.go
Yisaer: `czniz/sqlite` use `sqlite` as driver name, which is different from the original sqlite driver and will cause problems if we use `dburl` to parse url. This one forked the `czniz/sqlite` and fixed this problem.

internal/pkg/store/sql/sqlite/database.go
```diff
@@ -1,10 +1,10 @@
-// Copyright 2022-2022 EMQ Technologies Co., Ltd.
+// Copyright 2023-2023 EMQ Technologies Co., Ltd.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
+// http://www.apache.org/licenses/LICENSE-2.0
```
ngjaying, line 7: Here, can we revert? Otherwise, this will cause meaningless diff if others change this file.

ngjaying (changes requested):
Some small problem in copyright statement

ngjaying (approved):
cool, LGTM

PR #1750, opened 2023-04-10
feat: support expression for array index in SelectStmt

Yisaer:


internal/xsql/parser_test.go
```diff
@@ -35,6 +35,41 @@ func TestParser_ParseStatement(t *testing.T) {
 		stmt *ast.SelectStatement
 		err  string
 	}{
+		{
+			s: "SELECT arr[x+1:y] FROM tbl",
```
ngjaying, line 39: Shall we add more combinations to test, like arr[x], arr[1:x], arr[x:4]

PR #1751, opened 2023-04-10
feat(sink): kafka sink plugin

carlclone:


extensions/sinks/kafka/kafka.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	return nil
+}
+
+func (m *kafkaSink) kafkaWriteWithBackoff(ctx context.Context, log func(string, ...interface{}), interval, maxInterval time.Duration, messages ...kafkago.Message) error {
+	var berr error
+	tries := 0
+
+retry:
```
ngjaying, line 155: In previous kafkago.Writer initialization, there are properties like `MaxAttempts`, `WriteTimeout`. It looks like you are implementing retry and timeout mechanism in this function. Why do you need this if the Writer already has that?

ngjaying (commented):
Thanks for the great work. I have some questions about the implementation. Additionally, the doc and plugin build script (search other plugin like influx for example) are needed

extensions/sinks/kafka/kafka.go
carlclone: according to the official doc of segmentio/kafka-go , there is some extra error like LeaderNotAvailable which is not related to message sending , can be handle by this kind of code

https://github.com/segmentio/kafka-go
```
var err error
const retries = 3
for i := 0; i < retries; i++ {
    ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
    defer cancel()
    
    // attempt to create topic prior to publishing the message
    err = w.WriteMessages(ctx, messages...)
    if errors.Is(err, LeaderNotAvailable) || errors.Is(err, context.DeadlineExceeded) {
        time.Sleep(time.Millisecond * 250)
        continue
    }

    if err != nil {
        log.Fatalf("unexpected error %v", err)
    }
}

if err := w.Close(); err != nil {
    log.Fatal("failed to close writer:", err)
}
```

extensions/sinks/kafka/kafka.go
```diff
@@ -0,0 +1,221 @@
...
+	logger := ctx.GetLogger()
+	logger.Debugf("kafka sink receive %s", item)
+	var decodedBytes []byte
+	var err error
+	decodedBytes, _, err = ctx.TransformOutput(item)
+	if err != nil {
+		return fmt.Errorf("kafka sink transform data error: %v", err)
+	}
+
+	vctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
ngjaying, line 141: Is this timeout needed as the kafkaWriteWithBackoff already has an implicit timeout

PR #1752, opened 2023-04-10
fea(state): support get_keyed_state function to get data in redis

superrxan:


internal/binder/function/funcs_misc.go
```diff
@@ -627,6 +628,123 @@ func registerMiscFunc() {
...
+			if !ok {
+				return fmt.Errorf("key %v is not a string", args[0]), false
+			}
+
+			value, err := keyedstate.GetKeyedState(key)
+			if err != nil {
+				return args[2], true
+			}
+
+			if v, ok := args[1].(string); ok {
```
ngjaying, line 647: why not use cast package to do the conversion?

internal/binder/function/funcs_misc.go
```diff
@@ -627,6 +628,123 @@ func registerMiscFunc() {
...
+				}
+			} else {
+				return fmt.Errorf("expect string type for the 2nd parameter"), false
+			}
+		},
+		val: func(_ api.FunctionContext, args []ast.Expr) error {
+			if len(args) != 3 {
+				return fmt.Errorf("the args must be three")
+			}
+			if !ast.IsStringArg(args[0]) {
```
ngjaying, line 738: This only checks for literal. Refer to other validator functions

internal/keyedstate/kv.go
```diff
@@ -0,0 +1,46 @@
...
+package keyedstate
+
+import (
+	"github.com/lf-edge/ekuiper/internal/pkg/store"
+	kv2 "github.com/lf-edge/ekuiper/pkg/kv"
+)
+
+var manager *Manager
+
+type Manager struct {
```
ngjaying, line 24: Looks like manager == kv, why split to two levels

internal/keyedstate/kv.go
```diff
@@ -0,0 +1,46 @@
...
+	kv kv2.KeyValue
+}
+
+func InitManager() {
+	kv, _ := store.GetKV("keyed_state")
```
ngjaying, line 29: This will limit the keys in redis to have a prefix, seems wrong to me

ngjaying (changes requested):
Some problems

internal/binder/function/funcs_misc.go
superrxan: ack

internal/binder/function/funcs_misc.go
superrxan: ack

ngjaying (approved):

PR #1770, opened 2023-04-13
fix: fix build-plugins

Rui-Gan:


ngjaying (commented):
In zmq plugin, need to add an install script to install the dependencies.

ngjaying (approved):

ngjaying (changes requested):
Because different system requires different installation for zmq, it is hard to only change build-plugins.sh. Need to revert the previous update to base images

PR #1783, opened 2023-04-15
feat(sink): fix kafka sink plugin write delay , refactor to IOErr ret…

### LINKED ISSUE #1778: kafka sink take more than one second to send out message
**Environment**:  debian

- eKuiper version (e.g. `1.3.0`):   lfedge/ekuiper:1.10.0-alpha.0   
- Hardware configuration (e.g. `lscpu`): 
- OS (e.g. `cat /etc/os-release`): linux
- Others:

What happened:
kafka sink take more than one second to send out message
Another minor problem, when delivering message fails, if the error is recoverable, we should return error message starts with IOError, so that the sink cache can take effect. Check #1767 as an example.

What you expected to happen:
kafka sink should take less than 100 ms to send out message

use kafka sink to send data into kafka broker
check the metrics for rule

image

carlclone:
…ry mechanism

Closes #1778

### DIFF
--- a/extensions/sinks/kafka/kafka.go
+++ b/extensions/sinks/kafka/kafka.go
@@ -15,16 +15,15 @@
 package main
 
 import (
-	"context"
 	"fmt"
 	"github.com/lf-edge/ekuiper/pkg/api"
 	"github.com/lf-edge/ekuiper/pkg/cast"
+	"github.com/lf-edge/ekuiper/pkg/errorx"
 	kafkago "github.com/segmentio/kafka-go"
 	"github.com/segmentio/kafka-go/sasl"
 	"github.com/segmentio/kafka-go/sasl/plain"
 	"github.com/segmentio/kafka-go/sasl/scram"
 	"strings"
-	"time"
 )
 
 type kafkaSink struct {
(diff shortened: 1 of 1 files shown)


extensions/sinks/kafka/kafka.go
```diff
@@ -133,78 +133,31 @@ func (m *kafkaSink) Collect(ctx api.StreamContext, item interface{}) error {
...
-				case nil:
+	err := m.writer.WriteMessages(ctx, messages...)
+	switch err := err.(type) {
+	case kafkago.Error:
+		if err.Temporary() {
+			return fmt.Errorf(`%s: kafka sink fails to send out the data . %v`, errorx.IOErr, err)
+		}
+	case kafkago.WriteErrors:
+		count := 0
+		for i, _ := range messages {
```
ngjaying, line 144: Please fix the lint

extensions/sinks/kafka/kafka_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+package main
+
+import (
+	"testing"
+)
+
+func TestKafkaSink(t *testing.T) {
+
+	//sink := Kafka()
```
ngjaying, line 23: Still in draft?

superrxan (approved):

PR #1785, opened 2023-04-17
refactor: refactor dbPool

Yisaer:


extensions/util/pool.go
```diff
@@ -23,69 +23,68 @@ import (
...
+	GlobalPool = newDBPool()
 }
 
-type driverPool struct {
+type dbPool struct {
 	isTesting bool
 
 	sync.RWMutex
-	pool map[string]*dbPool
+	// url -> *sql.DB
```
ngjaying, line 39: The db url could have properties like: protocol//[hosts][/database][?properties]. So using url as key, two url with the same protocol, hosts, database may have different db instance. Is this expected?

Yisaer: Yes, I think we can keep this behavior for now.

ngjaying (approved):

PR #1800, opened 2023-04-20
feat: support projectset logicalPlan

Yisaer:


internal/topo/planner/planner.go
```diff
@@ -371,19 +375,70 @@ func createLogicalPlan(stmt *ast.SelectStatement, opt *api.RuleOption, store kv.
...
+	})
+	return err
+}
+
+// extractSRFFromFileds extracts the set-returning-functions in the select field.
+func extractSRFFromFileds(stmt *ast.SelectStatement) (map[string]struct{}, error) {
+	srfCount := 0
+	nonSrfCount := 0
+	srfMapping := make(map[string]struct{}, 0)
+	for _, field := range stmt.Fields {
```
ngjaying, line 425: how about nest unnest function, like abs(unnest(a))

internal/topo/planner/ProjectSetPlan.go
```diff
@@ -0,0 +1,26 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package planner
+
+type ProjectSetPlan struct {
```
ngjaying, line 17: I am not sure if it is necessary to have project & projectSet together in one plan. Can you find the benefits when doing implementation?

Yisaer: We will use `ProjectOp` for `ProjectSet` Executor. Thus we directly build a built-in `ProjectPlan` for `ProjectSetPlan` in order to build `ProjectOp` by `ProjectPlan`.

On the other hand, `ProjectSetPlan` do quite the similar thing as `ProjectPlan`, except `ProjectSet` will handle the `srf`

internal/topo/planner/planner.go
Yisaer: good catch, we should validate that `srf` shouldn't be nested for now.

internal/topo/planner/ProjectSetPlan.go
ngjaying: I have no idea now. Just think about if it is better to keep each plan "single responsibility". We can pipe from SRF to project.

Yisaer: I got it. So the final plan will be `ProjectSet` -> `Project` -> .... , that's ok to me.

ngjaying: Just an alternative way. During implementation, you may have more ideas about which way is better.

PR #1804, opened 2023-04-23
feat: support validate set-returning-functions in parser

Yisaer:


internal/xsql/sqlValidator.go
```diff
@@ -31,5 +31,83 @@ func Validate(stmt *ast.SelectStatement) error {
...
 		}
 	}
+
+	if err := validateSRFNestedForbidden("select", stmt.Fields); err != nil {
+		return err
+	}
+	if err := validateMultiSRFForbidden("select", stmt.Fields); err != nil {
+		return err
+	}
+	return validateSRFForbidden("where", stmt.Condition)
```
ngjaying, line 41: I assume SRF is only allowed in fields. If so, why only check condition here?

Yisaer: Here we validate condition shouldn't has srf.

Yisaer: updated.

internal/xsql/sqlValidator.go
```diff
@@ -31,5 +31,96 @@ func Validate(stmt *ast.SelectStatement) error {
...
 		}
 	}
+
+	if err := validateSRFNestedForbidden("select", stmt.Fields); err != nil {
+		return err
+	}
+	if err := validateMultiSRFForbidden("select", stmt.Fields); err != nil {
+		return err
+	}
+	if err := validateSRFForbidden("where", stmt.Condition); err != nil {
```
ngjaying, line 41: Can we just check not fields case. Pseudo code below:
```
if srf clause is not fields
then
   return validation error
end

Otherwise, we'll need to modify here each time we add new clause type.
```

Yisaer: updated.

ngjaying (approved):

PR #1807, opened 2023-04-23
feat(sink): add file sink/source compress support

carlclone:


internal/io/file/file_sink.go
```diff
@@ -39,6 +39,7 @@ type sinkConf struct {
 	HasHeader          bool     `json:"hasHeader"`
 	Delimiter          string   `json:"delimiter"`
 	Format             string   `json:"format"` // only use for validation; transformation is done in sink_node
+	CompressAlgorithm  string   `json:"compressAlgorithm"`
```
ngjaying, line 42: How about using name "compress" to align with MQTT sink

internal/io/file/file_writer.go
```diff
@@ -71,16 +96,29 @@ func createFileWriter(ctx api.StreamContext, fn string, ft FileType, headers str
...
 		}
-		err := fw.Writer.(*bufio.Writer).Flush()
-		if err != nil {
-			ctx.GetLogger().Errorf("file sink fails to flush with error %s.", err)
+		if fw.Compress!="" && fw.Compress!=NONE_COMPRESS {
+			e := fw.Writer.(io.Closer).Close()
+			if e!=nil {
+				ctx.GetLogger().Errorf("file sink fails to close compress writer with error %s.", err)
+			}
+			err = fw.fileBuffer.Flush()
```
ngjaying, line 111: flush after writer close?

internal/io/file/constant.go
```diff
@@ -22,6 +22,13 @@ const (
 	LINES_TYPE FileType = "lines"
 )
 
+const (
+	ZLIB  = "zlib"
```
ngjaying, line 26: I am not sure if zlib/flate is needed. I think gzip use those compression algorithm and add some headers to the file?

internal/io/file/file_sink.go
```diff
@@ -102,6 +103,15 @@ func (m *fileSink) Configure(props map[string]interface{}) error {
...
 		}
 	}
+	if c.Compress != ZLIB && c.Compress != GZIP && c.Compress != FLATE && c.Compress != NONE_COMPRESS && c.Compress != "" {
+		return fmt.Errorf("compressAlgorithm must be one of none, zlib, gzip or flate")
+	}
+
+	//todo; check whether rolling enabled when compress enabled , if not , error
+	//if (c.Compress!="" && c.Compress!="none") && (c.RollingNamePattern == "" || c.RollingNamePattern == "none")  {
+	//	return fmt.Errorf("when compress enabled, rollingNamePattern must be one of prefix, suffix")
+	//}
```
Yisaer, line 113: why remain these code?

internal/io/file/file_sink.go
```diff
@@ -102,6 +103,15 @@ func (m *fileSink) Configure(props map[string]interface{}) error {
 			c.Delimiter = ","
 		}
 	}
+	if c.Compress != ZLIB && c.Compress != GZIP && c.Compress != FLATE && c.Compress != NONE_COMPRESS && c.Compress != "" {
+		return fmt.Errorf("compressAlgorithm must be one of none, zlib, gzip or flate")
+	}
```
Yisaer, line 108: Can we maintained the support compress algorithm type in list, so that if we support the new compress algorithm, we don't have to update here.

internal/io/file/file_source.go
```diff
@@ -128,6 +132,10 @@ func (fs *FileSource) Configure(fileName string, props map[string]interface{}) e
 	if cfg.Delimiter == "" {
 		cfg.Delimiter = ","
 	}
+
+	if cfg.Compress != ZLIB && cfg.Compress != GZIP && cfg.Compress != FLATE && cfg.Compress != NONE_COMPRESS && cfg.Compress != "" {
+		return fmt.Errorf("compressAlgorithm must be one of none, zlib, gzip or flate")
+	}
```
Yisaer, line 138: ditto

internal/io/file/file_source.go
```diff
@@ -340,16 +348,38 @@ func (fs *FileSource) prepareFile(ctx api.StreamContext, file string) (io.Reader
 		ctx.GetLogger().Error(err)
 		return nil, err
 	}
+	var reader io.ReadCloser
+
+	switch fs.config.Compress {
+	case "flate":
```
Yisaer, line 354: use Constant here for the switch case.

internal/io/file/constant.go
```diff
@@ -22,6 +22,13 @@ const (
 	LINES_TYPE FileType = "lines"
 )
 
+const (
+	ZLIB          = "zlib"
+	GZIP          = "gzip"
+	FLATE         = "flate"
+	NONE_COMPRESS = "none"
```
Yisaer, line 29: What does `NONE_COMPRESS` stand for?

Yisaer (changes requested):
plz run `go fmt ./...` and `make lint` to fix lint and fmt problem.

internal/io/file/file_sink_test.go
```diff
@@ -439,15 +634,209 @@ func TestFileSinkRollingCount_Collect(t *testing.T) {
...
+				}
 				if !reflect.DeepEqual(contents, tt.contents[i]) {
 					t.Errorf("\nexpected\t %q \nbut got\t\t %q", tt.contents[i], string(contents))
 				}
 			}
 		})
 	}
 }
+
+func TestFileSinkCompress_Collect(t *testing.T) {
```
Yisaer, line 663: It seems `TestFileSinkCompress_Collect` both test the compress for fileSink and fileSource, can you split the testcase into 2 parts and move the `fileSource` testcase into `file_source_test`?

internal/io/file/file_source.go
```diff
@@ -33,18 +36,19 @@ import (
...
+	IsTable           bool     `json:"isTable"`
+	SendInterval      int      `json:"sendInterval"`
+	ActionAfterRead   int      `json:"actionAfterRead"`
+	MoveTo            string   `json:"moveTo"`
+	HasHeader         bool     `json:"hasHeader"`
+	Columns           []string `json:"columns"`
+	IgnoreStartLines  int      `json:"ignoreStartLines"`
+	IgnoreEndLines    int      `json:"ignoreEndLines"`
+	Delimiter string `json:"delimiter"`
+	Compress  string `json:"compress"`
```
ngjaying, line 51: In source side, this should be "decompression" to be aligned with MQTT source

internal/io/file/constant.go
ngjaying: - I think all these zlib/flate/gzip are the same compression algorithm. We can leave only gzip, and maybe add zstd
- Any reason not to use empty string "" as none. It is the default string init value.

internal/io/file/file_writer.go
Yisaer: I think it's ok here, `fw.Writer` is the compress writer here.

Yisaer (approved):

ngjaying:
LGTM,  sth wrong with the wasm test. Will rerun the test once it is fixed

ngjaying:
Could you please rebase on the master to see if the ut can pass? Thanks.

ngjaying:
Thank you very much!

PR #1808, opened 2023-04-23
refactor: use go builtin multi error wrapping to optimize error

Rui-Gan:


internal/binder/function/binder.go
```diff
@@ -51,17 +52,17 @@ func applyFactory(f binder.FactoryEntry) {
 }
 
 func Function(name string) (api.Function, error) {
-	e := make(errorx.MultiError)
+	errs := make([]error, 0)
```
ngjaying, line 55: Can we just define `var err error` and use errors.Join later.

ngjaying (changes requested):

internal/binder/function/binder.go
ngjaying: Additionally, the cached file is not needed

ngjaying (approved):

PR #1809, opened 2023-04-23
chore: deprecate sink runAsync property

Rui-Gan:


ngjaying (commented):
LGTM. But maybe need to add some test cases to pass the coverage criteria

ngjaying (approved):
The newly added code checks for deprecation which is not worth testing. So approve.

PR #1814, opened 2023-04-24
feat: support projectset operation for tupleRow

Yisaer:


internal/topo/node/projectset_node.go
```diff
@@ -0,0 +1,141 @@
...
+	logger := ctx.GetLogger()
+	defer func() {
+		logger.Infof("ProjectSetNode %s instance %d done, cancelling future items", node.name, ctx.GetInstanceId())
+	}()
+	stats, err := metric.NewStatManager(ctx, "op")
+	if err != nil {
+		infra.DrainError(ctx, err, errCh)
+		return
+	}
+	node.statManagers = append(node.statManagers, stats)
```
ngjaying, line 95: In operations.go, there is concurrency support. We may add it as well.

ngjaying (commented):
A general comment: This node is actually 80%+ the same as UnaryOperator except the part to deal with the operation output. Can we find a way to share the code instead of copy them?

internal/topo/node/projectset_node.go
Yisaer: updated.

ngjaying (approved):

PR #1824, opened 2023-04-25
fix(metrics): fix source processLatency in metrics

Rui-Gan:


internal/topo/node/source_node.go
```diff
@@ -149,7 +151,11 @@
 									continue
 								}
 								stats.IncTotalRecordsIn()
-								stats.ProcessTimeStart()
+								if t, _ := ctx.GetState(context.RcvTime); t == nil {
```
ngjaying, line 154: The state will be serialized which is unnecessary for this instant value. How about transfer it by the data(SourceTuple)? We can add a Timestamp field, which is necessary not just for metrics.

PR #1827, opened 2023-04-26
feat: support projesetset for collection

Yisaer:


internal/topo/operator/projectset_operator.go
```diff
@@ -32,34 +32,57 @@
...
-		if !ok {
-			return fmt.Errorf("can't find the result from the %v function", srfName)
+		newTuples, err := ps.handleSRFRow(srfName, input)
+		if err != nil {
+			return err
 		}
-		aValues, ok := aValue.([]interface{})
-		if !ok {
-			return fmt.Errorf("the result from the %v function should be array", srfName)
+		for _, tuple := range newTuples {
```
ngjaying, line 44: why need to copy the result?

internal/xsql/collection.go
```diff
@@ -157,6 +158,22 @@
...
 
+func (w *WindowTuples) RangeProjectSet(f func(i int, r CloneAbleRow) ([]CloneAbleRow, error)) ([]Collection, error) {
+	newCollections := make([]Collection, 0)
+	for i, oldTupleRow := range w.Content {
+		newTuples, err := f(i, oldTupleRow.Clone())
+		if err != nil {
+			return nil, err
+		}
+		for _, newTuple := range newTuples {
+			ns := w.Clone().(*WindowTuples)
```
ngjaying, line 169: why return multiple collections? we can just flatten to a collection. It is also save to replace the old content

internal/xsql/collection.go
```diff
@@ -37,6 +37,7 @@
 // Collection A collection of rows as a table. It is used for window, join, group by, etc.
 type Collection interface {
 	SortingData
+	RangeProjectSet(f func(i int, r CloneAbleRow) ([]CloneAbleRow, error)) ([]Collection, error)
```
ngjaying, line 40: This seems like only projectset needs, I don't think we should put it here. This interface should only provide some basic/atomic functions.

internal/topo/topotest/rule_test.go
```diff
@@ -22,6 +22,207 @@ import (
 	"github.com/lf-edge/ekuiper/pkg/api"
 )
 
+func TestSRFSQL(t *testing.T) {
```
ngjaying, line 25: Maybe add most tests in the operator level? Here is more expensive.

ngjaying (changes requested):
Suggest to add more comments ~

internal/topo/topotest/rule_test.go
Yisaer: it's too hard to build testcase in operator level for this case.

internal/topo/operator/projectset_operator.go
```diff
@@ -23,43 +23,101 @@ import (
...
-				newTupleRow.Set(srfName, v)
+		switch ts := input.(type) {
+		case *xsql.JoinTuples:
+			newTuples := make([]*xsql.JoinTuple, 0)
+			for _, tuple := range ps.collectionRows {
+				newTuples = append(newTuples, tuple.(*xsql.JoinTuple))
+			}
+			ts.Content = newTuples
+		case *xsql.GroupedTuplesSet:
+			newTuples := make([]*xsql.GroupedTuples, 0)
```
ngjaying, line 72: 0 -》len(ps.collectionRows)

ngjaying: Or make([]*xsql.GroupedTuples, 0, len(ps.collectionRows)) Or is it possible to have no new allocation/copy

ngjaying: Other types also have this problem

ngjaying (changes requested):

internal/topo/operator/projectset_operator.go
Yisaer: updated.

ngjaying (approved):

PR #1829, opened 2023-04-26
fix(metrics): fix source processLatency in metrics

Rui-Gan:


pkg/api/stream.go
```diff
@@ -17,22 +17,26 @@ package api
...
 }
 
 type DefaultSourceTuple struct {
 	Mess map[string]interface{} `json:"message"`
 	M    map[string]interface{} `json:"meta"`
+	Time time.Time              `json:"timestamp"`
 }
 
-func NewDefaultSourceTuple(message map[string]interface{}, meta map[string]interface{}) *DefaultSourceTuple {
+func NewDefaultSourceTuple(message map[string]interface{}, meta map[string]interface{}, timestamp time.Time) *DefaultSourceTuple {
```
ngjaying, line 35: Let's keep the old function with a default timestamp and create a new function for 3 parameters. Otherwise, users will need to update the customized plugin.

internal/xsql/source_tuple.go
```diff
@@ -25,3 +31,7 @@
 func (t *ErrorSourceTuple) Meta() map[string]interface{} {
 	return nil
 }
+
+func (t *ErrorSourceTuple) Timestamp() time.Time {
+	return conf.Clock.Now()
```
ngjaying, line 36: Use conf.GetNow()

extensions/sources/random/random.go
```diff
@@ -99,7 +101,7 @@ func (s *randomSource) Open(ctx api.StreamContext, consumer chan<- api.SourceTup
 				continue
 			}
 			logger.Debugf("Send out data %v", next)
-			consumer <- api.NewDefaultSourceTuple(next, nil)
+			consumer <- api.NewDefaultSourceTuple(next, nil, conf.Clock.Now())
```
ngjaying, line 104: I think we should get the time at the earliest, the same for all other sources.
And use conf.GetNow() should be fine to get the current time

ngjaying (changes requested):
General suggestion for all sources:
1. When to get the time: at the very begininng of each processing
2. How to get the time: conf.GetNow()

ngjaying (approved):

PR #1833, opened 2023-05-04
refactor: make DefaultContext.Decode return []map[string]interface{}

Yisaer:


pkg/api/stream.go
```diff
@@ -206,7 +206,7 @@ type StreamContext interface {
 	TransformOutput(data interface{}) ([]byte, bool, error)
 	// Decode is set in the source according to the format.
 	// It decodes byte array into map or map slice.
-	Decode(data []byte) (map[string]interface{}, error)
+	Decode(data []byte) ([]map[string]interface{}, error)
```
ngjaying, line 209: This is a breaking change in API. We are trying to avoid such changes in the 1.x version. We may only introduce such changes in v2.

ngjaying (changes requested):

internal/topo/context/decoder.go
```diff
@@ -37,3 +37,23 @@
...
+
+func (c *DefaultContext) DecodeIntoList(data []byte) ([]interface{}, error) {
+	v := c.Value(DecodeKey)
+	f, ok := v.(message.Converter)
+	if ok {
+		t, err := f.Decode(data)
+		if err != nil {
+			return nil, fmt.Errorf("decode failed: %v", err)
+		}
+		typeErr := fmt.Errorf("only map[string]interface{} and []map[string]interface{} is supported but got: %v", t)
```
ngjaying, line 49: why define a temporary variable here

internal/topo/context/decoder.go
```diff
@@ -37,3 +37,23 @@
...
+		if result, ok := t.(map[string]interface{}); ok {
+			return []interface{}{result}, nil
+		}
+		if result, ok := t.([]interface{}); ok {
```
ngjaying, line 53: Maybe also allow []map[string]interface{}. Since there are more than 1 check, is it better to use type switch

ngjaying (changes requested):
Not only MQTT source is affected. Are you going to modify other sources in another PR?

internal/topo/context/decoder.go
Yisaer: As we use `interface{}` for the `json.Unmarshal`. For the data below:
```json
[{"a":1},{"a":1}]
```
The converter will convert it into
```
[]interface{} {
 map[string]interface{} {"a":1}
 map[string]interface{} {"a":2}
}
```

Thus we don't assert `t` should be `[]map[string]interface{}` here

Yisaer:
> Not only MQTT source is affected. Are you going to modify other sources in another PR?

yes, I tend to modify other sources in next pr.

internal/topo/context/decoder.go
ngjaying: JSON is just one of the decode method. Custom method may return []map

internal/topo/context/decoder.go
```diff
@@ -37,3 +37,22 @@ func (c *DefaultContext) Decode(data []byte) (map[string]interface{}, error) {
...
+	f, ok := v.(message.Converter)
+	if ok {
+		t, err := f.Decode(data)
+		if err != nil {
+			return nil, fmt.Errorf("decode failed: %v", err)
+		}
+		switch r := t.(type) {
+		case map[string]interface{}:
+			return []interface{}{r}, nil
+		case []interface{}:
```
ngjaying, line 52: case []interface{}, []map[string]interface{}:

internal/topo/context/decoder.go
Yisaer: updated.

internal/topo/context/decoder.go
Yisaer: updated.

ngjaying (approved):

PR #1834, opened 2023-05-04
feat: add sink common property: fields

Rui-Gan:


internal/io/file/file_sink.go
```diff
@@ -234,6 +235,20 @@
 					}
 				}
 			}
+			if ctx.Value(context.SelectFields) != nil {
```
ngjaying, line 238: No need to save "fields", it can be gotten in configure

internal/topo/context/transform.go
```diff
@@ -20,13 +20,27 @@
...
+	bytes, ok, err := f(data)
+	if err != nil || ok || c.Value(SelectFields) == nil {
+		return bytes, ok, err
+	}
+
+	fields := c.Value(SelectFields).([]string)
+	s := c.Value(SelectKey)
+	sf, ok := s.(transform.SelectFunc)
+	if !ok {
+		return bytes, false, fmt.Errorf("no select configured")
```
ngjaying, line 43: What's the relationship between transform(template, format) and fields? We need to clearly define and document it.

ngjaying (changes requested):

internal/io/edgex/edgex_sink_test.go
```diff
@@ -598,8 +598,9 @@ func TestEdgeXTemplate_Apply(t1 *testing.T) {
 		var payload []map[string]interface{}
 		json.Unmarshal([]byte(t.input), &payload)
 		dt := t.conf["dataTemplate"]
-		tf, _ := transform.GenTransform(cast.ToStringAlways(dt), "json", "", "")
+		tf, sf, _ := transform.GenTransform(cast.ToStringAlways(dt), "json", "", "", []string{})
```
ngjaying, line 601: In normal cases (like MQTT), the 3 properties are run in order dataTemplate -> fields -> format. Currently dataTemplate/format are handled in transformFunc. I think fields is also part of the transform. Any benefits to mvoe it out? The sf may not work because tf will return bytes

ngjaying (changes requested):

internal/io/edgex/edgex_sink_test.go
Rui-Gan: I previously thought that plugin sink could override this select function, but it seems not necessary at present. I will modify the code according to your suggestion.

extensions/sinks/influx/influx.go
```diff
@@ -119,7 +122,7 @@ func (m *influxSink) Open(ctx api.StreamContext) (err error) {
 
 func (m *influxSink) Collect(ctx api.StreamContext, data interface{}) error {
 	logger := ctx.GetLogger()
-	if m.hasTransform {
+	if m.hasTransform || len(m.fields) > 0 {
```
ngjaying, line 125: Is it possible to not serialize/deserialize? Maybe do it in another PR

internal/io/file/file_sink.go
```diff
@@ -234,6 +235,22 @@
 					}
 				}
 			}
+			if len(m.c.Fields) > 0 {
```
ngjaying, line 238: how about just let headers = fields? When writing, make sure all fields are have value, even it is nil

internal/topo/transform/template.go
```diff
@@ -70,10 +73,13 @@
 		switch format {
 		case message.FormatJson:
 			if transformed {
-				return bs, transformed, nil
+				return selectJson(bs, fields, transformed)
 			}
 			j, err := json.Marshal(d)
-			return j, false, err
+			if err != nil {
```
ngjaying, line 79: should avoid too much time json marshal/unmarshal. Line 78 seems unnecessary.

internal/topo/transform/template.go
```diff
@@ -84,7 +90,18 @@
 				d = m
 			}
 			b, err := c.Encode(d)
-			return b, transformed, err
+			if err != nil {
+				return nil, false, fmt.Errorf("fail to encode data %v for error %v", d, err)
+			}
+			mm, ok := d.(map[string]interface{})
```
ngjaying, line 96: why need encode twice? why need out converter?

ngjaying (changes requested):

extensions/sinks/influx/influx.go
Rui-Gan: Yeah, I'll optimize it later.

internal/topo/transform/template.go
Rui-Gan: This is because the converter will record cols, and if you do not use out here, you will get an unwanted field after encode.

internal/topo/transform/template.go
```diff
@@ -53,7 +53,7 @@
 		}
 		tp = temp
 	}
-	return func(d interface{}) ([]byte, bool, error) {
+	return func(d interface{}, s bool) ([]byte, bool, error) {
```
ngjaying, line 56: Add some comments about the meaning of s.
The flow in this function is still a bit messy. If s=false means skip fields/format, we can check it only after tp and before format?

internal/topo/transform/template.go
```diff
@@ -70,10 +70,46 @@
...
+					return bs, true, nil
+				}
+				err = json.Unmarshal(bs, &d)
+			}
+			if !s || len(fields) == 0 {
+				j, err := json.Marshal(d)
+				if err != nil {
+					return nil, false, fmt.Errorf("fail to encode data %v for error %v", d, err)
+				}
+				return j, true, nil
```
ngjaying, line 83: The second return value should be false

internal/topo/transform/template.go
```diff
@@ -70,10 +70,46 @@
 		switch format {
 		case message.FormatJson:
 			if transformed {
```
ngjaying, line 72: In previous code, we assume after transform, the result is json bytes so that it is return immediately to save one marshal/unmarshal pair. Otherwise, all format can be handled in the same way.

internal/topo/transform/template.go
```diff
@@ -83,8 +119,22 @@
...
-			b, err := c.Encode(d)
-			return b, transformed, err
+			if !s || len(fields) == 0 {
+				b, err := c.Encode(d)
+				if err != nil {
+					return nil, false, fmt.Errorf("fail to encode data %v for error %v", d, err)
+				}
+				return b, true, nil
+			}
+			mm, ok := d.(map[string]interface{})
```
ngjaying, line 129: d must be map or []map. Should deal with both types; Maybe let selectMap accept d and return interface because Encode accept an interface.

internal/topo/transform/template.go
```diff
@@ -70,10 +70,46 @@
...
+				err = json.Unmarshal(bs, &d)
+			}
+			if !s || len(fields) == 0 {
+				j, err := json.Marshal(d)
+				if err != nil {
+					return nil, false, fmt.Errorf("fail to encode data %v for error %v", d, err)
+				}
+				return j, true, nil
+			}
+			switch d.(type) {
```
ngjaying, line 85: These logics are actually the same for all formats. json.Marshal == ctx.Encode

extensions/sinks/influx2/influx2.go
```diff
@@ -111,8 +114,8 @@ func (m *influxSink2) Open(ctx api.StreamContext) (err error) {
 
 func (m *influxSink2) Collect(ctx api.StreamContext, data interface{}) error {
 	logger := ctx.GetLogger()
-	if m.hasTransform {
-		jsonBytes, _, err := ctx.TransformOutput(data)
+	if m.hasTransform || len(m.fields) > 0 {
```
ngjaying, line 117: can keep the old logic, only deal with dateTemplate. Make the selectMap an util function to be used to handle fields

ngjaying (changes requested):

internal/io/file/file_sink.go
```diff
@@ -216,22 +217,26 @@
...
+						header = make([]string, len(v[0]))
+						i := 0
+						for k := range v[0] {
+							header[i] = k
+							i++
+						}
+					}
 				}
 			}
 			sort.Strings(header)
```
ngjaying, line 242: TODO: if headers are defined by user, we should find a way to keep the order

internal/topo/transform/template.go
```diff
@@ -25,9 +25,15 @@
...
 
-type TransFunc func(interface{}) ([]byte, bool, error)
+// TransFunc is the function to transform data
 
-func GenTransform(dt string, format string, schemaId string, delimiter string) (TransFunc, error) {
+// The second parameter indicates whether to select fields based on the fields property.
+// If it is false, then after the dataTemplate, output the result directly.
+// If it is true, then after the dataTemplate, select the fields based on the fields property.
+
+type TransFunc func(interface{}, bool) ([]byte, bool, error)
```
ngjaying, line 34: TODO: Add a breaking change notification in release note and plugin development tutorial

ngjaying (commented):
Some comment and todos that can do later

ngjaying (approved):

PR #1842, opened 2023-05-06
feat: support more source for json list object

### LINKED ISSUE #605: Cannot unmarshal array into Go value of type map[string]interface {} with MQTT source
- Create below stream.
```json
{
   "sql" : "create stream demo () WITH (FORMAT=\"JSON\", DATASOURCE=\"devices/+/messages\" )"
}
```
- Create below rule.
```json
{
  "id": "rule1",
  "sql": "SELECT * FROM demo",
  "actions": [
    {
      "log": {}
    },
    {
      "mqtt": {
        "server": "tcp://broker.emqx.io:1883",
        "topic": "devices/result",
        "qos": 1,
        "clientId": "demo_001"
      }
    }
  ]
}
```
- Send with below data

```json
[
  {
    "att1" : 1,
    "att2" : 2
  },
  {
    "att1" : 3,
    "att2" : 4
  }
]
```

In the log, it print below error.
```
time="2020-11-17 12:51:41" level=error msg="Invalid data format, cannot convert [\n  {\n    \"att1\" : 1,\n    \"att2\" : 2\n  },\n  {\n    \"att1\" : 3,\n    \"att2\" : 4\n  }\n] into JSON with error json: cannot unmarshal array into Go value of type map[string]interface {}" file="extensions/mqtt_source.go:165" rule=rule1
```

Yisaer:
Closes #605

### DIFF
--- a/internal/io/http/client.go
+++ b/internal/io/http/client.go
@@ -390,13 +390,19 @@ func decode(ctx api.StreamContext, data []byte) ([]map[string]interface{}, error
 	if err != nil {
 		return nil, err
 	}
-	if r2, ok := r1.(map[string]interface{}); ok {
-		return []map[string]interface{}{r2}, nil
-	}
-	if rlist, ok := r1.([]interface{}); ok {
-		r2 := make([]map[string]interface{}, len(rlist))
-		for i, m := range rlist {
-			r2[i] = m.(map[string]interface{})
+	switch rt := r1.(type) {
+	case map[string]interface{}:
+		return []map[string]interface{}{rt}, nil
+	case []map[string]interface{}:
+		return rt, nil
+	case []interface{}:
+		r2 := make([]map[string]interface{}, len(rt))
+		for i, m := range rt {
+			if rm, ok := m.(map[string]interface{}); ok {
+				r2[i] = rm
+			} else {
+				return nil, fmt.Errorf("only map[string]interface{} and []map[string]interface{} is supported")
+			}
 		}
 		return r2, nil
 	}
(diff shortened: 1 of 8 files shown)


internal/io/http/client.go
```diff
@@ -350,20 +351,22 @@
...
+		payloads, err := decode(ctx, c)
 		if err != nil {
 			return nil, c, err
 		}
-		ro := &bodyResp{}
-		err = cast.MapToStruct(payload, ro)
-		if err != nil {
-			return nil, c, fmt.Errorf("invalid body response: %v", err)
-		}
 		if resp.StatusCode < 200 || resp.StatusCode > 299 {
```
ngjaying, line 358: This was a bug, it actually wants to check ro.Code. Some http service do not use http status code, but put the code in the payload.

ngjaying (changes requested):

internal/io/http/client.go
Yisaer: updated.

internal/io/http/client.go
```diff
@@ -350,20 +351,22 @@ func (cc *ClientConf) parseResponse(ctx api.StreamContext, resp *http.Response,
...
-		err = cast.MapToStruct(payload, ro)
-		if err != nil {
-			return nil, c, fmt.Errorf("invalid body response: %v", err)
-		}
-		if resp.StatusCode < 200 || resp.StatusCode > 299 {
-			return nil, c, fmt.Errorf("http status code is not 200: %v", payload)
+		for _, payload := range payloads {
+			ro := &bodyResp{}
+			err = cast.MapToStruct(payload, ro)
+			if err != nil {
```
ngjaying, line 361: Not sure if we should fail whole response for one parsing error or still return partial result.

ngjaying (approved):

internal/io/http/client.go
Yisaer: I think fail is better, otherwise it may look like we discard some requests.

PR #1853, opened 2023-05-08
test: fix loop spin

xjasonlyu:
`default` was wrongly used in a for-loop.

### DIFF
--- a/internal/io/memory/memory_test.go
+++ b/internal/io/memory/memory_test.go
@@ -16,16 +16,18 @@ package memory
 
 import (
 	"fmt"
+	"reflect"
+	"testing"
+	"time"
+
 	"github.com/benbjohnson/clock"
 	"github.com/gdexlab/go-render/render"
+
 	"github.com/lf-edge/ekuiper/internal/conf"
 	"github.com/lf-edge/ekuiper/internal/io/memory/pubsub"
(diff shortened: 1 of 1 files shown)


internal/io/memory/memory_test.go
```diff
@@ -83,7 +85,6 @@ func TestSharedInmemoryNode(t *testing.T) {
 				t.Errorf("result %s should be equal to %s", res, expected)
 			}
 			return
-		default:
```
ngjaying, line 86: This is to return immediately if no receiving anything, otherwise the test will be blocked for error case.

ngjaying (changes requested):

internal/io/memory/memory_test.go
xjasonlyu: So instead, we need to `return` after `default`?

ngjaying: This is a little tricky.  There are actually two receivers and only one sender. So we need to skip if no receive and try for the next time.

xjasonlyu: I see, then I think we need to add a `time.Sleep(1 * time.Millisecond)` to prevent the loop spin.

xjasonlyu: OK, I will bypass this check in the lint PR.

PR #1858, opened 2023-05-08
feat: introduce sendManager in util

Yisaer:


internal/util/sendManager.go
```diff
@@ -0,0 +1,103 @@
...
+	"context"
+	"fmt"
+	"time"
+)
+
+type SendManager struct {
+	lingerInterval int
+	batchSize      int
+	bufferCh       chan interface{}
+	buffer         []interface{}
```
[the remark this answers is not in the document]
Yisaer: If the input is a []map, the batch size would be []map's length.

internal/util/sendManager.go
```diff
@@ -0,0 +1,103 @@
...
+	if batchSize < 1 && lingerInterval < 1 {
+		return nil, fmt.Errorf("either batchSize or lingerInterval should be larger than 0")
+	}
+	sm := &SendManager{
+		batchSize:      batchSize,
+		lingerInterval: lingerInterval,
+	}
+	sm.buffer = make([]interface{}, 0, batchSize)
+	sm.bufferCh = make(chan interface{})
+	sm.outputCh = make(chan []interface{}, 16)
```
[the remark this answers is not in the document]
Yisaer: In fact, we don't need `outputCh` as a buffer channel if the data in the `outpuCh` can be consumed immediately.  16 is a just proper buffer length I think in this case.

internal/io/sink/send_manager.go
```diff
@@ -0,0 +1,122 @@
...
+			sm.currIndex++
+			if sm.currIndex >= sm.batchSize {
+				sm.send()
+			}
+		}
+	}
+}
+
+func (sm *SendManager) runWithTickerAndBatchSize(ctx context.Context) {
+	ticker := time.NewTicker(time.Duration(sm.lingerInterval) * time.Millisecond)
```
ngjaying, line 90: You can use conf.GetTicker which will be testable.

Yisaer: updated.

internal/io/sink/send_manager.go
```diff
@@ -0,0 +1,123 @@
...
+			sm.currIndex++
+			if sm.currIndex >= sm.batchSize {
+				sm.send()
+			}
+		}
+	}
+}
+
+func (sm *SendManager) runWithTickerAndBatchSize(ctx context.Context) {
+	ticker := conf.GetTicker(sm.lingerInterval)
```
ngjaying, line 91: Can we get this function covered since we have the mock ticker? It is better to have the new code coverage bigger than 70%

Yisaer: updated.

ngjaying (commented):
Please take a look at TestFileSinkRolling_Collect in internal/io/file/file_sink_test.go. We can use mockClock to mock timing. Maybe we can test by that for time related send.

ngjaying (approved):
Thanks for the lightning quick response!

PR #1861, opened 2023-05-09
fix: process unhandled errors

xjasonlyu:


internal/server/server.go
```diff
@@ -204,6 +205,10 @@ func initRuleset() error {
 		}
 		conf.Log.Infof("start to initialize ruleset")
 		_, counts, err := rulesetProcessor.Import(content)
+		if err != nil {
```
ngjaying, line 208: @superrxan Is this expected to return for error? Moreover, suggest to use conf.Log.Errorf instead of Infof

internal/io/http/httppull_source_test.go
```diff
@@ -18,24 +18,26 @@ import (
...
+	"github.com/benbjohnson/clock"
+	"github.com/gorilla/mux"
+
+	"github.com/lf-edge/ekuiper/internal/conf"
+	"github.com/lf-edge/ekuiper/internal/io/mock"
+	"github.com/lf-edge/ekuiper/pkg/api"
 )
 
-func jsonOut(w http.ResponseWriter, err error, out interface{}) {
+func jsonOut(w http.ResponseWriter, _ error, out interface{}) {
```
ngjaying, line 37: if err error is useless, just remove from the parameters

internal/service/external_service_rule_test.go
```diff
@@ -517,10 +519,10 @@ func TestRestService(t *testing.T) {
 	}, 0)
 }
 
-func jsonOut(w http.ResponseWriter, err error, out interface{}) {
+func jsonOut(w http.ResponseWriter, _ error, out interface{}) {
```
ngjaying, line 522: just remove the error parameter if useless

ngjaying (changes requested):

internal/io/http/httppull_source_test.go
xjasonlyu: Done.

internal/service/external_service_rule_test.go
xjasonlyu: done.

internal/server/server.go
xjasonlyu: From my perspective, if we don't return for error, there could be an out-of-index panic when the later log calls for counts[x].

superrxan: I agree with return for error

PR #1865, opened 2023-05-09
feat: add . for object field

Rui-Gan:


docs/zh_CN/sqls/json_expr.md
```diff
@@ -53,11 +53,27 @@ SELECT demo.name->first FROM demo
...
 
+```
+SELECT demo.name.first FROM demo
+{"first" : "Tom"}
+```
+
+
+
 ```
 SELECT name->first AS fname FROM demo
```
ngjaying, line 64: We can add the equivalent writing for this name.first

ngjaying (commented):
This requirement is to smartly parse . to stream/table name or field name. It seems field.nestedField is not considered in this PR. Only table.field.nestedField is tested.

PR #1866, opened 2023-05-09
fix: keep the order for delimited format

Rui-Gan:


ngjaying (changes requested):
Can we add some testing?

ngjaying (approved):
Thank you!

PR #1875, opened 2023-05-11
feat: introduce sendManager in sink

Yisaer:


internal/io/mqtt/mqtt_sink.go
```diff
@@ -27,10 +28,29 @@ import (
...
-	Compression string `json:"compression"`
+	Tpc            string `json:"topic"`
+	Qos            byte   `json:"qos"`
+	Retained       bool   `json:"retained"`
+	Compression    string `json:"compression"`
+	BatchSize      int    `json:"batchSize"`
+	LingerInterval int    `json:"lingerInterval"`
+}
+
+func (c *AdConf) validate() error {
```
ngjaying, line 39: Think it over, the dynamic props are still supported actually. With batch support, we just change the payload that the sink received. The sink can adapt to the change by updating its template/dynamic prop.

ngjaying (changes requested):

ngjaying (changes requested):

ngjaying (approved):

PR #1877, opened 2023-05-11
feat: add . for object field

Rui-Gan:


internal/xsql/sqlValidator.go
```diff
@@ -115,3 +115,69 @@ func isSRFExists(node ast.Node) bool {
 	})
 	return exists
 }
+
+func validateFields(stmt *ast.SelectStatement) {
```
ngjaying, line 119: Not only fields/sortFields has the fieldRef. For example, where a.b > 10. Need to visit the full statement and update all fieldRef

ngjaying (changes requested):

internal/xsql/sqlValidator.go
```diff
@@ -115,3 +115,65 @@ func isSRFExists(node ast.Node) bool {
...
+	streamName := getStreamNames(stmt)
+	for i, field := range stmt.Fields {
```
ngjaying, line 121: Not only fields must be revisited, the on condition in joins should also be revisited.

internal/xsql/sqlValidator.go
```diff
@@ -115,3 +115,65 @@ func isSRFExists(node ast.Node) bool {
...
+		stmt.Fields[i].Expr = validateExpr(field.Expr, streamName)
+	}
+}
+
+func validateExpr(expr ast.Expr, streamName []string) ast.Expr {
+	switch expr.(type) {
```
ngjaying, line 127: Only vist binaryExpr is not enough.

ngjaying (changes requested):

ngjaying (approved):

PR #1883, opened 2023-05-11
feat: support more functions

Yisaer:


internal/binder/function/funcs_array.go
```diff
@@ -29,4 +35,99 @@ func registerArrayFunc() {
...
+					return i + 1, true
+				}
+			}
+			return 0, true
+		},
+		val: func(ctx api.FunctionContext, args []ast.Expr) error {
+			return ValidateLen(2, len(args))
+		},
+	}
+	builtins["slice"] = builtinFunc{
```
ngjaying, line 56: This one is already supported by [x:y]? Please check.

ngjaying (commented):
Another general thinking, can we unify the implementation for aggregate function and array function. Like max & arr_max, they may use in different SQL query but the implementation seems the same.

internal/binder/function/funcs_array.go
Yisaer: It seems `slice` is equal to `subset`, but I think it's ok to implement `slice` here.

ngjaying: What's the rationale here? To keep the core small, I would recommend not adding any redundant features.

Yisaer: updated.

ngjaying (approved):

PR #1896, opened 2023-05-15
feat: add sink common property: dataField

Rui-Gan:


internal/topo/transform/template.go
```diff
@@ -134,22 +120,41 @@ func GenTp(dt string) (*template.Template, error) {
...
 		var m map[string]interface{}
 		err := json.Unmarshal(input.([]byte), &m)
 		if err != nil {
-			return input, fmt.Errorf("fail to decode data %s for error %v", string(input.([]byte)), err)
+			return input, false, fmt.Errorf("fail to decode data %s for error %v", string(input.([]byte)), err)
 		}
 		input = m
 	}
 
+	if dataField != "" {
```
ngjaying, line 137: Do we need to handle slice input?

extensions/sinks/sql/sql.go
```diff
@@ -30,13 +31,13 @@ import (
...
-	Fields         []string `json:"fields"`
-	DataTemplate   string   `json:"dataTemplate"`
-	TableDataField string   `json:"tableDataField"`
-	RowkindField   string   `json:"rowkindField"`
-	KeyField       string   `json:"keyField"`
+	Url          string   `json:"url"`
+	Table        string   `json:"table"`
+	Fields       []string `json:"fields"`
+	DataTemplate string   `json:"dataTemplate"`
+	DataField    string   `json:"dataField"`
```
ngjaying, line 38: This is a breaking change, shall we keep both field now?

ngjaying (changes requested):

ngjaying (approved):

PR #1903, opened 2023-05-17
fix(rule): handle crash when call createState

superrxan:


internal/server/rule_manager.go
```diff
@@ -83,12 +86,21 @@ func createRule(name, ruleJson string) (string, error) {
...
+	if panicOrError != nil {
 		// Do not store to registry so also delete the KV
 		deleteRule(r.Id)
 		_, _ = ruleProcessor.ExecDrop(r.Id)
-		return r.Id, fmt.Errorf("create rule topo error: %v", err)
+		return r.Id, fmt.Errorf("create rule topo error: %v", panicOrError)
+	}
+
+	// Store to KV
+	err = ruleProcessor.ExecCreate(r.Id, ruleJson)
```
ngjaying, line 101: Why store twice?

ngjaying (approved):

PR #1906, opened 2023-05-17
ci(codecov): fix codecov config

xjasonlyu:


codecov.yaml
```diff
@@ -1,9 +1,12 @@
-codecov:
-  disable_default_path_fixes: true
-
-fixes:
-  - "github.com/lf-edge/ekuiper::"
+coverage:
+  precision: 2
+  round: nearest
+  range: "60...75"
```
ngjaying, line 4: Does this mean coverage should between this range? Why need an upper bound 75?

ngjaying (approved):

PR #1944, opened 2023-05-26
feat: file soure support batch

L-607:
file soure support batch reading multiple files at the same time


ngjaying:
Thanks @L-607. Please follow the link in DCO check to pass it.

Do you encounter a scenario to read files simultaneously?
What about adding a parameter to switch this behaviour?

PR #1946, opened 2023-05-26
feat: support column pruning in deserialization

Yisaer:


internal/converter/json/converter.go
```diff
@@ -166,47 +219,87 @@ func (f *FastJsonConverter) decodeArray(array []*fastjson.Value, field *ast.Json
 func (f *FastJsonConverter) decodeObject(obj *fastjson.Object, schema map[string]*ast.JsonStreamField) (map[string]interface{}, error) {
 	m := make(map[string]interface{})
 	for key, field := range schema {
+		if obj.Get(key) == nil {
```
ngjaying, line 222: Generally, I prefer to do the type conversion/validation here together with decode just like protobuf decode. fastjson.Value has the type information and it is efficient to get the type. If we do it later, we will need reflection to get the real type, which is very slow.

Surely, it is OK to do this in the next PR

internal/converter/converter.go
```diff
@@ -47,6 +47,10 @@ func GetOrCreateConverter(options *ast.Options) (message.Converter, error) {
 	if t == "" {
 		t = message.FormatJson
 	}
+	if t == message.FormatJson && len(options.Schema) > 0 {
```
ngjaying, line 50: Maybe consider passing this schema to the converter instantiator to eliminate the if branch. So that future converter extensions can leverage the information as well. (Maybe a to-do item for v2?)

internal/io/mqtt/mqtt_source.go
```diff
@@ -176,12 +175,13 @@ func getTuples(ctx api.StreamContext, ms *MQTTSource, env interface{}) []api.Sou
 			}
 		}
 	}
-	results, e := ctx.DecodeIntoList(payload)
+	var results []map[string]interface{}
```
ngjaying, line 178: It seems no change here

internal/topo/node/source_node.go
```diff
@@ -140,6 +146,8 @@ func (m *SourceNode) Open(ctx api.StreamContext, errCh chan<- error) {
 						for {
 							select {
 							case <-ctx.Done():
+								// schema may be changed after we close the topo
```
ngjaying, line 149: The sourceNode shall be destroyed after closing. Find a problem of closing?

internal/topo/node/source_pool.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021-2022 EMQ Technologies Co., Ltd.
+// Copyright 2021-2023 EMQ Technologies Co., Ltd.
```
ngjaying, line 1: Seems no change in the file content

internal/topo/planner/planner.go
```diff
@@ -212,8 +217,14 @@ func transformSourceNode(t *DataSourcePlan, sources []*node.SourceNode, options
 		if len(sources) > 0 {
 			srcNode = getMockSource(sources, string(t.name))
 		}
-		if srcNode == nil {
-			srcNode = node.NewSourceNode(string(t.name), t.streamStmt.StreamType, pp, t.streamStmt.Options, options.SendError)
+		if isSchemaless {
```
ngjaying, line 220: If isSchemaless, t.StreamFields is nil as well? Do we need an "if" here?

ngjaying (commented):
Some comments

internal/topo/node/source_node.go
Yisaer: I add the reason in comment

internal/converter/converter.go
Yisaer: we do conversion to convert number into string, thus we should allow the `typeNumber` for the string. That is to say, this is the part of the conversion job and worked in json converter layer here.

PR #1947, opened 2023-05-26
feat: support for more low-priority array funcs

wangxye:


internal/binder/function/funcs_array.go
```diff
@@ -420,4 +427,174 @@ func registerArrayFunc() {
...
+				if !ok {
+					return fmt.Errorf("unknown built-in function: %s.", funcName), false
+				}
+
+				if fs.fType != ast.FuncTypeScalar {
+					return fmt.Errorf("first argument should be a scalar function."), false
+				}
+				eargs := make([]ast.Expr, len(params))
+				if err := fs.val(nil, eargs); err != nil {
+					return fmt.Errorf("function should accept exactly one argument."), false
```
Yisaer, line 518: I think the error message is not accurate here. 
```suggestion
					return fmt.Errorf("validate function arguments failed"), false
```

internal/binder/function/funcs_array.go
```diff
@@ -420,4 +427,174 @@ func registerArrayFunc() {
...
+
+			var nullReplacement string
+			if len(args) == 3 {
+				nullReplacement, ok = args[2].(string)
+				if !ok {
+					return errorArrayThirdArgumentNotStringError, false
+				}
+			}
+
+			for i, v := range array {
```
Yisaer, line 555: This will give the unexpected output as `'a::b'`for `array_join(['a',null,'b'],':')`

internal/binder/function/funcs_array.go
```diff
@@ -420,4 +427,174 @@ func registerArrayFunc() {
...
+		exec: func(ctx api.FunctionContext, args []interface{}) (interface{}, bool) {
+			array, ok := args[0].([]interface{})
+			if !ok {
+				return errorArrayFirstArgumentNotArrayError, false
+			}
+
+			for i := range array {
+				j := rand.Intn(i + 1)
+				array[i], array[j] = array[j], array[i]
+			}
```
Yisaer, line 592: use `rand.Shuffle` instead.

Yisaer (changes requested):
rest lgtm

internal/binder/function/funcs_array.go
wangxye: Do you mean that only if nullReplacement exists, the null value in the array will be replaced? For example, the result of `array_join(['a',null,'b'],':')` should be: `'a:b'`?

Yisaer: I think so.

Yisaer (approved):

PR #1950, opened 2023-05-29
ci: optimize heavy ci check order

xjasonlyu:


.github/workflows/build_packages.yaml
```diff
@@ -6,9 +6,11 @@ concurrency:
 
 on:
   workflow_run:
-    workflows: ["Run fvt tests","Run test cases"]
+    workflows: [ "Run fvt tests","Run test cases" ]
```
Rory-Z, line 9: Will this cause the workflow to be triggered repeatedly?

.github/workflows/run_test_case.yaml
```diff
@@ -88,3 +90,8 @@ jobs:
       with:
         name: stream.log
         path: log/stream.log
+
+  run_fvt_tests:
+    needs:
+      - run_go_tests
```
ngjaying, line 96: Can we rename run_go_tests to run_unit_tests

.github/workflows/build_packages.yaml
Rory-Z: So they are `And` relationships and it only gets triggered when both workflows("Run fvt tests" and "Run test cases") complete?

xjasonlyu: Yes, this is what we expected.

Rory-Z: Cool, looks good for me

.github/workflows/run_test_case.yaml
```diff
@@ -88,3 +90,8 @@ jobs:
       with:
         name: stream.log
         path: log/stream.log
+
+  run_fvt_tests:
+    needs:
+      - run_unit_tests
+    uses: ./.github/workflows/run_fvt_tests.yaml
```
Rory-Z, line 97: Is it work ? Is this the new feature for Github workflow ?

xjasonlyu: Yes, it works:

Screenshot 2023-05-30 at 4 31 58 PM

This feature exists for a while I guess.

PR #1952, opened 2023-05-30
fix(build): add proxy for make docker

superrxan:


ngjaying:
This is not complied with our vision, so close

PR #1955, opened 2023-05-31
fix: pre-correct lint detected errors

xjasonlyu:


extensions/functions/geohash/geohash.go
```diff
@@ -210,7 +210,7 @@ func (r *geohashDecode) Exec(args []interface{}, _ api.FunctionContext) (interfa
 
 func (r *geohashDecodeInt) Exec(args []interface{}, _ api.FunctionContext) (interface{}, bool) {
 	hash, ok := args[0].(uint64)
-	if !ok || 0 > hash {
+	if !ok {
```
ngjaying, line 213: Looks like logic change. Here it also check if hash < 0

ngjaying (changes requested):

extensions/functions/geohash/geohash.go
xjasonlyu: hash is uint64 type, which means hash always >= 0. So `0 > hash` is always false.

ngjaying (approved):

PR #1957, opened 2023-05-31
ci(lint): enable linter for multiple modules

xjasonlyu:


extensions/functions/labelImage/labelImage.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//nolint:typecheck
```
ngjaying, line 15: Why ignore these 3?

xjasonlyu: the `github.com/mattn/go-tflite` pkg seems to be depended on CGO, which causes error like `xxx.h file not found`.

xjasonlyu: reduced the number of nolint annotation to one.

ngjaying (approved):

PR #1962, opened 2023-06-01
feat: support conversion in fastjson

Yisaer:


ngjaying (commented):
There is a `strictValidation` property when creating a stream. Maybe we need to handle both true/false for that property in another PR.

Yisaer:
> case "string", "bytea", why the code are the same. There are actually different types. bytea is expected to be []byte.

updated.

ngjaying (approved):

PR #1967, opened 2023-06-02
feat: let `startRule` rerun the rule

Yisaer:
let startRule rerun the rule in order to re-optimize the rule, then we can generate the plan with the latest schema.

### DIFF
--- a/internal/server/rule_manager.go
+++ b/internal/server/rule_manager.go
@@ -189,26 +189,7 @@ func reRunRule(name string) error {
 	if !ok {
 		return fmt.Errorf("Rule %s is not found in registry, please check if it is created", name)
 	} else {
-		r := rs.Rule
-		var err error
-		deleteRule(name)
-		// Validate the topo
-		panicOrError := infra.SafeRun(func() error {
-			rs, err = createRuleState(r)
-			return err
-		})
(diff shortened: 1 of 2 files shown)


Yisaer:
I think we still need to keep `registry` after craete rule to save `ruleState` in `registry` in order to let us stop rule after starting rule. @ngjaying

internal/server/rule_manager.go
```diff
@@ -180,11 +180,30 @@ func deleteRule(name string) (result string) {
...
+}
+
+// reRunRule rerun the rule from optimize to Open the operator in order to refresh the schema
+func reRunRule(name string) error {
 	rs, ok := registry.Load(name)
 	if !ok {
 		return fmt.Errorf("Rule %s is not found in registry, please check if it is created", name)
 	} else {
-		err := rs.Start()
+		r := rs.Rule
```
ngjaying, line 192: Looks like you only want to re-construct the topo of the rule. It seems not need to change the rule registry? Just like updateRule, just need to updateTopo.

ngjaying (commented):
A suggestion.

internal/server/rule_manager.go
Yisaer: updated.

PR #1971, opened 2023-06-05
feat: add nil check for functions

Yisaer:


internal/binder/function/funcs_agg.go
```diff
@@ -260,6 +290,11 @@ func registerAggFunc() {
 	builtins["var"] = builtinFunc{
 		fType: ast.FuncTypeAgg,
 		exec: func(ctx api.FunctionContext, args []interface{}) (interface{}, bool) {
+			for _, arg := range args {
```
ngjaying, line 293: Looks like this is the same for all func, can we add this to callValuer: check if any arg is nil, if find any, just return nil

Yisaer: updated.

PR #1972, opened 2023-06-05
fix: fix walk function

Rui-Gan:
```
SELECT *,  CASE WHEN ts - lag(ts) OVER (WHEN temperature < 0 AND lag(temperature) > 0) BETWEEN 1000 AND 1100 THEN 1 ELSE 0 END AS condition FROM mockStream
```
In the given example, the presence of only a "when" clause implies that the "walk" function needs to account for this specific case.


ngjaying (changes requested):
Can we add a ut?

Rui-Gan:
> Can we add a ut?

Ok, I will add it.

ngjaying (approved):

PR #1977, opened 2023-06-07
fix: add safe directory for build pkg

xjasonlyu:


.github/workflows/build_packages.yaml
```diff
@@ -38,11 +38,7 @@ jobs:
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
-    - uses: docker/setup-buildx-action@v2
```
Rory-Z, line 41: I think we need keep this line

xjasonlyu: This line is duplicated. We have `setup-buildx-action@v2` after `setup-qemu-action@v2` at line 42.

PR #1985, opened 2023-06-09
fix doc for function echo

Patrick-Star125:


ngjaying:
Thank you. Please follow the Details link after DCO check to pass it. Also, would you mind a check for other json files to check if the links are valid?

ngjaying (approved):

ngjaying:
Merged, thank you!

PR #1995, opened 2023-06-14
feat: add grafana panel

Yisaer:
This panel included 4 parts:
1. server
2. operator
3. sink
4. source

image

In server part, we can find the memory/cpu/goroutines/uptime info:

image

In operator/sink/source part, we can find records in/out cps, latency P90/P99 and exceptions info:

image

We can also use variable to switch the metrics for ekuiper instance and rule:

image


Yisaer:
> Are there a best practice about where to put this file?

updated.

ngjaying (approved):

PR #2005, opened 2023-06-19
feat: support delay slidingWindow

Yisaer:


internal/topo/node/watermark_op.go
```diff
@@ -214,7 +217,17 @@ func (w *WatermarkOp) addAndTrigger(ctx api.StreamContext, d *xsql.Tuple) {
 		}
 		// Update watermark
 		if w.sendWatermark {
-			_ = w.Broadcast(&xsql.WatermarkTuple{Timestamp: watermark})
+			if w.delay > 0 {
```
ngjaying, line 220: This is wrong. Watermark has nothing to do with the processing time (real time). It only deals with the event time. This op is dedicated to send out watermark tuple (tick) and do not need to deal with logics like delay.

ngjaying (changes requested):
For event time, do not handle delay in watermark_op. Handle it in event_timer_triggerer. Addtitionally, we should add test cases in topo_test, testWidnow and testEventWindow.

ngjaying:
Please add doc as well

Yisaer:
> Please add doc as well

updated.

internal/topo/node/event_window_trigger.go
```diff
@@ -151,7 +158,11 @@ func (o *WindowOperator) execEventWindow(ctx api.StreamContext, inputs []*xsql.T
 						o.triggerTime = inputs[0].Timestamp
 					}
 					if windowEndTs > 0 {
-						inputs = o.scan(inputs, windowEndTs, ctx)
+						if o.window.Delay > 0 && o.window.Type == ast.SLIDING_WINDOW {
```
ngjaying, line 161: Is it better to calculate the window end in the getNextWindow() to unify the processing?

ngjaying (commented):
A question about event window imp

ngjaying (approved):
Great work!

PR #2006, opened 2023-06-20
fix(doc): unicode issue

LenaLenaPan:


ngjaying (changes requested):
Need to rebase with the master and remove the first commit(mine).

PR #2008, opened 2023-06-20
feat: support slidingwindow condition

Yisaer:


internal/topo/node/window_op.go
```diff
@@ -89,6 +92,10 @@ func NewWindowOp(name string, w WindowConfig, options *api.RuleOption) (*WindowO
 			o.trigger = w
 		}
 	}
+	if w.TriggerCondition != nil {
+		o.triggerCondition = &operator.FilterOp{Condition: w.TriggerCondition}
```
ngjaying, line 96: Just use valuer to evaluate the condition. No need for an operation.

ngjaying (changes requested):

ngjaying (approved):
Good job

PR #2010, opened 2023-06-20
perf: Avoid string copying during JSON parsing

huskar-t:
This submission implements a performance optimization for JSON parsing by avoiding the copying of strings into byte arrays.

### DIFF
--- a/internal/hack/hack.go
+++ b/internal/hack/hack.go
@@ -1,0 +1,7 @@
+package hack
+
+import "unsafe"
+
+func StringToBytes(str string) []byte {
+	return unsafe.Slice(unsafe.StringData(str), len(str))
+}
--- a/cmd/kuiper/main.go
+++ b/cmd/kuiper/main.go
@@ -27,6 +27,7 @@ import (
 	"github.com/urfave/cli"
 
 	"github.com/lf-edge/ekuiper/internal/conf"
+	"github.com/lf-edge/ekuiper/internal/hack"
 	"github.com/lf-edge/ekuiper/internal/pkg/model"
 	"github.com/lf-edge/ekuiper/pkg/infra"
 )
@@ -1104,7 +1105,7 @@ func main() {
 						rulesArray := c.String("rules")
 						if rulesArray != "" {
 							var rules []string
-							err := json.Unmarshal([]byte(rulesArray), &rules)
+							err := json.Unmarshal(hack.StringToBytes(rulesArray), &rules)
 							if err != nil {
 								fmt.Printf("rules %s unmarshal error %s", rules, err)
 								return nil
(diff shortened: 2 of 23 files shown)


ngjaying:
Could you please also paste your benchmark result for the hack package.

huskar-t:
> Could you please also paste your benchmark result for the hack package.

image

ngjaying:
> > Could you please also paste your benchmark result for the hack package.
> 
> image

Amazing improvement! Are there any risks? Additionally, is it better to rename the package 'hack'? How do you like to move it to cast package, which already have functions of type conversions.

huskar-t:
@ngjaying 
Since Go strings are immutable, the byte arrays must be immutable.
```go
func main() {
	str := "abc"
	a := string2bytes(str)
	fmt.Println("len:", len(a))
	fmt.Println("cap:", cap(a))
	fmt.Println(a)
	a[0] = 'd'
	fmt.Println(a)
	fmt.Println(str)
}

func string2bytes(str string) []byte {
	return unsafe.Slice(unsafe.StringData(str), len(str))
}
// output:
// len: 3
// cap: 3
// [97 98 99]

// segment fault at a[0] = 'd'
```
it can be used if the byte array is read-only

huskar-t:
@ngjaying 
Because len=cap, the append operation will expand the byte array, so the append operation is safe.
```go
func main() {
	str := "abc"
	a := string2bytes(str)
	fmt.Println("len:", len(a))
	fmt.Println("cap:", cap(a))
	fmt.Println(a)
	fmt.Printf("address of bytes before append : %x\n", (*reflect.SliceHeader)(unsafe.Pointer(&a)).Data)
	a = append(a, 'e')
	fmt.Printf("address of bytes after  append : %x\n", (*reflect.SliceHeader)(unsafe.Pointer(&a)).Data)
	a[0] = 'd'
	fmt.Printf("address of bytes after  change : %x\n", (*reflect.SliceHeader)(unsafe.Pointer(&a)).Data)
	fmt.Println(len(a))
	fmt.Println(cap(a))
	fmt.Println(a)
	fmt.Println(str)
}

func string2bytes(str string) []byte {
	return unsafe.Slice(unsafe.StringData(str), len(str))
}

//output
//len: 3
//cap: 3
//[97 98 99]
//address of bytes before append : ca7ea5
//address of bytes after  append : c00001c0f8
//address of bytes after  change : c00001c0f8
//4
//8
//[100 98 99 101]
//abc
```

internal/hack/hack.go
```diff
@@ -0,0 +1,7 @@
+package hack
```
ngjaying, line 1: Suggest to move this to pkg/cast package.

ngjaying (changes requested):

internal/hack/hack.go
huskar-t: done

ngjaying (approved):

ngjaying:
Thank you!

PR #2013, opened 2023-06-21
feat(function): add hex/int conversion functions

L-607:
feat(function): add hex/int conversion functions

### DIFF
--- a/docs/en_US/sqls/functions/transform_functions.md
+++ b/docs/en_US/sqls/functions/transform_functions.md
@@ -73,3 +73,19 @@ chr(col)
 ```
 
 Returns the ASCII character that corresponds to the given Int argument.
+
+## HEX2DEC
+
+```text
+hex2dec(col)
+```
+
+Returns the decimal value of the given hexadecimal string. The data type of the parameter needs to be string. If the parameter is `"0x10"` or `"10"`, convert it to `16`.
+
+## DEC2HEX
+
+```text
+dec2hex(col)
+```
+
+Returns the hexadecimal string of the given Int type decimal, if the parameter is `16`, convert it to `"0x10"`.
--- a/docs/zh_CN/sqls/functions/transform_functions.md
+++ b/docs/zh_CN/sqls/functions/transform_functions.md
@@ -6,7 +6,6 @@
 
 ```text
 cast(col,  "bigint")
-``
 ```
 
 将值从一种数据类型转换为另一种数据类型。 支持的类型包括：bigint，float，string，boolean，bytea 和 datetime。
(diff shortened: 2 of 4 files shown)


docs/en_US/sqls/functions/string_functions.md
```diff
@@ -188,4 +188,20 @@ Removes all leading and trailing whitespace (tabs and spaces) from the provided
 upper(col)
 ```
 
-Returns the uppercase version of the given string.
\ No newline at end of file
+Returns the uppercase version of the given string.
```
ngjaying, line 191: Suggest to document this in transform_functions. Also, use uppercase in the title like ##HEX2DEC

Also, need to point out that the hex here must have `0x` prefix.

internal/binder/function/funcs_str.go
```diff
@@ -357,4 +358,30 @@ func registerStrFunc() {
 		val:   ValidateOneStrArg,
 		check: returnNilIfHasAnyNil,
 	}
+	builtins["hex2dec"] = builtinFunc{
```
ngjaying, line 361: Suggest to move to transform functions

internal/binder/function/funcs_str.go
```diff
@@ -357,4 +358,30 @@ func registerStrFunc() {
...
+		fType: ast.FuncTypeScalar,
+		exec: func(ctx api.FunctionContext, args []interface{}) (interface{}, bool) {
+			arg0 := strings.TrimPrefix(cast.ToStringAlways(args[0]), "0x")
```
ngjaying, line 364: Do not use ToStringAlways, it will convert all types. We should only accept string type here.

internal/binder/function/funcs_str.go
```diff
@@ -357,4 +358,30 @@ func registerStrFunc() {
...
+		check: returnNilIfHasAnyNil,
+	}
+	builtins["dec2hex"] = builtinFunc{
+		fType: ast.FuncTypeScalar,
+		exec: func(ctx api.FunctionContext, args []interface{}) (interface{}, bool) {
+			arg0, err := cast.ToInt(args[0], cast.STRICT)
+			if err != nil {
+				return err, false
+			}
+			hex := "0x" + strconv.FormatInt(int64(arg0), 16)
```
ngjaying, line 381: hexStr := fmt.Sprintf("0x%X", num).  Maybe do a benchmark to check which one is faster.

ngjaying (changes requested):

ngjaying:
And the commit title should be more specific like "feat(function): add hex/int conversion functions"

ngjaying (approved):

PR #2025, opened 2023-06-26
fix(log): use `%v` to format interface{} value

xjasonlyu:


ngjaying:
This kind of problems exist in many places, like filterOp. Is there a way to lint it.

xjasonlyu:
No, there's no lint can detect this issue so far. It's difficult to know if a `func(string, ...any)` function is a printf-style function.

xjasonlyu:
> This kind of problems exist in many places, like filterOp.

I've manually fixed these issues.

ngjaying (approved):

PR #2026, opened 2023-06-26
refactor: use conf.Log for maxprocs.Set

xjasonlyu:


internal/server/server.go
```diff
@@ -87,7 +89,13 @@ func StartUp(Version, LoadFileType string) {
 	conf.InitConf()
 	factory.InitClientsFactory()
 
-	err := store.SetupWithKuiperConfig(conf.Config)
+	undo, err := maxprocs.Set(maxprocs.Logger(conf.Log.Infof))
+	if err != nil {
+		panic(err)
```
ngjaying, line 94: Do we really need to panic here? We can continue without it, right?

ngjaying (changes requested):

internal/server/server.go
xjasonlyu: Done.

ngjaying (approved):

PR #2032, opened 2023-06-26
feat: implement string function of expr

xuyukeviki:


ngjaying:
Thank you!. Is this part of explain statement implementation? Please add unit tests.

xuyukeviki:
> Thank you!. Is this part of explain statement implementation? Please add unit tests.

yep, I will add unit tests later. This pr is aimed to provide string function of expr so that we can explain logical plan easily

Yisaer (approved):

pkg/ast/statement.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021-2023 EMQ Technologies Co., Ltd.
```
Yisaer, line 1: This shouldn't be changed.

pkg/ast/expr.go
```diff
@@ -26,6 +31,7 @@ type NameNode interface {
 type Expr interface {
 	Node
 	expr()
+	String() string
```
Yisaer, line 34: I think `String` is not clear to describe what this function does，plz also add the comment

PR #2041, opened 2023-06-28
feat: add extended wildcard

Rui-Gan:


internal/xsql/valuer.go
```diff
@@ -462,7 +462,21 @@ func (v *ValuerEval) Eval(expr ast.Expr) interface{} {
...
 		}
 	case *ast.Wildcard:
-		val, _ := v.Valuer.Value("*", "")
+		all, _ := v.Valuer.Value("*", "")
+		val, ok := all.(Message)
+		if !ok {
+			return fmt.Errorf("unexpected wildcard value %v", all)
+		}
+		for _, key := range expr.Except {
+			delete(val, key)
```
ngjaying, line 471: Will this affect the original map which is supposed to be readonly?

Rui-Gan: Yes, it has the potential to affect message. I will fix it.

internal/xsql/row.go
```diff
@@ -442,6 +442,10 @@ func (t *Tuple) Pick(allWildcard bool, cols [][]string, wildcardEmitters map[str
 			// invalidate cache, will calculate again
 			t.cachedMap = nil
 		}
+	} else if len(except) > 0 {
```
ngjaying, line 445: Do not modify Message.

internal/topo/planner/projectPlan.go
```diff
@@ -41,6 +42,11 @@
 			switch ft := field.Expr.(type) {
 			case *ast.Wildcard:
 				p.allWildcard = true
+				p.exceptNames = ft.Except
```
ngjaying, line 45: The pruneColumns should be affected. This can be dealt with by another PR

ngjaying (changes requested):
some problems, please check the comments

Rui-Gan:
Ok, I will complete the other two TODOs in the upcoming PR.

internal/xsql/row.go
```diff
@@ -442,6 +442,15 @@ func (t *Tuple) Pick(allWildcard bool, cols [][]string, wildcardEmitters map[str
...
+		pickedMap := make(map[string]interface{})
+		messages := map[string]interface{}(t.Message)
```
ngjaying, line 447: This seems unnecessary?

ngjaying (commented):
one comment

ngjaying (approved):

PR #2046, opened 2023-06-30
fix: support gc logs when started

Yisaer:


ngjaying:
Would you like to add some tests for errors to pass codecov check?

superrxan (approved):

PR #2048, opened 2023-06-30
docs: fix doc error

Yisaer:


ngjaying (changes requested):
Would it be better to have more events in the diagram? Right now, it is not very clear because two window have the same content.

Yisaer:
> Would it be better to have more events in the diagram? Right now, it is not very clear because two window have the same content.

updated

ngjaying (approved):

PR #2050, opened 2023-06-30
perf: reduce data duplication

huskar-t:


ngjaying (approved):
Thank you buddy, perfect work

PR #2059, opened 2023-07-04
feat: export config path

### LINKED ISSUE #1007: Hardcoded path to store dependent files causes issues in Yocto setups
**Environment**:

- eKuiper version (e.g. `1.3.1`): 
- Hardware configuration (e.g. `lscpu`): KVM x86_64
- OS (e.g. `cat /etc/os-release`): Linux 5.4.x
- Others:

When we are building eKuiper along with EdgeX components using Dunfell Yocto recipe, we are facing issues as many installable files are conflicting with files already installed due to other recipes.
eKuiper try to put its files into /etc/services path and /etc/functions path. it has hardcoded this all over.
in our case systemd is adding '/etc/services' as a file with network port information.
some of our recipe is adding '/etc/functions' as a file.

We are building Yocto build of Linux with Dunfell and systemd enabled.

```
Error: Transaction check error:
  file /etc/services conflicts between attempted installs of netbase-2:6.1-r0.core2_64 and rules-engine-1.3.1-r1.core2_64
  file /etc/functions conflicts between attempted installs of px-red-rootfs-base-1.0+git0+4cc4fd9ad7_4fa63909c4-21r7.noarch and rules-engine-1.3.1-r1.core2_64
```

Please setup environment variables which can be configured about the path to store its files.
Or please define its own directory to keep its content like '/etc/ekuiper'

Yisaer:
Closes #1007

### DIFF
--- a/cmd/kuiper/main.go
+++ b/cmd/kuiper/main.go
@@ -52,13 +52,10 @@ func streamProcess(client *rpc.Client, args string) {
 	}
 }
 
-var (
-	Version      = "unknown"
-	LoadFileType = "relative"
-)
+var Version = "unknown"
 
 func main() {
-	conf.LoadFileType = LoadFileType
+	conf.PathConfig.LoadFileType = "relative"
 	app := cli.NewApp()
 	app.Version = Version
 
--- a/cmd/kuiper/main.go
+++ b/cmd/kuiper/main.go
@@ -61,17 +61,11 @@ var (
 var (
 	loadFileType string
 	etcPath      string
-	dataPath     string
-	logPath      string
-	pluginsPath  string
 )
 
 func init() {
 	flag.StringVar(&loadFileType, "loadFileTye", "", "loadFileType indicates the how to load path")
 	flag.StringVar(&etcPath, "etc", "", "etc indicates the path of etc dir")
-	flag.StringVar(&dataPath, "data", "", "data indicates the path of data dir")
-	flag.StringVar(&logPath, "log", "", "log indicates the path of log dir")
-	flag.StringVar(&pluginsPath, "plugins", "", "plugins indicates the path of plugins dir")
 	if len(loadFileType) > 0 {
 		conf.PathConfig.LoadFileType = loadFileType
 	} else {
@@ -79,15 +73,6 @@ func init() {
 	}
 	if len(etcPath) > 0 {
 		conf.PathConfig.EtcDir = etcPath
-	}
-	if len(dataPath) > 0 {
-		conf.PathConfig.DataDir = dataPath
-	}
-	if len(logPath) > 0 {
-		conf.PathConfig.LogDir = logPath
-	}
-	if len(pluginsPath) > 0 {
-		conf.PathConfig.PluginsDir = pluginsPath
 	}
 }
 
--- a/cmd/kuiper/main.go
+++ b/cmd/kuiper/main.go
@@ -92,7 +92,6 @@ func init() {
 }
 
 func main() {
-	conf.PathConfig.LoadFileType = "relative"
 	app := cli.NewApp()
 	app.Version = Version
 
--- a/cmd/kuiper/main.go
+++ b/cmd/kuiper/main.go
@@ -72,7 +72,7 @@ func init() {
 		conf.PathConfig.LoadFileType = LoadFileType
 	}
 	if len(etcPath) > 0 {
-		conf.PathConfig.EtcDir = etcPath
+		conf.PathConfig.Dirs["etc"] = etcPath
 	}
 }
 
--- a/cmd/kuiper/main.go
+++ b/cmd/kuiper/main.go
@@ -53,7 +53,10 @@ func streamProcess(client *rpc.Client, args string) {
 	}
 }
 
-var Version = "unknown"
+var (
+	Version      = "unknown"
+	LoadFileType = "relative"
+)
 
 var (
 	loadFileType string
@@ -71,6 +74,8 @@ func init() {
 	flag.StringVar(&pluginsPath, "plugins", "", "plugins indicates the path of plugins dir")
 	if len(loadFileType) > 0 {
 		conf.PathConfig.LoadFileType = loadFileType
+	} else {
+		conf.PathConfig.LoadFileType = LoadFileType
 	}
 	if len(etcPath) > 0 {
 		conf.PathConfig.EtcDir = etcPath
--- a/cmd/kuiper/main.go
+++ b/cmd/kuiper/main.go
@@ -17,6 +17,7 @@ package main
 import (
 	"bufio"
 	"encoding/json"
+	"flag"
 	"fmt"
 	"net/rpc"
 	"os"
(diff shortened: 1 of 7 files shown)


cmd/kuiperd/main.go
```diff
@@ -15,14 +15,45 @@
 package main
 
 import (
+	"flag"
+
+	"github.com/lf-edge/ekuiper/internal/conf"
 	"github.com/lf-edge/ekuiper/internal/server"
 )
 
+var Version = "unknown"
```
ngjaying, line 24: LoadFileType must not be deleted. It is actually passed in compile time. Check MakeFile and search LoadFileType. If we remove this, it will be a breaking change.

ngjaying (changes requested):
We should keep the LoadFileType var in kuiperd.

ngjaying (commented):
Please try to add more ut to pass 70% coverage

cmd/kuiper/main.go
```diff
@@ -57,8 +58,40 @@ var (
...
+var (
+	loadFileType string
+	etcPath      string
+	dataPath     string
+	logPath      string
+	pluginsPath  string
+)
+
+func init() {
+	flag.StringVar(&loadFileType, "loadFileTye", "", "loadFileType indicates the how to load path")
```
ngjaying, line 70: kuiper client only needs the etc path.

cmd/kuiperd/main.go
```diff
@@ -23,6 +26,39 @@ var (
...
+)
+
+func init() {
+	flag.StringVar(&loadFileType, "loadFileTye", "", "loadFileType indicates the how to load path")
+	flag.StringVar(&etcPath, "etc", "", "etc indicates the path of etc dir")
+	flag.StringVar(&dataPath, "data", "", "data indicates the path of data dir")
+	flag.StringVar(&logPath, "log", "", "log indicates the path of log dir")
+	flag.StringVar(&pluginsPath, "plugins", "", "plugins indicates the path of plugins dir")
+	if len(loadFileType) > 0 {
+		conf.PathConfig.LoadFileType = loadFileType
```
ngjaying, line 44: If/else login are the same?

cmd/kuiperd/main.go
```diff
@@ -23,6 +26,39 @@ var (
...
+	} else {
+		conf.PathConfig.LoadFileType = LoadFileType
+	}
+	if len(etcPath) > 0 {
+		conf.PathConfig.EtcDir = etcPath
```
ngjaying, line 49: What's the default value if not set?

docs/en_US/configuration/configuration.md
```diff
@@ -34,3 +34,15 @@ CONNECTION__EDGEX__REDISMSGBUS__PORT => edgex.redismsgbus.port int etc/connectio
...
 
 The environment variables are separated by "__", the content of the first part after the separation matches the file name of the configuration file, and the remaining content matches the different levels of the configuration items. The file name could be `KUIPER` and `MQTT_SOURCE` in the `etc` folder; or  `CONNECTION` in `etc/connection` folder. Otherwise, the file should in `etc/sources` folder.
+
+### command line parameters
+
+eKuiper supports importing configuration from command line parameters, as follows:
+
+| configuration name | type | configuration role |
+|-----------------|--------|---------------------- --------------------------------|
+| loadFileTye | string | Set the way to load files, support "relative" and "absolute" |
```
ngjaying, line 44: Typo here

docs/en_US/configuration/configuration.md
```diff
@@ -34,3 +34,15 @@ CONNECTION__EDGEX__REDISMSGBUS__PORT => edgex.redismsgbus.port int etc/connectio
...
+| etc | string | Set absolute path of etc directory |
```
ngjaying, line 45: Need to explicitly point out the following paths are only supported in absolute mode.

docs/en_US/configuration/configuration.md
```diff
@@ -34,3 +34,15 @@ CONNECTION__EDGEX__REDISMSGBUS__PORT => edgex.redismsgbus.port int etc/connectio
 ```
 
 The environment variables are separated by "__", the content of the first part after the separation matches the file name of the configuration file, and the remaining content matches the different levels of the configuration items. The file name could be `KUIPER` and `MQTT_SOURCE` in the `etc` folder; or  `CONNECTION` in `etc/connection` folder. Otherwise, the file should in `etc/sources` folder.
+
+### command line parameters
+
+eKuiper supports importing configuration from command line parameters, as follows:
```
ngjaying, line 40: Can we give an example here?

internal/conf/path.go
```diff
@@ -32,7 +44,7 @@
 )
 
 var (
-	LoadFileType    = "relative"
+	PathConfig      PathConfigure
```
ngjaying, line 47: The AbsoluteMapping values are not used? Expect them to be the default path.

ngjaying (changes requested):

cmd/kuiperd/main.go
Yisaer: it's `loadFileType` and `LoadFileType`

docs/en_US/configuration/configuration.md
```diff
@@ -34,3 +34,21 @@ CONNECTION__EDGEX__REDISMSGBUS__PORT => edgex.redismsgbus.port int etc/connectio
...
+
+| configuration name | type | configuration role |
+|-----------------|--------|---------------------- --------------------------------|
+| loadFileType | string | Set the way to load files, support "relative" and "absolute" |
+| etc | string | Set absolute path of etc directory, only valid when loadFileType is "absolute" |
+| data | string | Set the absolute path of the data directory, only valid when loadFileType is "absolute" |
+| log | string | Set the absolute path of the log directory, only valid when loadFileType is "absolute" |
+| plugins | string | Set the absolute path of the plugins directory, only valid when loadFileType is "absolute" |
+
+举例如下:
```
ngjaying, line 50: Need translation

docs/zh_CN/configuration/configuration.md
```diff
@@ -35,3 +35,21 @@ CONNECTION__EDGEX__REDISMSGBUS__PORT => edgex.redismsgbus.port int etc/connectio
...
+
+| 配置名           | 类型   |       配置作用                                        |
+|-----------------|--------|-----------------------------------------------------|
+| loadFileType     | string | 设置加载文件的方式，支持 "relative" 与 "absolute" 两种方式      |
+| etc             | string | 设置 etc 目录的绝对路径，只有当 loadFileType 是 "absolute" 时有效                                |
+| data            | string | 设置 data 目录的绝对路径，只有当 loadFileType 是 "absolute" 时有效                               |
+| log             | string | 设置 log 目录的绝对路径，只有当 loadFileType 是 "absolute" 时有效                                |
+| plugins         | string | 设置 plugins 目录的绝对路径，只有当 loadFileType 是 "absolute" 时有效                            |
+
+example:
```
ngjaying, line 51: Need translation

internal/conf/path.go
```diff
@@ -69,6 +85,16 @@ func GetPluginsLoc() (string, error) {
 
 func absolutePath(loc string) (dir string, err error) {
 	for relDir, absoluteDir := range AbsoluteMapping {
+		switch loc {
+		case etcDir:
```
ngjaying, line 89: This is wrong. The loc could be somthing like "etc/sources/abc.yaml" not just "etc". The test case only covers when loc==etcDir case. ,

Previously, AbsoluteMapping is a const. Generally, we just need to let it point to the flag value if is explicitly set. No need to touch the logic here.

ngjaying (changes requested):

internal/conf/path.go
Yisaer: updated.

internal/conf/path_test.go
```diff
@@ -58,3 +62,49 @@ func TestGetDataLoc_Funcs(t *testing.T) {
...
+	PathConfig.Dirs["data"] = "/data/kuiper"
+	PathConfig.Dirs["log"] = "/log/kuiper"
+	PathConfig.Dirs["plugins"] = "/tmp/plugins"
+
+	testcases := []struct {
+		dir    string
+		expect string
+	}{
+		{
+			dir:    etcDir,
```
ngjaying, line 88: Suggest to test deeper dir like etcDir/test

internal/conf/path_test.go
```diff
@@ -38,6 +40,8 @@ func TestAbsolutePath(t *testing.T) {
 			a: "/var/lib/kuiper/plugins",
 		},
 	}
+	initAbsolutePathConfig()
```
ngjaying, line 43: Suggest to remove the adding lines to test the default behavior

ngjaying (commented):
Add test suggestions

ngjaying (approved):
Perfect!

PR #2060, opened 2023-07-04
fix: wildcard contains alias

Rui-Gan:
rule:
```
select collect(*), a AS ab from demo GROUP BY ab, TUMBLINGWINDOW(ss, 10);
```
output:
```
{
  "ab": 2
  "collect": [{"a": 2, "ab": 2, "b": 1}]
}
```
This pr is primarily aimed at addressing the issue.

### DIFF
--- a/internal/xsql/collection.go
+++ b/internal/xsql/collection.go
@@ -223,7 +223,11 @@ func (w *WindowTuples) Meta(key, table string) (interface{}, bool) {
 }
 
 func (w *WindowTuples) All(_ string) (map[string]interface{}, bool) {
-	return w.ToMap(), true
+	if len(w.Content) == 0 {
+		m := make(map[string]interface{})
+		return m, true
+	}
+	return w.Content[0].All("")
 }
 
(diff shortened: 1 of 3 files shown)


internal/xsql/collection.go
```diff
@@ -223,7 +223,11 @@ func (w *WindowTuples) Meta(key, table string) (interface{}, bool) {
 }
 
 func (w *WindowTuples) All(_ string) (map[string]interface{}, bool) {
-	return w.ToMap(), true
+	if len(w.Content) == 0 {
+		m := make(map[string]interface{})
+		return m, true
+	}
+	return w.Content[0].All("")
```
ngjaying, line 230: Currently, select * for window will return all rows. Will this change that behavior?

ngjaying (commented):
Have a question about collection wildcard.

internal/xsql/collection.go
Rui-Gan: In the ToMap function, only w.Content[0] is being selected, so I don't think it will have any impact on the behavior.

ngjaying (approved):

PR #2064, opened 2023-07-05
feat: add timezone support

xjasonlyu:


ngjaying (changes requested):
What is the default timezone value? Local? Need to add that to the default kuiper.yaml explicitly.  Also, need to add the doc.

xjasonlyu:
> What is the default timezone value? Local? Need to add that to the default kuiper.yaml explicitly. Also, need to add the doc.

The default timezone is UTC. This will prevent behavior changes due to different system’s timezone settings. 

I will open another PR for docs.

docs/zh_CN/configuration/global_configurations.md
```diff
@@ -44,6 +44,15 @@ basic:
...
 用户将名为 KuiperSyslogKey 的环境变量的值设置为 true 时，日志将打印到系统日志中。
 
+## 时区配置
+
+```yaml
+# The global time zone from the IANA time zone database, or UTC if not set.
+timezone: UTC
+```
+
+基于 IANA 时区数据库的全局时区配置，如果留空则使用 UTC 作为默认时区，设置为 Local 时则使用系统时区。
```
ngjaying, line 54: Are there links for IANA?  Is `Local` already supported in the code, I don't find any special treatment for the value.

xjasonlyu: IANA link: https://www.iana.org/time-zones

`Local` is supported by Golang std time pkg:

> Local represents the system's local time zone. On Unix systems, Local consults the TZ environment variable to find the time zone to use.

ngjaying (approved):

PR #2066, opened 2023-07-05
feat: support limit clause

Yisaer:


ngjaying (commented):
Is limit always follow project op? If so, can we make it a property of project op in which we can avoid a lot of unneccessary calculation

ngjaying (commented):
Is it possible to leverage the limit when range over the result instead of copying after all are caculated.

ngjaying (approved):
Feature completed so merged. Consider optimization later.

PR #2067, opened 2023-07-05
fix: fix source processLatency in metrics 

Rui-Gan:


internal/io/file/file_source.go
```diff
@@ -276,7 +275,7 @@ func (fs *FileSource) publish(ctx api.StreamContext, file io.Reader, consumer ch
 		ctx.GetLogger().Debug("Sending tuples")
 		for _, m := range resultMap {
 			select {
-			case consumer <- api.NewDefaultSourceTupleWithTime(m, meta, rcvTime):
+			case consumer <- api.NewDefaultSourceTupleWithTime(m, meta, conf.GetNow()):
```
ngjaying, line 278: Can we let the first recv time begins from the very start?

ngjaying (changes requested):

internal/io/file/file_source.go
Rui-Gan: ok, I will fix it

ngjaying (approved):

PR #2082, opened 2023-07-10
feat(math): add common SQL math funcs/aliases

xjasonlyu:
Added some common MySQL/PostgreSQL-style math functions and its aliases.

Refer to: 
- https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html
- https://www.postgresql.org/docs/9.1/functions-math.html

### DIFF
--- a/internal/binder/function/funcs_array_test.go
+++ b/internal/binder/function/funcs_array_test.go
@@ -562,7 +562,7 @@ func TestArrayCommonFunctions(t *testing.T) {
 			args: []interface{}{
 				"pow", []interface{}{0, -0.4, 1.2},
 			},
-			result: fmt.Errorf("unknown built-in function: pow."),
+			result: fmt.Errorf("validate function arguments failed."),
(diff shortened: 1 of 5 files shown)


internal/binder/function/funcs_array_test.go
```diff
@@ -557,13 +557,6 @@ func TestArrayCommonFunctions(t *testing.T) {
 			},
 			result: []interface{}{0, 0.4, 1.2},
 		},
-		{
-			name: "array_map",
```
ngjaying, line 561: Suggest to replace with a similar case instead of just deleting it

ngjaying (changes requested):
Need ut for new functions

internal/binder/function/funcs_array_test.go
xjasonlyu: Done.

ngjaying (approved):

PR #2085, opened 2023-07-10
feat:  extended wildcard follow up works

Rui-Gan:


ngjaying (commented):
Is it possible to use * except in where clause? It is not tested.

Rui-Gan:
> Is it possible to use * except in where clause? It is not tested.

I add UT using * except and * replace in the Having clause, but I currently cannot think of an example where * except can be used in the Where clause.

internal/topo/planner/planner_test.go
```diff
@@ -1870,6 +1898,444 @@ func Test_createLogicalPlan(t *testing.T) {
 				},
 			}.Init(),
 		},
+		{ // 19
+			sql: `SELECT * EXCEPT(id1, name), meta(device) FROM src1`,
```
ngjaying, line 1902: What about `select name, * EXCEPT(id1, name)`

Rui-Gan: For `select name, * EXCEPT(id1, name)`, streamFields will contain "temp", "myarray", and "name"; similarly, for` SELECT id1 FROM src1 GROUP BY TUMBLINGWINDOW(ss, 10) HAVING count(* EXCEPT(id1, name)) > 0`, streamFields will contain "temp", "myarray", and "id1".

ngjaying (approved):

PR #2086, opened 2023-07-10
feat(math): support log func with base

xjasonlyu:


internal/binder/function/funcs_math.go
```diff
@@ -271,18 +272,38 @@ func registerMathFunc() {
...
+				if e != nil {
+					return e, false
+				}
+				r = math.Log(x) / math.Log(v)
+			}
+
+			if !math.IsNaN(r) {
+				return r, true
+			}
+			return nil, true
```
ngjaying, line 296: What's the expected return for NaN?

xjasonlyu: Any NaN values are supposed to be returned as nil.

> If called with one argument, the function returns the decimal logarithm of X. If X is less than or equal to 0, the function returns nil; if called with two arguments, the function returns the base B logarithm of X. Returns nil if X is less than or equal to 0, or if B is less than or equal to 1.

ngjaying: Is this the "standard" behavior?

xjasonlyu: MySQL’s log function has the similar behavior: https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_log

ngjaying (approved):

PR #2097, opened 2023-07-14
fix(upload): support url for file upload (#23)

superrxan:


ngjaying:
Try to pass codecov target

internal/server/rest.go
```diff
@@ -170,8 +170,50 @@ func createRestServer(ip string, port int, needToken bool) *http.Server {
...
+	if f.Name == "" {
+		return fmt.Errorf("invalid body: name is required")
+	}
+	return nil
+}
+
+func upload(file *fileContent) error {
+	err := getFile(file)
+	if err != nil {
+		return err
```
ngjaying, line 191: Please try to cover these error condtions as well.

ngjaying (changes requested):
LGTM. Hopefully, we can cover the error conditons to make coverage pass 70%

internal/server/rest.go
superrxan: ack

ngjaying (approved):

PR #2098, opened 2023-07-17
fix: exit process after an error occurs

ke4nec:
exit process after an error occurs like bind failed "bind: address already in use"

Signed-off-by: ke4nec <ke4nec@qq.com>


ngjaying (approved):
We suppose use can still use CLI to manage withour REST. After discussions, we think REST API is important enough to fail the starting. Thanks.

PR #2100, opened 2023-07-17
fix(httppull): refresh token before header parsing

xjasonlyu:


internal/io/http/httppull_source.go
```diff
@@ -90,6 +90,14 @@ func (hps *PullSource) doPull(ctx api.StreamContext, rcvTime time.Time, omd5 *st
 		}
 	}
 
+	// check oAuth token expiration
+	if hps.accessConf != nil && hps.accessConf.ExpireInSecond > 0 &&
+		int(time.Now().Sub(hps.tokenLastUpdateAt).Abs().Seconds()) >= hps.accessConf.ExpireInSecond {
+		ctx.GetLogger().Debugf("Refreshing token for HTTP pull")
+		if err := hps.refresh(ctx); err != nil {
+			ctx.GetLogger().Warnf("Refresh HTTP pull token error: %v", err)
```
xjasonlyu, line 98: @ngjaying Wondering if we should return an error here instead of logging a warning? 

If the refreshing process failed, the following `parseHeaders` will use the outdated tokens and fail the pulling process anyway.

ngjaying: Usually, the expire setting have some buffer so that the old token is likely still valid when the process failed. And in the next round, the refresh will retry. So I prefer to keep this.

xjasonlyu: OK, I see.

xjasonlyu:
This PR is ready to merge.

Issue #2090, opened 2023-07-11
Refer to alias in the same select statement


ngjaying:
In standard SQL, alias cannot be referred in the same select statement. We need CTE or subquery to do alias reference which is complex.  Propose to support it to simplify the usage. An example as below, where `sum`, `sum2` are defined and used in the same select clause.

```SQL
SELECT a + b AS sum,
       sum + 2 AS sum2,
       sum2 + 10 AS sum3
FROM my_table;
```

PR #2101, opened 2023-07-17
feat: support ref alias in select

Yisaer:


internal/topo/planner/analyzer_test.go
```diff
@@ -103,11 +103,11 @@ var tests = []struct {
 		r:   newErrorStruct(""),
 	},
 	{ // 10
-		sql: `SELECT sum(temp) as temp, count(temp) as temp FROM src1`,
-		r:   newErrorStruct("duplicate alias temp"),
+		sql: `SELECT sum(temp) as temp1, count(temp) as temp FROM src1`,
```
ngjaying, line 106: Shall we keep both the previous test and the new one?

internal/topo/planner/analyzer.go
```diff
@@ -265,12 +375,13 @@
 
 type fieldsMap struct {
 	content       map[string]streamFieldStore
+	alias         map[string]struct{}
```
ngjaying, line 378: The content should also have the alias information. Is it worth to reuse it?

ngjaying (changes requested):

internal/topo/planner/analyzer.go
Yisaer: We need to collect the aliasName first before bind the schema info for column if they reference the alias, then we can bind the alias info with column schema info.

ngjaying (approved):

ngjaying:
Closes #2090

PR #2113, opened 2023-07-20
fix: analyze funcs reference alias analyze

Yisaer:


ngjaying (changes requested):
Is it possible to increase ut coverage?

Yisaer:
> Is it possible to increase ut coverage?

I think it's too hard to increase ut coverate as most uncoveraged patchs are `return error`

ngjaying (approved):

PR #2125, opened 2023-07-25
feat: support accmulative agg funcs

Yisaer:


internal/binder/function/funcs_agg.go
```diff
@@ -24,6 +25,203 @@
...
+
+func min(a, b float64) float64 {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func registerGlobalAggFunc() {
+	builtins["global_avg"] = builtinFunc{
```
ngjaying, line 53: how about name it "cum_aum" which means cumulative sum?

ngjaying (commented):
Do we support OVER WHEN/PARTITION BY for these functions?

internal/binder/function/funcs_agg.go
Yisaer: I think using `acc` (accumulate) as prefix is more clear.

internal/binder/function/funcs_agg.go
```diff
@@ -24,6 +25,217 @@ import (
...
+
+func min(a, b float64) float64 {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func registerGlobalAggFunc() {
+	builtins["acc_avg"] = builtinFunc{
```
ngjaying, line 43: I think these functions are indeed analytic(stateful) function which keep and update states. Is it better to put them in the analytic functions? Also, they need to support ON Partion By/When clause like the normal analytic functions

ngjaying (changes requested):

internal/binder/function/funcs_agg.go
Yisaer: updated.

internal/binder/function/funcs_analytic.go
```diff
@@ -238,3 +239,259 @@
...
+				}
+				val = float64(math.MinInt64)
+			}
+			m := val.(float64)
+			if !validData {
+				return m, true
+			}
+			switch v := args[0].(type) {
+			case int:
+				v1 := float64(v)
```
ngjaying, line 330: Shall we change the type?

ngjaying (changes requested):
Need docs

ngjaying (approved):

PR #2138, opened 2023-07-31
feat: support separate debug & logpath for rules

xjasonlyu:


xjasonlyu:
@ngjaying This feature has been tested locally and is working as expected.

internal/topo/topo.go
```diff
@@ -120,8 +125,37 @@ func (s *Topo) addEdge(from api.TopNode, to api.TopNode, toType string) {
 // stream starts execution.
 func (s *Topo) prepareContext() {
 	if s.ctx == nil || s.ctx.Err() != nil {
-		contextLogger := conf.Log.WithField("rule", s.name)
-		ctx := kctx.WithValue(kctx.Background(), kctx.LoggerKey, contextLogger)
+		ctxLogger := &logrus.Logger{
```
ngjaying, line 128: Are there cases to reuse logger as before?

xjasonlyu: If we reused logger, any changes for rule logger will affect the global logger, which I believe is not what we expect.

ngjaying: That is the default behavior previously so it should be harmless. With this change, will it have problems when writing to the same file simultaneously by multiple rules?

xjasonlyu: Fixed.

ngjaying (approved):

ngjaying:
@Yisaer Please check if this breaks the log gc

Yisaer (changes requested):
we clear outdated log when restart eKuiper in `gcOutdatedLog` with logPath defined in config.

When we supported rule-level logPath, I think we should also clear the outdated log.

PR #2139, opened 2023-07-31
docs: add governance

Yisaer:


ngjaying:
According to the charter, we will add the roles to contributing file. I will close this PR and move some content there.

PR #2140, opened 2023-07-31
feat: support window function row_number

Yisaer:


docs/en_US/sqls/functions/window_functions.md
```diff
@@ -0,0 +1,11 @@
+# Window Functions
```
ngjaying, line 1: Please also add a link in the parent page.

docs/en_US/sqls/functions/window_functions.md
```diff
@@ -0,0 +1,11 @@
...
+
+A window function performs a calculation across a set of table rows that are somehow related to the current row. This is comparable to the type of calculation that can be done with an aggregate function.
```
ngjaying, line 3: Add a notice that this kind of functions can only be used in the select clause

internal/topo/planner/planner.go
```diff
@@ -463,6 +465,14 @@ func createLogicalPlan(stmt *ast.SelectStatement, opt *api.RuleOption, store kv.
 		p.SetChildren(children)
 		children = []LogicalPlan{p}
 	}
+	windowFuncFields, windowFuncsNames := extractWindowFuncFields(stmt)
```
ngjaying, line 468: Do we support to use window function in order by? Looks like not supported in the planner.

ngjaying (changes requested):

internal/topo/planner/planner.go
Yisaer: we will support it later.

ngjaying (approved):

PR #2142, opened 2023-08-01
fix: make alias reference the same name column

Yisaer:


ngjaying:
LGTM. Please add the doc.

ngjaying (approved):

PR #2163, opened 2023-08-07
docs: add sql example

Yisaer:


ngjaying (changes requested):
The updates should emphasize on the window: Join, group by and having etc. must run in window  for streaming source.

docs/en_US/sqls/query_language_elements.md
```diff
@@ -338,6 +410,12 @@ FROM stream1
 GROUP BY column_name
 ```
 
+example:
+
+```sql
+select * from demo group by a;
```
ngjaying, line 416: need window

docs/en_US/sqls/query_language_elements.md
```diff
@@ -403,6 +493,12 @@ The case expression evaluates a list of conditions and returns one of multiple p
 
 There are two types of case expression: simple case expression and searched case expression.
 
+exmaple:
+
+```sql
+select * from demo where a > 10 limit 10;
```
ngjaying, line 499: Better to have a window?

ngjaying (changes requested):

ngjaying (approved):

PR #2164, opened 2023-08-07
feat: support validate window function in parser

Yisaer:


internal/xsql/sqlValidator.go
```diff
@@ -39,9 +39,32 @@ func Validate(stmt *ast.SelectStatement) error {
...
 		return err
 	}
+	if err := validateWindowFunction(stmt); err != nil {
+		return err
+	}
 	return validateSRFForbidden(stmt)
 }
 
+func validateWindowFunction(stmt *ast.SelectStatement) error {
+	if exits := isWindowFunctionExists(stmt.Joins); exits {
```
ngjaying, line 49: Can we just check the whole stmt？ It seems in isWindowFunctionExist, the fields are skipped. Otherwise, you'll need to check more clause once we add.

ngjaying (changes requested):

ngjaying (approved):

PR #2168, opened 2023-08-10
fix: fix validate api status code

Rui-Gan:


ngjaying:
LGTM. Shall we update the doc?

ngjaying (approved):

PR #2169, opened 2023-08-10
fix: parse time should consider loc

Yisaer:


superrxan (commented):
1. add rulePatrolInterval: "10s" in etc/kuiper.yaml file

pkg/schedule/schedule.go
```diff
@@ -38,6 +39,34 @@ func IsInScheduleRanges(now time.Time, timeRanges []api.DatetimeRange) (bool, er
...
+		if err != nil {
+			return false, err
+		}
+		return isInTimeRangeWithLoc(now, start, end, loc)
+	}
+	return isInTimeRange(now, start, end)
+}
+
+func isInTimeRangeWithLoc(now time.Time, start string, end string, loc *time.Location) (bool, error) {
+	s, err := time.ParseInLocation(layout, start, loc)
```
ngjaying, line 53: Can we use cast.InterfaceToTime? So that you don't need to handle timezone by yourself here. And it can support more time format. It is the universal time cast function in this project.

ngjaying (changes requested):

ngjaying (approved):

Issue #1869, opened 2023-05-10
Call for function implementations


Yisaer:
There are some functions we haven't supported in eKuiper. Here are some functions we listed group by the priority. It's always welcome if someone helps to support any of the unimplemented function.

Here is the example for implementation array and object functions https://github.com/lf-edge/ekuiper/pull/1871

If you have any interest for implementing the above functions, plz comment on this issue first in order to avoid do the duplicated work.

## Array functions

https://rockset.com/docs/array-functions/

### High

 * [x] ARRAY_POSITION
 * [x] LENGTH
 * [x] SLICE
 * [x] ELEMENT_AT
 * [x] ARRAY_CONTAINS
 * [x] ARRAY_SORT
 * [x] ARRAY_CONCAT

### Medium

 * [x] ARRAY_CREATE
 * [x] ARRAY_REMOVE
 * [x] ARRAY_LAST_POSTION
 * [x] ARRAY_CONTAINS_ANY
 * [x] ARRAY_INTERSECT
 * [x] ARRAY_UNION
 * [x] ARRAY_MAX
 * [x] ARRAY_MIN
 * [x] ARRAY_EXCEPT
 * [x] REPEAT
 * [x] SEQUENCE

### LOW

 * [x] CARDINALITY
 * [x] ARRAY_FLATTEN
 * [x] ARRAY_DISTINCT
 * [x] ARRAY_MAP
 * [x] ARRY_JOIN
 * [x] ARRAY_SHUFFLE

## Object functions

https://rockset.com/docs/object-functions/

### High

 * [x] ELEMENT_AT
 * [x] LENGTH
 * [x] KEYS
 * [x] VALUES

### Medium

 * [x] OBJECT
 * [x] ZIP
 * [x] ITEMS
 * [x] MERGE
 * [ ] ERASE

wangxye:
Hi, I am interested in this issue. Can I do this task?

ngjaying:
@wangxye Sure, welcome

Yisaer:
> Hi, I am interested in this issue. Can I do this task?

sure, please declare the functions u want to implement before start coding.

wangxye:
I can start with the medium and low priorities of array functions.

Tannic-Paprika:
Hi, I am interested in solving the high and medium priorities of array functions.

Yisaer:
> Hi, I am interested in solving the high and medium priorities of array functions.

[...]

PR #2177, opened 2023-08-12
feat(func):support array_join

hantmac:
- support array_join
- fix issue https://github.com/lf-edge/ekuiper/issues/1869


hantmac:
@ngjaying Oh thanks, I should search the code before writing 😢

hantmac:
BTW I found many funcs have been implemented but the issue is not updated, plz check.

PR #2179, opened 2023-08-14
feat(func): support func erase

hantmac:
- support `erase` func


ngjaying:
You are so fast, amazing.  Unfortunately, merge is already implemented as name "object_concat". Maybe we can just add erase.

hantmac:
> You are so fast, amazing. Unfortunately, merge is already implemented as name "object_concat". Maybe we can just add erase.

Aha, there is why I have not found `merge` 😢

internal/binder/function/funcs_obj.go
```diff
@@ -155,4 +156,51 @@ func registerObjectFunc() {
...
+			res := make(map[string]interface{})
+			argMap, ok := args[0].(map[string]interface{})
+			if !ok {
+				return fmt.Errorf("the first argument should be map[string]interface{}, got %v", args[0]), false
+			}
+			eraseArray := make([]string, 0)
+			v := reflect.ValueOf(args[1])
+			switch v.Kind() {
+			case reflect.Slice, reflect.Array:
+				array := args[1].([]string)
```
ngjaying, line 182: The args[1] is likely to be []any, the type cast here will fail. You can use cast.ToStringSlice here to do the conversion.

ngjaying (changes requested):

internal/binder/function/funcs_obj.go
hantmac: Thanks! Fixed.

ngjaying: ah, here I mean "github.com/lf-edge/ekuiper/pkg/cast"

hantmac: > ah, here I mean "github.com/lf-edge/ekuiper/pkg/cast"

Thanks the `good first issue` 😅

ngjaying (approved):

Yisaer (approved):

PR #2182, opened 2023-08-14
fix: fix acc agg function for invalid data

Yisaer:


internal/binder/function/funcs_analytic.go
```diff
@@ -262,7 +262,7 @@ func registerGlobalAggFunc() {
 			}
 			if v1 == nil && v2 == nil {
 				if args[0] == nil || !validData {
-					return 0, true
+					return 0, false
```
ngjaying, line 265: if when is eval false, shall we return the previous result instead of just 0; And why return false? I expect to return lastResult, true

ngjaying (changes requested):

internal/binder/function/funcs_analytic.go
```diff
@@ -261,10 +261,9 @@ func registerGlobalAggFunc() {
 				return fmt.Errorf("when arg is not a bool but got %v", args[len(args)-2]), false
 			}
 			if v1 == nil && v2 == nil {
-				if args[0] == nil || !validData {
-					return 0, true
+				if !validData {
+					return 0, false
```
ngjaying, line 265: Shall we return the latest agg value instead of 0 value? Not only this place, all other places have the same problem

ngjaying (changes requested):

PR #2183, opened 2023-08-14
fix: time function should consider timezone

Yisaer:


pkg/cast/time.go
```diff
@@ -70,6 +70,14 @@ func init() {
 	now.TimeFormats = append(now.TimeFormats, JSISO, ISO8601)
 }
 
+func GetLocalTimeZone() *time.Location {
```
ngjaying, line 73: The name is misleading. It is actually to get the configured timezone..

ngjaying (changes requested):

pkg/cast/time.go
```diff
@@ -70,6 +70,14 @@ func init() {
 	now.TimeFormats = append(now.TimeFormats, JSISO, ISO8601)
 }
 
+func GetConfiguredTimeZone() *time.Location {
+	return localTimeZone
+}
+
+func SetLocalTimeZone(l *time.Location) {
```
ngjaying, line 77: There is already a SetTimeZone function below, this seems like a duplication.

ngjaying (changes requested):

pkg/cast/time.go
```diff
@@ -70,6 +70,10 @@ func init() {
 	now.TimeFormats = append(now.TimeFormats, JSISO, ISO8601)
 }
 
+func GetConfiguredTimeZone() *time.Location {
+	return localTimeZone
+}
+
 var localTimeZone = time.UTC
```
ngjaying, line 77: Shall this be local?

ngjaying (changes requested):

ngjaying (approved):

PR #2185, opened 2023-08-14
fix(sink): hang when cache priority set

ngjaying:


internal/topo/node/sink_node.go
```diff
@@ -321,15 +321,17 @@ func (m *SinkNode) Open(ctx api.StreamContext, result chan<- error) {
 										receiveQ(data)
 									case data := <-dataOutCh:
 										normalQ(data)
```
magichan, line 323: the hang problems can also be resolved by add continue keyword 
```go
for {
	select {
	case data := <-m.input:
		receiveQ(data)
	case data := <-dataOutCh:
		normalQ(data)
	case <-ctx.Done():
		doneQ()
		return nil
	default:
		select {
		case data := <-dataOutCh:
			normalQ(data)
		case data := <-rq.Out:
			resendQ(data)
                default:
                      continue
		}
	}
}
```

ngjaying: By this way, it will keep looping. Here we want to block and wait until one channel has data.

magichan: yes, the code would waste cpu.

Yisaer (approved):

PR #2186, opened 2023-08-15
revise: add log for rule checker

Yisaer:


ngjaying:
Suggest to add more logs:

- When the scheduled rule starts to run 
- When the scheduled rule stop running

Because in the rule status, we can only get the instant status.

ngjaying (approved):

PR #2196, opened 2023-08-17
fix: fix schedule rule state

Yisaer:
For schedule rule and long running schedule rule:

| action | state |
|  ----  | ----  |
| start schedule rule  | `Running` |
| stopped by auto scheduling | `Stopped: waiting for next schedule.` |
| stopped by manually stop | `Stopped: canceled manually.` |
| Current time is after all cron datetime ranges | `Stopped: schedule terminated.` |

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -305,7 +305,7 @@ func handleScheduleRuleState(now time.Time, r *api.Rule, state string) error {
 	conf.Log.Debugf("rule %v origin state: %v, sginal: %v", r.Id, state, scheduleActionSignal)
 	switch scheduleActionSignal {
 	case scheduleRuleActionStart:
-		return startRule(r.Id)
+		return startRuleInternal(r.Id)
 	case scheduleRuleActionStop:
 		stopRuleInternal(r.Id)
 	}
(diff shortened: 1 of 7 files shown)


internal/server/server.go
```diff
@@ -306,7 +306,7 @@ func handleScheduleRuleState(now time.Time, r *api.Rule, state string) error {
 	if toStart {
 		return startRule(r.Id)
 	} else if toStop {
-		stopRule(r.Id)
+		stopRule(r.Id, true)
```
ngjaying, line 309: schedule start/stop does not need to update trigger status in the db. Can we call a new function instead of changing the existing function?

internal/server/server.go
```diff
@@ -326,7 +326,7 @@ func handleScheduleRule(now time.Time, r *api.Rule, state string) (bool, bool) {
 				break
 			}
 		}
-		if isInRange && state != "Running" {
+		if isInRange && state != "Running" && state != "Stopped: canceled manually." {
```
ngjaying, line 329: Is the constant working here?

ngjaying (changes requested):

ngjaying (approved):

PR #2212, opened 2023-08-28
fix: fix tumbling window would miss data

Yisaer:


internal/topo/node/window_op.go
```diff
@@ -554,7 +554,7 @@ func (o *WindowOperator) scan(inputs []*xsql.Tuple, triggerTime int64, ctx api.S
 	length := o.window.Length + o.window.Delay
 	// Sync table
 	for _, tuple := range inputs {
-		if o.window.Type == ast.HOPPING_WINDOW || o.window.Type == ast.SLIDING_WINDOW {
+		if o.window.Type == ast.HOPPING_WINDOW || o.window.Type == ast.SLIDING_WINDOW || o.window.Type == ast.TUMBLING_WINDOW {
```
ngjaying, line 557: Tumbling window was handled in else. Does it work to change line 568, > to >=

ngjaying (changes requested):

ngjaying (approved):

PR #2215, opened 2023-08-30
feat: support httppull lookup

Yisaer:


internal/io/http/client.go
```diff
@@ -116,7 +116,7 @@ func (cc *ClientConf) InitConf(device string, props map[string]interface{}) erro
 	default:
 		return fmt.Errorf("Not supported HTTP method %s.", c.Method)
 	}
-	if c.Interval <= 0 {
+	if needInterval && c.Interval <= 0 {
```
ngjaying, line 119: RestSink does not need interval too. There must be some mechanism to bypass this check. Maybe we can reuse that instead of add a new property

internal/io/http/httppull_lookup.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (l *lookupSource) Lookup(ctx api.StreamContext, _ []string, keys []string, values []interface{}) ([]api.SourceTuple, error) {
+	resps, err := l.pull(ctx)
+	if err != nil {
+		return nil, err
+	}
+	matched := l.lookupJoin(resps, keys, values)
+	var results []api.SourceTuple
+	for _, resp := range matched {
+		meta := make(map[string]interface{})
```
ngjaying, line 54: The meta is immutable, we can just reuse it.

ngjaying (changes requested):

ngjaying (approved):

PR #2228, opened 2023-09-01
fix: fix return wrong error object

Yisaer:


internal/xsql/valuer.go
```diff
@@ -777,9 +777,10 @@ func (v *ValuerEval) simpleDataEval(lhs, rhs interface{}, op ast.Token) interfac
 	// Evaluate if both sides are simple types.
 	switch lhs := lhs.(type) {
 	case bool:
-		rhs, ok := rhs.(bool)
+		originRHS := rhs
```
ngjaying, line 780: Does this problem also happen in other case statement?

Yisaer: no

Issue #2233, opened 2023-09-04
The 'triggered' parameter is not valid


winc-link:
The 'triggered' parameter is not valid

**Environment**:

lfedge/ekuiper:1.10.0-slim

**What happened and what you expected to happen**:

I hope to not change the rule state when the 'triggered' parameter is false.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

PR #2236, opened 2023-09-05
fix: fix updateTopo ignore triggered

Yisaer:
fix: https://github.com/lf-edge/ekuiper/issues/2233


internal/topo/rule/ruleState.go
```diff
@@ -384,6 +392,24 @@ func (rs *RuleState) start() error {
 	return nil
 }
 
+func (rs *RuleState) updateTopoWithLock() error {
```
ngjaying, line 395: Why doing this? Need some comment

Yisaer: updated.

ngjaying (approved):

PR #2238, opened 2023-09-06
chore(ci): support multiple versions of documents

Swilder-M:


.github/workflows/deploy_docs.yaml
```diff
@@ -6,6 +6,8 @@ concurrency:
 
 on:
   push:
+    tags:
+      - '*.*.*'
     branches:
       - master
```
ngjaying, line 12: We will update fixpack in mater-x.xx branch, such as master-1.10 or master-1.11. Do we need to add these branches?

Swilder-M: There are two scenarios that trigger the document deployment workflow:

1. push to the master branch, where the documents will be updated to the latest directory.
2. Create a tag, which will be updated to the `vx.y` version of the document.

In your case, the branching doesn't affect the documentation. If an older version of the fix ends up creating a tag, then the documentation will be updated to the corresponding version of the directory, not the latest version.

ngjaying: OK, thanks.

ngjaying:
@Rory-Z Please take a look

Rory-Z (approved):

PR #2261, opened 2023-09-15
fix: fix start rule failed

Yisaer:


internal/server/rule_manager.go
```diff
@@ -153,8 +154,13 @@ func recoverRule(r *api.Rule) string {
...
 
 func updateRule(ruleId, ruleJson string) error {
+	rs, ok := registry.Load(ruleId)
+	if !ok {
+		return fmt.Errorf("Rule %s registry not found, try to delete it and recreate", ruleId)
+	}
+	originState := rs.Rule.Triggered
 	// Validate the rule json
-	r, err := ruleProcessor.GetRuleByJson(ruleId, ruleJson)
+	r, err := ruleProcessor.GetRuleByJson(ruleId, ruleJson, processor.WithOriginTriggerState(originState))
```
ngjaying, line 163: I think the only problem here is that the default triggered should be true. There is no need to know the origin state, we default it to true unless user explicitly set it to false.

ngjaying (changes requested):

ngjaying (approved):

PR #2266, opened 2023-09-18
chore: linear cut conf and store dependency

Yisaer:


internal/conf/logger/logger.go
```diff
@@ -0,0 +1,60 @@
...
+	filename "github.com/keepeye/logrus-filename"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	Log     *logrus.Logger
+	LogFile *os.File
+)
+
+func init() {
```
ngjaying, line 30: Why adding init as InitLogger already be invoked explicitly.

internal/conf/time.go
```diff
@@ -18,11 +18,19 @@ import (
...
 
 	"github.com/benbjohnson/clock"
+
+	"github.com/lf-edge/ekuiper/internal/conf/logger"
 )
 
 var Clock clock.Clock
 
 func InitClock() {
+	if logger.Log == nil {
```
ngjaying, line 28: Can we say conf init() already guarantee logger exists? So this is unnecessary.

ngjaying (changes requested):

internal/conf/logger/logger.go
Yisaer: In some unit test, `InitLogger` won't be called by `conf.init` as expected.

ngjaying: It looks like current implementation will init logger twice. Can we just remove InitLogger in conf?

Yisaer: updated.

ngjaying (approved):

PR #2267, opened 2023-09-18
feat: add android packaging

bing-zhub:
android_ekuiper

Many users want to deploy ekuiper on Android, in order to facilitate this part of the user, this PR compiles the ekuiper cross-platform to Android, and packages it to produce apk files.

Implementation:
Based on the cross-platform GUI toolkit [fyne](https://developer.fyne.io/started/mobile), using absolute path to read/write configuration files/logs, all configuration files and runtime files will be saved to `/storage/emulated/0/ekuiper/` directory, if you need to modify the configuration information, please use the file manager to modify the corresponding file and then restart ekuiper.

Problems:
Can't get file storage permission automatically, need to enable it manually.


ngjaying:
Thanks! Please follow the link in the DCO Details to pass the DCO. Also, could you please use English in the as the default UI language.

bing-zhub:
Okay, I'll pass the DCO, and then switch the default UI language to English. I'll close this PR for now, I'll start a new one later.

PR #2272, opened 2023-09-21
docs: add doc for the configuration

Yisaer:


docs/en_US/configuration/global_configurations.md
```diff
@@ -20,6 +20,14 @@ basic:
   maxAge: 72
   # Whether to ignore case in SQL processing. Note that, the name of customized function by plugins are case-sensitive.
   ignoreCase: false
```
ngjaying, line 22: I think we can add more descriptions. The cfgStorageType should be file and kv only. If kv is set, the real kv type is defined in the store section, which could be sqlite/redis and fdb in the future. Do not need to mention fdb for now.

ngjaying (changes requested):

ngjaying (approved):

PR #2276, opened 2023-09-25
feat: support kafka source extension

Yisaer:


ngjaying (commented):
Suggest to put Kafka source/sink in one package to share the Kafka connection establishment code.

extensions/sources/kafka/ext/kafka.go
```diff
@@ -0,0 +1,161 @@
...
+			return
+		default:
+		}
+		msg, err := s.reader.ReadMessage(ctx)
+		if err != nil {
+			logger.Errorf("Recv kafka error %v", err)
+			errCh <- err
+			return
+		}
+		data := make(map[string]interface{})
```
ngjaying, line 148: Use ctx.Decode to support all formats instead of just json

ngjaying (changes requested):

ngjaying (approved):

PR #2280, opened 2023-09-27
docs: add doc for the redisSub and redisPub

retoool:


docs/en_US/guide/sources/builtin/redisSub.md
```diff
@@ -0,0 +1,64 @@
+## RedisSub 数据源
```
ngjaying, line 1: RedisSub Source

ngjaying (changes requested):
Thanks you! Just find a small problem to fix.

ngjaying (approved):
Thank you @retoool

PR #2282, opened 2023-09-27
feat: support kafka tls sink

Yisaer:


internal/pkg/cert/cert.go
```diff
@@ -9,10 +9,10 @@ import (
 )
 
 type TlsConfigurationOptions struct {
-	SkipCertVerify bool
-	CertFile       string
-	KeyFile        string
-	CaFile         string
+	SkipCertVerify bool   `json:"skipCertVerify"`
```
ngjaying, line 12: Suggest to unify the namings of these properties to MQTT/REST

ngjaying (changes requested):

Issue #186, opened 2020-03-25
Add JSON output option to `cli` command


dcmartin:
Provide option for `cli` command to output JSON rather than human-readable-text (HRT).  For example, currently the output for `docker exec -it kuiper bin/cli describe stream motion` where `motion` is one defined stream, results in:

```
Fields
--------------------------------------------------------------------------------
device	string
camera	string
count	bigint

DATASOURCE: +/+/+/event/end/+
FORMAT: JSON
```

A structured result in JSON would be much more efficient for automating interactions with the CLI in scripts, as well as to provide a simple ReSTful API (which I am about to do).

ngjaying:
> Provide option for `cli` command to output JSON rather than human-readable-text (HRT). For example, currently the output for `docker exec -it kuiper bin/cli describe stream motion` where `motion` is one defined stream, results in:
> 
> ```
> Fields
> --------------------------------------------------------------------------------
> device	string
> camera	string
> count	bigint
> 
> DATASOURCE: +/+/+/event/end/+
> FORMAT: JSON
> ```
> 
> A structured result in JSON would be much more efficient for automating interactions with the CLI in scripts, as well as to provide a simple ReSTful API (which I am about to do).

There is a RESTful API in place, check https://github.com/emqx/kuiper/tree/master/docs/en_US/restapi

dcmartin:
Excellent!  +1 on the REST API.  I will make use of it.  Nonetheless, I think it would be great to have the CLI also provide the option to produce JSON output.

jinfahua:
@dcmartin Are you interested in developing feature ``CLI also provide the option to produce JSON output`` ?  :-)

PR #2286, opened 2023-10-05
feat: add -json flag to describe

jerempy:
Add ability to get output from describe as json format.

Closes: [#186](https://github.com/lf-edge/ekuiper/issues/186)

I added a converter to parse the regular describe return val. I did this so as not to pass a flag all the way down, also not to potentially impact anything else in other places uses the `execDescribe` - since that method outputs a predictable format I just added a parses to convert it to json.

examples:

* No Flag
`go run main.go describe stream topic`
Fields
--------------------------------------------------------------------------------
USERID  bigint
FIRST_NAME      string
LAST_NAME       string
NICKNAMES       array(string)
Gender  boolean
ADDRESS struct(STREET_NAME string, NUMBER bigint)

DATASOURCE: users
FORMAT: JSON
KEY: USERID

* With flag
`go run main.go describe stream topic -json`
{
        "Fields": [
                {
                        "Name": "USERID",
                        "Type": "bigint"
                },
                {
                        "Name": "FIRST_NAME",
                        "Type": "string"
                },
                {
                        "Name": "LAST_NAME",
                        "Type": "string"
                },
                {
                        "Name": "NICKNAMES",
                        "Type": "array(string)"
                },
                {
                        "Name": "Gender",
                        "Type": "boolean"
                },
                {
                        "Name": "ADDRESS",
                        "Type": "struct(STREET_NAME string, NUMBER bigint)"
                }
        ],
        "Options": {
                "DATASOURCE:": "users",
                "FORMAT:": "JSON",
                "KEY:": "USERID"
        }
}

### DIFF
--- a/internal/processor/stream_test.go
+++ b/internal/processor/stream_test.go
@@ -380,3 +380,55 @@ func TestInferredStream(t *testing.T) {
 		}
 	}
 }
+
+func TestDescribeToJson(t *testing.T) {
+	tests := []struct {
+		s string
+		r string
+	}{
+		{
+			s: "Fields\n--------------------------------------------------------------------------------\n" +
+				"USERID\tbigint\nFIRST_NAME\tstring\nLAST_NAME\tstring\nNICKNAMES\tarray(string)\nGender\tboolean\nADDRESS\t" +
+				"struct(STREET_NAME string, NUMBER bigint)\n\nDATASOURCE: users\nFORMAT: JSON\nKEY: USERID\n",
+			r: `{
+	"Fields": [
+		{
+			"Name": "USERID",
+			"Type": "bigint"
+		},
+		{
+			"Name": "FIRST_NAME",
+			"Type": "string"
+		},
+		{
+			"Name": "LAST_NAME",
+			"Type": "string"
+		},
+		{
+			"Name": "NICKNAMES",
+			"Type": "array(string)"
+		},
+		{
+			"Name": "Gender",
+			"Type": "boolean"
+		},
+		{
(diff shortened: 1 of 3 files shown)


ngjaying:
> Any insight on why the test is failing? 
> looks like in the redis pubsub module but not sure why other than maybe test timing out?

Looks like your change has nothing to do with the failure, I just rerun the test.

internal/processor/stream_test.go
```diff
@@ -380,3 +380,46 @@ func TestInferredStream(t *testing.T) {
 		}
 	}
 }
+
+func TestDescribeToJson(t *testing.T) {
+	input := "Fields\n--------------------------------------------------------------------------------\nUSERID\tbigint\nFIRST_NAME\tstring\nLAST_NAME\tstring\nNICKNAMES\t" +
```
Rui-Gan, line 385: It would be helpful if you could modify it to align with the format of other tests in the project, making it easier to add new tests.

Rui-Gan (changes requested):

internal/processor/stream_test.go
Rui-Gan: cool, what I meant was that you can enhance the code by incorporating the following structure, which allows to conveniently add test samples in the future:
``` go
tests := []struct {
		input   string
		want   string
		err string
	}{}
```

This approach simplifies the process of adding new test cases at a later stage.

Rui-Gan (approved):

ngjaying:
Will merge this one. Please add another PR to update the doc. Thanks.

Issue #1765, opened 2023-04-13
support create plugin from local file


carlclone:
**What would you like to be added/modified**:

now:
```
influxPlugin.txt:
{
   "file":"http://localhost:8081/influx.zip"
}

./kuiper create plugin sink kafka -f influxPlugin.txt

```

expected:

```
./kuiper create plugin sink influx -f ./influx.zip
```

**Why is this needed**:

ngjaying:
Looks graceful. But the plugin creation parameters is more than just the zip file:

```
type IOPlugin struct {
	Name       string   `json:"name"`
	File       string   `json:"file"`
	ShellParas []string `json:"shellParas"`
}
```

Maybe use another option name (-f already been taken) for a shortcut to create zip only plugin?

hantmac:
How about using it like this:

```
influxPlugin.txt:
{
   "file":"local/path/influx.zip"
}

./kuiper create plugin sink kafka -f influxPlugin.txt

```

PR #2288, opened 2023-10-06
feat: create plugin using local file

jerempy:
Adds ability to create plugins using something like `./kuiper create plugin sink influx -zf ./influx.zip`

Closes: https://github.com/lf-edge/ekuiper/issues/1765

The approach I took was to add an additional flag "-zf" or "zipfile".
if passed it will take the path to the file and wrap it in the expected json and then run the function normally if it was passed as a json file.

### DIFF
--- a/cmd/kuiper/main.go
+++ b/cmd/kuiper/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/rpc"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


ngjaying:
> unfortunately i dont understand the stack trace on the failing test action - any insight?

No problem, we'll take a look.

Rui-Gan:
> unfortunately i dont understand the stack trace on the failing test action - any insight?

In the `fileExists` function, true is returned if the file does not exist. In the `readDef` function, if the `fileExists` function returns false, that is, if the file exists, it returns error; In short, this `fileExists` function should return `!os.IsNotExist(err)`.

Rui-Gan:
The current implementation does not appear to allow the use of local files for creating plugins. 

To create a plugin, you cannot directly use local files with ekuiper. Instead, you need to host the local file on an HTTP server and use the URL of the plugin file to create the plugin.

Hence, if you attempt to reuse the previous code implementation for plugin creation, it will not currently support the use of local files.

The statement above reflects my understanding.  If there are any inaccuracies or mistakes, please don't hesitate to correct them.

ngjaying:
> The current implementation does not appear to allow the use of local files for creating plugins.
> 
> To create a plugin, you cannot directly use local files with ekuiper. Instead, you need to host the local file on an HTTP server and use the URL of the plugin file to create the plugin.
> 
> Hence, if you attempt to reuse the previous code implementation for plugin creation, it will not currently support the use of local files.
> 
> The statement above reflects my understanding. If there are any inaccuracies or mistakes, please don't hesitate to correct them.

If CLI is run in the same machine of kuiperd, it is possible to load the local file. Besides http://xxx, the file also allows file:/// protocol. So when calling local kuiperd, we may just convert local file to file:/// format to meet the current API.

Rui-Gan (approved):
Great, please add another PR to update the doc. Thanks.

ngjaying:
@jerempy Thanks very much!

PR #2290, opened 2023-10-07
feat: revise kafka sink

Yisaer:


ngjaying (changes requested):
The TLS config should be consistent to MQTT and REST sink.

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/keepeye/logrus-filename v0.0.0-20190711075016-ce01a4391dd1
 	github.com/klauspost/compress v1.16.4
 	github.com/lestrrat-go/file-rotatelogs v2.4.0+incompatible
+	github.com/lf-edge/ekuiper/extensions v0.0.0-20231009091421-d4e87939603d
```
ngjaying, line 26: Why need this? We should not depend on extensions to avoid circular dependency

internal/server/rest.go
```diff
@@ -142,6 +142,7 @@ func createRestServer(ip string, port int, needToken bool) *http.Server {
 
 	r := mux.NewRouter()
 	r.HandleFunc("/", rootHandler).Methods(http.MethodGet, http.MethodPost)
+	r.HandleFunc("/system/status", getMetricsHandler).Methods(http.MethodGet)
```
ngjaying, line 145: This file looks like irrelevant to the PR

ngjaying (changes requested):

ngjaying (approved):

PR #2291, opened 2023-10-07
feat: support metrics handler

Yisaer:


internal/server/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type Metrics struct {
+	kp *process.Process
+}
+
+func NewMetrics() *Metrics {
+	kProcess, err := process.NewProcess(int32(os.Getpid()))
+	if err != nil {
+		panic(fmt.Sprintf("Can not initialize process for ekuiperd : %v", err))
```
ngjaying, line 31: Shall we panic here?

internal/server/rest.go
```diff
@@ -141,6 +141,7 @@ func createRestServer(ip string, port int, needToken bool) *http.Server {
 
 	r := mux.NewRouter()
 	r.HandleFunc("/", rootHandler).Methods(http.MethodGet, http.MethodPost)
+	r.HandleFunc("/system/status", getMetricsHandler).Methods(http.MethodGet)
```
ngjaying, line 144: How about merge to rootHandler? Root handler also return elapse time which is also a status, so I prefer to not create a new API.

internal/server/metrics.go
```diff
@@ -0,0 +1,49 @@
...
+func NewMetrics() *Metrics {
+	kProcess, err := process.NewProcess(int32(os.Getpid()))
+	if err != nil {
+		conf.Log.Warnf("Can not initialize process for ekuiperd : %v", err)
+	}
+	return &Metrics{kp: kProcess}
+}
+
+func (m *Metrics) GetCpuUsage() string {
+	percent, _ := m.kp.CPUPercent()
```
ngjaying, line 39: Need nil check and fallback as m.kp could be nil.

ngjaying (changes requested):

ngjaying (approved):

PR #2319, opened 2023-10-15
feat: Concealing Passwords in Config Key Retrieval

Rui-Gan:


internal/meta/yamlConfigMeta.go
```diff
@@ -153,13 +154,26 @@ func GetYamlConf(configOperatorKey, language string) (b []byte, err error) {
...
 	if b, err = json.Marshal(cf); nil != err {
 		return nil, fmt.Errorf(`%s%v`, getMsg(language, source, "json_marshal_fail"), cf)
 	} else {
 		return b, err
 	}
 }
 
+func hiddenPassword(configkeys map[string]map[string]interface{}) map[string]map[string]interface{} {
+	for key, kvs := range configkeys {
+		for k := range kvs {
```
ngjaying, line 167: Need to check embed password as well. Another thing to check is the SQL connection url.

ngjaying (changes requested):

ngjaying (approved):

PR #2329, opened 2023-10-17
fix(build): android pkg name

ngjaying:


.github/workflows/build_packages.yaml
```diff
@@ -123,9 +123,9 @@ jobs:
         run: |
           ver=$(git describe --tags --always)
           cd app
-          fyne package -os android/arm64 -name ekuiper -metadata version=$ver -release -appID github.com.lfedge.ekuiper -icon icon.png || true
+          fyne package -os android/arm64 -name kuiper-$ver -metadata version=$ver -release -appID github.com.lfedge.ekuiper -icon icon.png || true        
           mkdir -p ../_packages
-          mv ekuiper.apk ../_packages/
+          mv kuiper* ../_packages/
```
Rory-Z, line 128: We also need change 404 line

Rory-Z: I think we need to remind people in some way that the package name generated by fyne package is not `kuiper-$ver`, in order to reduce future maintenance costs.

ngjaying: How? Any suggestions?

Rory-Z: No good ideas, maybe we can add some comments.

Rory-Z (approved):

PR #2338, opened 2023-10-19
chore: make kafka client error as io error

Yisaer:
https://github.com/segmentio/kafka-go/blob/f2d9e087266352d276b3cb951b4c668c114b8b75/produce.go#L164-L170

### DIFF
--- a/extensions/sinks/kafka/ext/kafka.go
+++ b/extensions/sinks/kafka/ext/kafka.go
@@ -238,12 +238,15 @@ func (m *kafkaSink) Collect(ctx api.StreamContext, item interface{}) error {
 	case kafkago.WriteErrors:
 		count := 0
 		for i := range messages {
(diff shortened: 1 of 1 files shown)


extensions/sinks/kafka/ext/kafka.go
```diff
@@ -243,7 +243,7 @@ func (m *kafkaSink) Collect(ctx api.StreamContext, item interface{}) error {
 				continue
 
 			case kafkago.Error:
-				if err.Temporary() {
+				if err.Temporary() || strings.Contains(err.Error(), "kafka.(*Client).Produce:") {
```
ngjaying, line 246: I don't think this works, and this is irrelevant to the host not found problem. It is not a kafkago.Error actually.

ngjaying (changes requested):

ngjaying (approved):

PR #2393, opened 2023-11-06
fix: fix several bugs about logs

Yisaer:
1. fix doc spell error
2. support validate mqtt sink topic
3. remove useless error log
4. revise buffer logs


ngjaying (commented):
Can we split the log level as a standalone PR? Because it is a feature.

internal/io/mqtt/mqtt_sink.go
```diff
@@ -60,6 +61,9 @@ func (ms *MQTTSink) Configure(ps map[string]interface{}) error {
 	if adconf.Tpc == "" {
 		return fmt.Errorf("mqtt sink is missing property topic")
 	}
+	if strings.Contains(adconf.Tpc, "#") || strings.Contains(adconf.Tpc, "+") {
```
ngjaying, line 64: This can be covered by ut

ngjaying (changes requested):

ngjaying:
Commit title and message should be more specific. Please list all changes in the message body

ngjaying (approved):

PR #2394, opened 2023-11-06
feat: support log level

Yisaer:


etc/kuiper.yaml
```diff
@@ -1,4 +1,6 @@
 basic:
+  # debug | info | warn | error | fatal | panic
```
ngjaying, line 2: Need to document the precedence of logLevel and debug option

ngjaying (approved):

PR #2397, opened 2023-11-08
feat: support validate rule response

Yisaer:


internal/server/rule_manager.go
```diff
@@ -351,11 +353,22 @@ func getRuleTopo(name string) (string, error) {
 	}
 }
 
-func validateRule(name, ruleJson string) (bool, error) {
+func validateRule(name, ruleJson string) ([]string, bool, error) {
 	// Validate the rule json
-	_, err := ruleProcessor.GetRuleByJson(name, ruleJson)
+	rule, err := ruleProcessor.GetRuleByJson(name, ruleJson)
```
ngjaying, line 358: Will it validate if streams/tables are existed?

Yisaer: updated.

ngjaying (approved):

PR #2401, opened 2023-11-09
feat: support kafka source rewindable

Yisaer:


docs/en_US/guide/sources/plugin/kafka.md
```diff
@@ -22,7 +22,6 @@ default:
   groupID: ""
   partition: 0
   maxBytes: 1000000
-  offset: 0
```
ngjaying, line 25: we can keep the offset configuration which represents the offset of first run. If qos is set, the offset state will override it.

ngjaying (changes requested):

ngjaying (approved):

PR #2428, opened 2023-11-21
feat: support sql source reconnect

Yisaer:


extensions/sources/sql/sql.json
```diff
@@ -60,6 +60,36 @@
           "zh_CN": "间隔时间"
         }
       },
+      {
+        "name": "maxRetry",
```
ngjaying, line 64: These two props have the potential to be common properties. Try to make sure the names are the same across all sources.

extensions/sources/sql/ext/sql.go
```diff
@@ -89,6 +96,16 @@ func (m *sqlsource) Open(ctx api.StreamContext, consumer chan<- api.SourceTuple,
 			logger.Debugf("Query the database with %s", query)
 			rows, err := m.db.Query(query)
 			if err != nil {
+				logger.Errorf("sql source meet error, err:%v", err)
```
ngjaying, line 99: Can we treat the t.C as the retry interval? Then we will keep retrying.

extensions/sources/sql/ext/sql.go
```diff
@@ -28,8 +30,10 @@ import (
 )
 
 type sqlConConfig struct {
-	Interval int    `json:"interval"`
-	Url      string `json:"url"`
+	Interval             int    `json:"interval"`
+	Url                  string `json:"url"`
+	MaxAttempts          int    `json:"maxAttempts"`
+	MaxReconnectInterval int    `json:"maxReconnectInterval"`
```
ngjaying, line 36: The SQL source is kind of pull source. It will do another pull after an interval which is kind of a retry. So I don't think we need another interval here. Just need a maxAttempts in one time

extensions/sources/sql/ext/sql.go
```diff
@@ -89,6 +96,16 @@ func (m *sqlsource) Open(ctx api.StreamContext, consumer chan<- api.SourceTuple,
...
 			rows, err := m.db.Query(query)
 			if err != nil {
+				logger.Errorf("sql source meet error, err:%v", err)
+				canReconnect, err2 := m.Reconnect(err)
+				if canReconnect {
+					if err2 == nil {
+						continue
+					} else {
+						logger.Errorf("Run sql query(%s) error %v", query, err2)
+						errCh <- err2
```
ngjaying, line 106: Do not send out recoverable error to error channel, it will stop the whole rule. Just send out an ErrorSourceTuple

ngjaying (changes requested):

ngjaying (approved):

PR #2437, opened 2023-11-27
fix: fix sqlserver lookup table

Yisaer:


extensions/sources/sql/ext/sqlLookup.go
```diff
@@ -41,6 +43,10 @@ func (s *sqlLookupSource) Open(ctx api.StreamContext) error {
 	if err != nil {
 		return fmt.Errorf("connection to %s Open with error %v", s.url, err)
 	}
+	index := strings.Index(s.url, "://")
```
ngjaying, line 46: You can use util.ParseDriver

extensions/sources/sql/ext/sqlLookup.go
```diff
@@ -62,29 +68,7 @@ func (s *sqlLookupSource) Configure(datasource string, props map[string]interfac
...
-			query += " AND "
-		}
-		switch v := values[i].(type) {
-		case string:
-			query += fmt.Sprintf("`%s` = '%s'", k, v)
-		default:
-			query += fmt.Sprintf("`%s` = %v", k, v)
-		}
-	}
+	query := s.buildQuery(fields, keys, values)
```
ngjaying, line 71: Is it better/possible to use the query builder which already consider different db.

ngjaying (changes requested):

ngjaying (approved):

PR #2439, opened 2023-11-28
fix: support pruning struct column

Yisaer:
Prune embed fields.


internal/topo/planner/planner_test.go
```diff
@@ -77,10 +77,16 @@ func Test_createLogicalPlan(t *testing.T) {
 					value STRING,
 					hum BIGINT
 				) WITH (TYPE="file");`,
+		"src3": `CREATE STREAM src3 (
+					a struct(b struct(c bigint,d bigint),e bigint),
```
ngjaying, line 81: Have we try embedded struct and mix of array struct

pkg/ast/sourceStmt.go
```diff
@@ -54,6 +54,8 @@ type JsonStreamField struct {
 	Type       string                      `json:"type"`
 	Items      *JsonStreamField            `json:"items,omitempty"`
 	Properties map[string]*JsonStreamField `json:"properties,omitempty"`
+
+	Selected bool
```
ngjaying, line 58: This looks like a state. Is this used as a singleton? Otherwise, the shared state may be affected by each other.

ngjaying (approved):

PR #2454, opened 2023-12-05
fix: fix memory leak when stop rule

Yisaer:
1. fix goroutine leak when stop rule in `runTopo`
2. de-reference old Topo in order to release memory after stop/close rule
3. close channel to release memory after `removePub`


internal/io/memory/pubsub/manager.go
```diff
@@ -105,6 +105,9 @@ func RemovePub(topic string) {
 	if sinkConsumerChannels, exists := pubTopics[topic]; exists {
 		sinkConsumerChannels.count -= 1
 		if len(sinkConsumerChannels.consumers) == 0 && sinkConsumerChannels.count == 0 {
+			for _, ch := range sinkConsumerChannels.consumers {
```
ngjaying, line 108: This may not needed. Channel can be gc without closing

ngjaying (approved):

PR #2458, opened 2023-12-06
feat: revise sql error log

Yisaer:


ngjaying (commented):
How about just return in the error message

ngjaying (approved):

PR #2478, opened 2023-12-12
docs(android): add notice for baseDir

ngjaying:


Yisaer (approved):

ngjaying:
Fix the issue by another PR, so the doc is not needed now.

PR #2490, opened 2023-12-18
fix: fix websocket  server publish data

Yisaer:


ngjaying (commented):
Is it possible to add ut for the server mode?

ngjaying (approved):

PR #2494, opened 2023-12-18
fix: fix lookup plan column pruning

Yisaer:


internal/topo/planner/dataSourcePlan.go
```diff
@@ -210,7 +210,7 @@ func (p *DataSourcePlan) PruneColumns(fields []ast.Expr) error {
 				}
 			}
 		case *ast.FieldRef:
-			if !p.isWildCard && (f.StreamName == ast.DefaultStream || f.StreamName == p.name) {
+			if f.StreamName == ast.DefaultStream || f.StreamName == p.name {
```
ngjaying, line 213: Do we need to differentiate between schema-less and schemaful? If schema is defined, wildcard can skip this check?

ngjaying (approved):

PR #2502, opened 2023-12-21
feat: support merge schema

Yisaer:


internal/topo/node/source_pool.go
```diff
@@ -296,6 +314,20 @@ func (ss *sourceSingleton) attach(instanceKey string, bl int) error {
 	return err
 }
 
+func (ss *sourceSingleton) detachSchema(schemaKey string) {
+	ss.Lock()
+	defer ss.Unlock()
+	decodeConverter := ss.ctx.Value(kctx.DecodeKey)
+	if decodeConverter != nil {
+		fastDecoder, ok := decodeConverter.(*json.FastJsonConverter)
```
ngjaying, line 322: Should not refer to specific converter. Add AttachSchema and DetachSchema to the converter interface?

ngjaying (changes requested):

internal/topo/node/source_node.go
```diff
@@ -111,6 +111,7 @@ func (m *SourceNode) Open(ctx api.StreamContext, errCh chan<- error) {
 			props["delimiter"] = m.options.DELIMITER
 			m.options.Schema = nil
 			if m.schema != nil {
+				m.options.SchemaKey = ctx.GetRuleId()
```
ngjaying, line 114: The name of schemaKey is misleading. Since the rule id is already stored in context, why do we need to re-define it here?

ngjaying (approved):

Issue #2407, opened 2023-11-09
Incorrect Behavior of eKuiper API for start/stop Rules


EscanorUt:
## Environment
- eKuiper version: 1.11.4
- OS: docker-alpine

## Bug Description
The eKuiper API to stop rules is responding with 200 OK HTTP Status when providing inexistent rule id unlike the start API which returns 400 Bad Request HTTP Status

## Expected Behavior
The expected behavior is for the eKuiper stop rule API to return 400 Bad Request HTTP Status when the id provided does not exist.

## Steps to Reproduce
1. Try to stop a rule that does not exist
2. Observe that the API returns a "200 OK" status code instead of the expected "400 Bad Request" status code.

## Additional Information
![image](https://github.com/lf-edge/ekuiper/assets/145465122/f2805012-bd78-4c7f-a5b8-7f86b2a88b85)

ngjaying:
OK, it is reasonable. Currently stop does not check existence. It returns 200 only means the rule is NOT started.

PR #2505, opened 2023-12-23
fix(rest): return 404 when rule not found

carlclone:
#2407

### DIFF
--- a/internal/server/rule_manager.go
+++ b/internal/server/rule_manager.go
@@ -233,7 +233,7 @@ func startRuleInternal(name string) error {
 func reRunRule(name string, isInternal bool) error {
 	rs, ok := registry.Load(name)
 	if !ok {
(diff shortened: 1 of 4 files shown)


internal/server/rule_manager.go
```diff
@@ -256,8 +256,7 @@ func stopRuleInternal(name string) {
 	}
 }
 
-func stopRule(name string) (result string) {
-	var err error
+func stopRule(name string) (err error, result string) {
```
Yisaer, line 259: error should be the last return valur

carlclone: can you take a look at the failed test case , not understand why

ngjaying: looks like a random error. I just rerun it.

Yisaer (approved):

Issue #2094, opened 2023-07-14
Complete SQL functions for math and time


xjasonlyu:
**What would you like to be added/modified**:

Add and complete similar MySQL/Postgres-style math and time functions in eKuiper:

- [x] https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html
- [x] https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html

Current missing mathematical funcitons
- [x] conv
- [x] cot
- [x] crc32
- [x] degrees
- [x] format
- [x] radians

**Why is this needed**:

Easy for time and math manipulation when using SQL functions.

yai-dev:
@xjasonlyu Hi there, i will help to time functions for eKuiper and i'm working on it. (this comment is for mark purpose only :))

ngjaying:
Still lack some math functions.

AyushK-101:
@ngjaying I would like to work on this, could you tell me about some of the math functions that are yet to be added.

ngjaying:
> @ngjaying I would like to work on this, could you tell me about some of the math functions that are yet to be added.

@AyushK-101 Thank you. I updated the issue to list all missing math functions. Please check.

ngjaying:
> @ngjaying can i please work on this issue?

Sure, please comment about what function are you working on once started

SmartPotter:
@ngjaying, I want to work on the degrees mathematical function, May I know what exactly you are looking for?

ShikhaaT44:
Can i work on any functions which are yet unclaimed to work upon? I'm interested on working on degrees function

PR #2506, opened 2023-12-25
feat: implement CONV function

carlclone:
#2094

### DIFF
--- a/docs/en_US/sqls/functions/mathematical_functions.md
+++ b/docs/en_US/sqls/functions/mathematical_functions.md
@@ -255,3 +255,22 @@ degrees(col)
 ```
 
 converted from radians to degrees
+
+## CONV
+
+```text
+conv(N,from_base,to_base)
+```
+
+converts numbers between different number bases. Returns a string representation of the number N, converted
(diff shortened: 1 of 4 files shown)


Yisaer (approved):

docs/en_US/sqls/functions/mathematical_functions.md
```diff
@@ -255,3 +255,12 @@ degrees(col)
...
 
 converted from radians to degrees
+
+## CONV
+
+```text
+conv(col,from_base,to_base)
+```
+
+converts numbers between different number bases. The argument col is interpreted as an integer, but may be specified as
```
ngjaying, line 265: The return result is a string? Need to clarify that.

ngjaying (changes requested):

docs/en_US/sqls/functions/mathematical_functions.md
carlclone: done

ngjaying:
@carlclone Due to dependency updates, the current python plugin tests fail. We have fixed that in another PR. Please rebase to master to pass the tests. Thanks.

ngjaying (approved):

PR #2509, opened 2023-12-26
feat: revise cmdline config

Yisaer:


ngjaying (commented):
The absolute mode is used by build ci for deb and rpm. Need to fix that if removing this mode.

PR #2522, opened 2024-01-03
feat(func): add locale argument support

carlclone:


docs/en_US/sqls/functions/locale_support.md
```diff
@@ -0,0 +1,115 @@
+# Locale Support List
```
ngjaying, line 1: Is it OK to just add this part into string_functions.md?

ngjaying (approved):

PR #2545, opened 2024-01-09
feat: support runtime schema in rule explain

Yisaer:


ngjaying (approved):
Approve for now because feature is working. I prefer to extract schema calculation from specific converter like json. Maybe we can refactor later.

PR #2550, opened 2024-01-10
feat: support raw tls config

Yisaer:
1. support raw tls config (use tls file if both raw config and file defined.)
2. union the tls configure for all source/sink


docs/en_US/guide/sinks/builtin/mqtt.md
```diff
@@ -14,6 +14,9 @@ The action is used for publish output message into an MQTT server.
...
 | privateKeyPath       | true     | The private key path. It can be either absolute path, or relative path, which is similar to use of certificationPath.                                                                                                                                                                                                                                     |
 | rootCaPath           | true     | The location of root ca path. It can be an absolute path, or a relative path, which is similar to use of certificationPath.                                                                                                                                                                                                                               |
+| rawCert            | true | base64 encoded original text of cert |
```
ngjaying, line 17: Two suggestions:
1. Name the property to be a counterpart of the original property. certificationRaw vs certificationPath; privateKeyRaw vs privateKeyPath, rootCaRaw vs. rootCaPath. This can help people easily find out the relationship between them.
2. Demonstrate the priority of each pair, xxPath takes precedence over xxRaw.

internal/pkg/cert/cert.go
```diff
@@ -15,21 +15,110 @@
...
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
 	"os"
 
 	"github.com/lf-edge/ekuiper/internal/conf"
 	"github.com/lf-edge/ekuiper/pkg/cast"
 )
 
+func GenTLSConfig(props map[string]interface{}) (*tls.Config, *TlsConfigurationOptions, error) {
```
ngjaying, line 28: Looks like the second return value *TlsConfigurationOptions is useless outside. Can we remove it?

ngjaying (changes requested):

docs/en_US/guide/sinks/builtin/mqtt.md
```diff
@@ -14,9 +14,9 @@ The action is used for publish output message into an MQTT server.
...
 | privateKeyPath       | true     | The private key path. It can be either absolute path, or relative path, which is similar to use of certificationPath.                                                                                                                                                                                                                                     |
 | rootCaPath           | true     | The location of root ca path. It can be an absolute path, or a relative path, which is similar to use of certificationPath.                                                                                                                                                                                                                               |
-| rawCert            | true | base64 encoded original text of cert |
-| rawKey             | true | base64 encoded original text of key |
-| rawCA              | true | base64 encoded original text of CA |
+| rawCertification            | true | base64 encoded original text of cert, use `certificationPath` first if both defined |
```
ngjaying, line 17: Is it better to align with the existing naming pattern? By this way, when ordering by alphabet, the two property can be side by side: certificationRaw and certiricationPath.

internal/io/http/client.go
```diff
@@ -172,16 +172,11 @@ func (cc *ClientConf) InitConf(device string, props map[string]interface{}, with
 			return fmt.Errorf("headers must be a map or a string")
 		}
 	}
-	tlsOpts, err := cert.GenTlsConfigurationOptions(props)
+	tlscfg, tlsOpts, err := cert.GenTLSConfig(props)
 	if err != nil {
 		return err
 	}
 	c.TlsConfigurationOptions = tlsOpts
```
ngjaying, line 179: Looks like the c.TlsConfigurationOptions is useless. Can we just remove it and let the GenTLSConfig only return two values.

ngjaying (approved):

Issue #2530, opened 2024-01-04
More array and object functions


ngjaying:
- [x] object_pick(key...): Returns a new object containing some of the key-value pairs from an existing object. Example, object `{"a":1,  "b":2, "c":3}`, object_pick('a','c') return `{"a":1,  "c":3}`
- [ ] kvpair_array_to_obj(array, [keyFieldName], [valueFieldName]): Returns a new converting from a kv pair array, such as `[{"key":"key1", "value":1},{"key":"key2", "value":2}]`, to `{"key1":1, "key2":2}`.
- [ ] obj_to_kvpair_array(obj, [keyFieldName], [valueFieldName]): A revert function of kvpair_array_to_obj

rextang:
Hi, I'm interested in this issue. Can I implement these functions?

ngjaying:
> Hi, I'm interested in this issue. Can I implement these functions?

Sure, warm welcome

PR #2560, opened 2024-01-17
feat: support object_pick function

rextang:
The function signature of object_pick, which was mentioned in #2530, has changed slightly to keep consistent with the erase function.

### DIFF
--- a/docs/en_US/sqls/functions/object_functions.md
+++ b/docs/en_US/sqls/functions/object_functions.md
@@ -162,3 +162,21 @@ result:
 ```sql
 {"baz": [1, 2, 3], "bar": 'hello world'}
 ```
+
+## OBJECT_PICK
+
+```text
+object_pick(obj, k)
+```
+
+If k is a string, return a new object where the key k is preserved. If k is an array of strings, return a new object where the keys in k are preserved.
+
+```sql
+erase({"baz": [1, 2, 3], "bar": 'hello world',"foo":'emq'}, 'foo')
+```
+
+result:
+
+```sql
+{"baz": [1, 2, 3], "bar": 'hello world'}
+```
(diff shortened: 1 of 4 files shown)


docs/en_US/sqls/functions/object_functions.md
```diff
@@ -162,3 +162,21 @@ result:
...
+## OBJECT_PICK
+
+```text
+object_pick(obj, k)
+```
+
+If k is a string, return a new object where the key k is preserved. If k is an array of strings, return a new object where the keys in k are preserved.
+
+```sql
+erase({"baz": [1, 2, 3], "bar": 'hello world',"foo":'emq'}, 'foo')
```
ngjaying, line 175: The example uses erase, not pick

internal/binder/function/funcs_obj.go
```diff
@@ -240,4 +240,56 @@ func registerObjectFunc() {
 		},
 		check: returnNilIfHasAnyNil,
 	}
+	builtins["object_pick"] = builtinFunc{
+		fType: ast.FuncTypeScalar,
+		exec: func(ctx api.FunctionContext, args []interface{}) (interface{}, bool) {
+			contains := func(array []string, target string) bool {
```
ngjaying, line 246: Suggest to move this out as a common function instead of define it each time.

ngjaying (changes requested):

docs/en_US/sqls/functions/object_functions.md
rextang: Updated.

internal/binder/function/funcs_obj.go
rextang: Yes, moved it out in the same file. Since there is no more usage in other places.

PR #2561, opened 2024-01-17
feat: support message processed metrics

Yisaer:


internal/topo/node/sink_node.go
```diff
@@ -530,6 +530,9 @@ func sendDataToSink(ctx api.StreamContext, sink api.Sink, outData interface{}, s
 	} else {
 		ctx.GetLogger().Debugf("success")
 		stats.IncTotalRecordsOut()
+		if outs, ok := outData.([]map[string]interface{}); ok {
```
ngjaying, line 533: If not OK, shall we add 1?

internal/topo/node/sink_node.go
```diff
@@ -549,6 +552,9 @@ func resendDataToSink(ctx api.StreamContext, sink api.Sink, outData interface{},
 	} else {
 		ctx.GetLogger().Debugf("success resend")
 		stats.IncTotalRecordsOut()
+		if outs, ok := outData.([]map[string]interface{}); ok {
```
ngjaying, line 555: The same as above

internal/topo/node/operations.go
```diff
@@ -153,12 +153,14 @@ func (o *UnaryOperator) doOp(ctx api.StreamContext, errCh chan<- error) {
 				stats.ProcessTimeEnd()
 				for _, v := range val {
 					_ = o.Broadcast(v)
+					stats.IncTotalMessagesProcessed(1)
```
ngjaying, line 156: Need to add 1 in case nil, which is a sign for filter

internal/topo/node/metric/stats_manager.go
```diff
@@ -24,6 +24,7 @@ import (
 const (
 	RecordsInTotal       = "records_in_total"
 	RecordsOutTotal      = "records_out_total"
+	MessagesOutTotal     = "messages_out_total"
```
ngjaying, line 27: Why not align the name as other places? message_processed_total

ngjaying (changes requested):

ngjaying (approved):

PR #2566, opened 2024-01-18
fix: delete default username of redisPubSub

### LINKED ISSUE #2565: Default username in RedisPubSub config causes connection issues with Redis versions 5 or earlier
In the RedisPubSub, the JSON and YAML config files include a default username, which causes connection issues when connecting to Redis versions 5 or earlier.

retoool:
Closes #2565


ngjaying (approved):
Great work, thank you!

PR #2569, opened 2024-01-19
refactor: remove node concurrency

ngjaying:
1. concurrency param becomes useless, will add it back later with in-node concurrency
2. metrics now becomes only 1, remove mutex

### DIFF
--- a/internal/server/rpc_test.go
+++ b/internal/server/rpc_test.go
@@ -15,6 +15,7 @@
 package server
 
 import (
+	"fmt"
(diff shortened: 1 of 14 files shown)


Yisaer (approved):

zhang943:
> 1. concurrency param becomes useless, will add it back later with in-node concurrency
> 2. metrics now becomes only 1, remove mutex

@ngjaying Is the `in-node concurrency` supported now?

PR #2591, opened 2024-01-29
feat: support neuron/kafka ping

Yisaer:


ngjaying (changes requested):
Request to add ut. If hard to setup external connection, at least test ping failure case.

ngjaying (approved):

PR #2603, opened 2024-02-01
feat: add security to edgex via openziti

dovholuknf:
EdgexFoundry has elected to adopt OpenZiti for additional security. This is an initial attempt at what integrating github.com/openziti/ziti into ekuiper for EdgexFoundry might look like.   (I tried my best to follow the commit guidelines, but I'm sure I didn't do something quite right, apologies in advance) 

This PR adds functionality to ekuiper. It allows ekuiper to be removed from the underlay network if desired. In this PR the `restPort` is no longer available on the underlay (IP-based) network as it is exclusively accessible via OpenZiti if enabled.

**Considerations/Notes:**
* Logging - OpenZiti uses pfxlog which uses the `logrus.StandardLogger()` and had to be adapted to in order to allow OpenZiti logging in the ekuiper logs for debugging
* VaultSecret - edgex uses vault to deliver vault tokens to services. this vault token can be exchanged for a jwt which is then usable as authentication to the OpenZiti overlay network and other edgex services. A small `VaultSecret` was created to keep the vault token fresh and obtain new jwt's on a regular interval
* http clients have been adapted to allow connecting/sending data over the OpenZiti overlay network if enabled using the environment variables EDGEX_CREDENTIALS, EDGEX_CREDENTIAL_NAME, OPENZITI_CONTROLLER. If this is not desired, I'm happy to rework it how you see fit.
* http server `ListenAndServe/ListenAndServeTLS` are changed to `Serve(listener)/ServeTLS(listener)` respectively. http server is able to be taken off the IP-based underlay network entirely (making it unattackable by conventional IP-based attacks)


ngjaying:
@dovholuknf Thanks. Please follow the Details link in DCO check to sign the commit.

dovholuknf:
> @dovholuknf Thanks. Please follow the Details link in DCO check to sign the commit.

Ahhh shoot. I forgot. Thanks

dovholuknf:
I have fixed the DCO issue. Thanks

internal/io/http/client.go
```diff
@@ -208,8 +218,63 @@ func (cc *ClientConf) InitConf(device string, props map[string]interface{}, with
 		}
 	}
 
-	tr := &http.Transport{
-		TLSClientConfig: tlscfg,
+	var tr *http.Transport
+	edgexCredentialFile := os.Getenv("EDGEX_CREDENTIALS")
```
ngjaying, line 222: Use the env as switch will affect all http connections. We should add a parameter in the props instead.

ngjaying (commented):
When do you expect to merge this? We may help if available.

internal/conf/logger/logger.go
```diff
@@ -54,10 +55,50 @@ func InitLogger() {
 			break
 		}
 	}
+
+	if os.Getenv("OPENZITI_CONTROLLER") != "" {
```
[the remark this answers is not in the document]
dovholuknf: basically the OpenZiti SDK uses a library called pfxlog. pfxlog is using the `logrus.StandardLogger` but ekuiper makes its own logrus logger. pfxlog is mostly just a formatter for logrus, but by using the StandardLogger it means that without something like this block, any log messages that are written end up wrtten twice. Once by the hook I added and again by the StandardLogger (but formatted howver pfxlog wants to format it). For example when running in my dev env I would see this. See how the same message is logged twice and once formatted with ascii escape code coloring

image

Setting the standard logger to Panic level effectively eliminates this duplication. Then the hook is added to allow openziti's logs to show in ekuiper's (the first line)

internal/io/http/client.go
dovholuknf: that is true. OpenZiti __should__ try to use an OpenZiti connection first, and then fall back to default http transport in that case. OpenZiti is only used if the url being sent matches a pattern indicating it should be intercepted. I'm happy to add a param to the props file if that's the best way to do it. I only did it this way, to get this particular feedback. I'll look to rework with the props file.

internal/conf/vault/vault.go
```diff
@@ -0,0 +1,179 @@
+//go:build edgex
+// +build edgex
```
dovholuknf, line 2: oh one more thing here. I gated this behind the edgex build flag. I need to try building without specifying the flag to ensure it still builds. Also looking for feedback on whether or not this is the preferred approach

ngjaying: Actually, the whole openziti thing here depends on EdgeX's setting. I would rather guard them all behind the edgex build flag.

internal/server/server.go
```diff
@@ -189,12 +197,72 @@ func StartUp(Version string) {
 
 	// Start rest service
 	srvRest := createRestServer(conf.Config.Basic.RestIp, conf.Config.Basic.RestPort, conf.Config.Basic.Authentication)
+
+	var ln net.Listener
+	var lerr error
+	edgexCredentialFile := os.Getenv("EDGEX_CREDENTIALS")
```
ngjaying, line 203: When will this env set?

dovholuknf:
sorry i've been away from this for a long long time... i'm nearly through the edgex work... i am hoping to return to this this week or next week.

internal/server/server.go
dovholuknf: i have reworked how this works. please review. I have added four (4) new environment variables all in `internal/edgex/vault.go`:
* EDGEX_SERVICE_NAME
* EDGEX_CREDENTIALS
* EDGEX_VAULT_HOST
* EDGEX_VAULT_PORT

they are expected to be set when ekuiper starts -- likely from docker. i could adapt to a config file if you want/need/prefer. marking this as resolved...

internal/conf/vault/vault.go
dovholuknf: i have put all edgex/openziti related features behind a build tag. please review the mechanism I went with. I think it makes sense... see what you think. marking resolved.

dovholuknf:
This should be ready to re-review now. If you would like to see instructions on testing with edgex - just let me know and i can  provide a set of docker steps here

internal/server/server_edgex.go
```diff
@@ -0,0 +1,33 @@
+//go:build edgex
+
+// Copyright 2022-2024 EMQ Technologies Co., Ltd.
```
dovholuknf, line 3: let me know what i should do with these copyrights, i must have copied/pasted this. should I put my company in here? EdgeX Foundry? OpenZIti? etc. thx

ngjaying: If it is originally composed by you in your working time, just put your company here.

dovholuknf: thank you. updated.

dovholuknf:
the linter failed to run but it looks like it was some kind of error?  can you maybe rerun it?

also is the test case failure due to something i did? I'm surprised to see that based on what i added?
```

--- FAIL: TestExtensions (0.03s)
    mock_topo.go:392: Drop stream fails: ext is not found.
    mock_topo.go:392: Drop stream fails: ext2 is not found.
    plugin_rule_test.go:81: The test bucket size is 2.
```

i'll try to fix the 'max commit len' issue -- 58 chars at the moment? ;/

ngjaying:
> the linter failed to run but it looks like it was some kind of error? can you maybe rerun it?
> 
> also is the test case failure due to something i did? I'm surprised to see that based on what i added?
> 
> ```
> 
> --- FAIL: TestExtensions (0.03s)
>     mock_topo.go:392: Drop stream fails: ext is not found.
>     mock_topo.go:392: Drop stream fails: ext2 is not found.
>     plugin_rule_test.go:81: The test bucket size is 2.
> ```
> 
> i'll try to fix the 'max commit len' issue -- 58 chars at the moment? ;/

OK， we can try to fix lint and ut problems

dovholuknf:
> OK， we can try to fix lint and ut problems

much appreciated. i found the linter error by looking at the raw logs. the "UI view" just showed all the errors and confused me. i ran gofumpt and hopefully appeased the linter.

dovholuknf:
i think i figured out how to run gci and gofumpt and hopefully the linter will be happy this time

dovholuknf:
> plugin was built with a different version of package github.com/lf-edge/ekuiper/internal/conf/logger (previous failure)

I do think I somehow broke that test now that i see this. I changed this file. I'm definitely going to ask for help figuring out how to fix that. Also thanks for pushing the  'approve' button on all the linter commits. I sure hope I have them all sorted now :(

ngjaying:
> > plugin was built with a different version of package github.com/lf-edge/ekuiper/internal/conf/logger (previous failure)
> 
> I do think I somehow broke that test now that i see this. I changed this file. I'm definitely going to ask for help figuring out how to fix that. Also thanks for pushing the 'approve' button on all the linter commits. I sure hope I have them all sorted now :(

Sure, I'll take a look later today but not now since I am busy with other things in hand. Once you have a commit merged, the checks will be run without a need of approval.

ngjaying:
###  Build packages error
The default build has edgex tag, the build package check will run the built kuiperd to check. Then the caErr cause a panic in openziti.go AuthenicatedContext.
I guess there are some dependency services which are not existed in the test env. Is EdgeX still have secure and non-secure mode? Maybe we need to have a configuration in eKuiper to switch between secure and non-secure and default to non-secure to pass the tests. Thus, I make a commit for that to add enableOpenZiti config.

### UT error
It is caused by the logger adaptor. It seems to me that the adaptor is unneccessary to apply to the logger globally. We can decorate it only when calling open ziti functions. Thus, I make a commit for that.

internal/io/http/client_edgex.go
```diff
@@ -0,0 +1,63 @@
...
+	edgex_vault "github.com/lf-edge/ekuiper/internal/edgex"
+)
+
+func init() {
+	newTransport = newZeroTrustTransport
+}
+
+var zitiTransport *http.Transport
+
+func newZeroTrustTransport(tlscfg *tls.Config, logger *logrus.Logger) *http.Transport {
```
ngjaying, line 40: When using EdgeX secure mode, is the zero trust transport applied to all http request or just apply to the request for EdgeX url? If the latter, we don't need to always return this transport. It should be a property in the REST sink and only those sink with this property need to get the zero trust transport.

dovholuknf: It depends on how the transport is returned. the zero trust transport needs to only intercept the addresses it's configured to intercept but it should definitely fallback to using the regular underlay if the url is not one it's expected to intercept. So it should work in a "split-tunnel" type of way. 

Thanks for asking -- I tested to verify that would be the case and it's not currently. I'll fix that today.

dovholuknf: I've changed the transport to use a fallback dialer which will activate if no matching service is found

internal/io/http/client_edgex.go
```diff
@@ -0,0 +1,72 @@
...
+			return zitiTransport
+		} else {
+			ctx := edgex_vault.AuthenicatedContext(logger)
+
+			zitiContexts := ziti.NewSdkCollection()
+			zitiContexts.Add(ctx)
+
+			zitiTransport = http.DefaultTransport.(*http.Transport).Clone() // copy default transport
+			zitiTransport.TLSClientConfig = tlscfg
+			dialer := zitiContexts.NewDialerWithFallback(context.Background(), &ZitiUnderlayDialer{})
```
dovholuknf, line 63: here's the "Fallback" dialer to allow for split-tunnel type http clients

ngjaying:
@dovholuknf Thank you. All looks good now. Is it possible to add more unit tests? If not, I think we can merge for now. And we'll need doc updates in another PR.

dovholuknf:
> @dovholuknf Thank you. All looks good now. Is it possible to add more unit tests? If not, I think we can merge for now. And we'll need doc updates in another PR.

Thank you for the help! I'd really like to say yes to adding tests but I'm really strapped trying to get this and all the other edgex work out the door. i will put it on my backlog though and maybe in the coming weeks I can add them? Unit tests aren't probably too useful here other than maybe making sure configuration is set and defaults processed since it really needs an openziti overlay -- if that makes sense? I'm happy to look at it in the future, when I can get to it. Right now it's tight. :(

dovholuknf:
I'll need help with what kind of doc and 'how/where' to doc. If you have any pointers on that, please let me know! Maybe make an issue and assign it to me?

dovholuknf:
and one related question - will a docker container be produced from this repo that I could reference in the edgex project? sorry for all the comments...

ngjaying:
> and one related question - will a docker container be produced from this repo that I could reference in the edgex project? sorry for all the comments...

Yes, we can trigger an alpha release, that would produce a docker image.

ngjaying:
> I'll need help with what kind of doc and 'how/where' to doc. If you have any pointers on that, please let me know! Maybe make an issue and assign it to me? -- same for test now that i think about it?

OK, I'll create an issue for that.

ngjaying (approved):

dovholuknf:
> Yes, we can trigger an alpha release

that would be much appreciated. I looked but it doesn't seem like one exists yet. Should I use GH Discussions for this request? Add a new issue? or just keep commenting here? :) thanks again

ngjaying:
> > Yes, we can trigger an alpha release
> 
> that would be much appreciated. I looked but it doesn't seem like one exists yet. Should I use GH Discussions for this request? Add a new issue? or just keep commenting here? :) thanks again

I will tag a version today. Discussions thread is OK for me.

ngjaying:
https://github.com/lf-edge/ekuiper/releases/tag/v1.14.0-alpha.2  tagged

PR #2610, opened 2024-02-05
fix: fix protobuf oneof decode

Yisaer:


internal/converter/ext_converter.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build schema || !core
+//go:build schema || !core || full
```
ngjaying, line 15: This is innecessary. full is a subset of !core

ngjaying (changes requested):

ngjaying (approved):

PR #2624, opened 2024-02-19
feat: check rule before creating it

Yisaer:
1. check rule before creating it
2. fix redisSub conf


internal/server/rule_manager.go
```diff
@@ -83,26 +83,22 @@ func createRule(name, ruleJson string) (string, error) {
...
-		}
-		return r.Id, fmt.Errorf("create rule topo error: %v", panicOrError)
+	if err != nil {
+		return r.Id, err
 	}
+
+	// Store to KV
+	err = ruleProcessor.ExecCreate(r.Id, ruleJson)
+	if err != nil {
+		return r.Id, fmt.Errorf("store the rule error: %v", err)
```
ngjaying, line 99: Need to run deleteRule(r.id) to remove the running rule from rule registry.

ngjaying (changes requested):

internal/server/rule_manager.go
ngjaying: Looks like still not moving deleteRule here. The line 93 deleteRule is not needed since the rule registry fails to create

ngjaying (changes requested):

internal/server/rule_manager.go
```diff
@@ -83,26 +83,24 @@ func createRule(name, ruleJson string) (string, error) {
...
-		// Do not store to registry so also delete the KV
+	if err != nil {
+		return r.Id, err
+	}
+
+	// Store to KV
+	err = ruleProcessor.ExecCreate(r.Id, ruleJson)
+	if err != nil {
 		deleteRule(r.Id)
 		_, _ = ruleProcessor.ExecDrop(r.Id)
```
ngjaying, line 100: This line is not needed now.

internal/server/rule_manager.go
```diff
@@ -83,26 +83,24 @@ func createRule(name, ruleJson string) (string, error) {
...
-		if ewc, ok := panicOrError.(errorx.ErrorWithCode); ok {
-			return r.Id, errorx.NewWithCode(ewc.Code(), fmt.Sprintf("create rule topo error: %v", panicOrError.Error()))
-		}
-		return r.Id, fmt.Errorf("create rule topo error: %v", panicOrError)
+		return r.Id, fmt.Errorf("store the rule error: %v", err)
```
ngjaying, line 101: Is error code needed here?

ngjaying (changes requested):

ngjaying (approved):

PR #2635, opened 2024-02-21
feat: support validate conf before creating

Yisaer:


pkg/api/stream.go
```diff
@@ -146,6 +146,10 @@ type Sink interface {
 	Closable
 }
 
+type PropsValidator interface {
```
ngjaying, line 149: As the experimental interface, shall we move these and ping interface to internal package. So that we can easily remove them for future changes.

ngjaying (changes requested):

ngjaying (approved):

PR #2642, opened 2024-02-26
feat: introduce failpoint and fix resend error

Yisaer:


internal/io/http/rest_sink.go
```diff
@@ -50,25 +60,23 @@ func (ms *RestSink) collectWithUrl(ctx api.StreamContext, item interface{}, desU
 	}
 
 	resp, err := ms.sendWithUrl(ctx, decodedData, item, desUrl)
+	failpoint.Inject("injectRestTemporaryError", func(val failpoint.Value) {
```
ngjaying, line 63: Just curious, is it possible to mock the error in the UT instead of inject an error in the implementation code?

Yisaer: For now it can't

PR #2644, opened 2024-02-27
fix: check redis connector db

Yisaer:


internal/io/redis/lookup.go
```diff
@@ -66,13 +57,32 @@ func (s *lookupSource) Configure(datasource string, props map[string]interface{}
...
+	if datasource != "/$$TEST_CONNECTION$$" {
+		db, err := cast.ToInt(datasource, cast.CONVERT_ALL)
+		if err != nil {
+			return fmt.Errorf("invalid datasource, it must be an integer but got %s", datasource)
+		}
+		s.db = db
+	} else {
+		s.db = 0
+	}
+	if s.db < 0 || s.db > 15 {
```
ngjaying, line 73: Why this is not inside the validate func

Yisaer: Due to the db is passed by datasource in Configure.

ngjaying (approved):

PR #2645, opened 2024-02-27
feat: support validate rule id

Yisaer:


internal/processor/rule.go
```diff
@@ -185,6 +188,25 @@ func (p *RuleProcessor) GetRuleByJson(id, ruleJson string) (*api.Rule, error) {
...
+	for _, char := range id {
+		if !isLegalChar(char) {
+			return fmt.Errorf("ruleID:%v invalid", id)
+		}
+	}
+	return nil
+}
+
+func isLegalChar(char rune) bool {
+	if (char >= '0' && char <= '9') ||
```
ngjaying, line 201: Is Chinese name still supported? Shall we find a common rule which can still tolerate most of the names.

ngjaying (changes requested):

ngjaying (approved):

PR #2661, opened 2024-03-02
docs: fix Chinese doc errors for substring func

Dream95:
substring function not support Negative integer

### DIFF
--- a/docs/zh_CN/sqls/functions/string_functions.md
+++ b/docs/zh_CN/sqls/functions/string_functions.md
@@ -148,7 +148,7 @@ rtrim(col)
 substring(col, start, end)
 ```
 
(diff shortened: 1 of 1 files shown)


ngjaying:
Thanks. Actually, we expect to support negative index. So instead of modifying the doc, shall we modify the code?

Dream95:
What will be substring('abcde',2,-1)  return？

ngjaying:
> What will be substring('abcde',2,-1) return？

-1 means (length-1).  This will return index [2 to 4] which is "cde". Is it reasonable?

ngjaying:
Currently，let's just fix the doc. @Dream95 Could you please take a look at the Details of DCO check and try to pass it. Thanks.

Yisaer (approved):

PR #2670, opened 2024-03-05
fix: fix influx sink configure

Yisaer:


ngjaying (changes requested):
We can add a test in the configure test.

ngjaying (approved):

PR #2675, opened 2024-03-05
feat: union mqtt sink tls error behavior

Yisaer:


ngjaying (commented):
Are there same problems in the mqtt source?

ngjaying (approved):

Issue #2668, opened 2024-03-05
tdengine sink add "fields"  error


gooker:
**Environment**:

- eKuiper version (e.g. `1.3.0`): 1.12.4
- Hardware configuration (e.g. `lscpu`):x86_64 
- OS (e.g. `cat /etc/os-release`):rocky9.3
- Others:

**What happened and what you expected to happen**:

如果设置TDengine的动态插入规则,设置了"fields": 则提示失败,不使用fields无问题
```
status: stopped
message: Stopped: runtime error: invalid memory address or nil pointer dereference.
```
```
{
  "tdengine": {
    "host":        "hostname",
    "port":        6030,
    "database":    "dab",
    "table":       "tableName", 
    "tsfieldname": "ts",
    "fields":      ["f1","f2"],
    "sTable":      "myStable",
    "tagFields":   ["f3","f4"]
  }
}
```

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
不确定是配置问题还是程序问题.

Yisaer:
@gooker

Please provide the log in the eKuiper which related to the `invalid memory address or nil pointer dereference.`

gooker:
> @gooker
> 
> Please provide the log in the eKuiper which related to the `invalid memory address or nil pointer dereference.`

```



time="2024-03-05T10:46:53+08:00" level=debug msg="rule tdengine11 origin state: Stopped: canceled manually., sginal: 0" file="server/server.go:353"
time="2024-03-05T10:46:53+08:00" level=debug msg="rule tdengine12 origin state: Stopped: runtime error: invalid memory address or nil pointer dereference., sginal: 0" file="server/server.go:353"
time="2024-03-05T10:46:53+08:00" level=debug msg="rule tdenginezs origin state: Stopped: canceled manually., sginal: 0" file="server/server.go:353"
time="2024-03-05T10:46:54+08:00" level=info msg="Rule tdengine12 is replaced." file="processor/rule.go:111"
time="2024-03-05T10:46:54+08:00" level=info msg="Init rule with options &{Debug:false LogFilename: IsEventTime:false LateTol:1000 Concurrency:1 BufferLength:1024 SendMetaToSink:false SendError:true Qos:0 CheckpointInterval:300000 Restart:0xc0009cccf0 Cron: Duration: CronDatetimeRange:[]}" file="planner/planner.go:44"
[...]

PR #2680, opened 2024-03-06
fix: tdengine plugins rules stopped when fields not in streams

gooker:
fix: tdengine plugins rules stopped when fields not in streams
fix: tdengine tags not in fields sqlbuild error (#2668)


ngjaying:
Thank you. Please follow the Details link in the DCO check to signoff.

PR #2713, opened 2024-03-15
feat: add conf to remain rule metrics

Yisaer:


ngjaying (commented):
Did you consider delete and then create a rule with the same id? Will this clean the old stats?

Yisaer:
> Did you consider delete and then create a rule with the same id? Will this clean the old stats?

We thought it as the same rule.

ngjaying:
> > Did you consider delete and then create a rule with the same id? Will this clean the old stats?
> 
> We thought it as the same rule.

Will the metrics increasing from the last state. For example, the old instance has 100 records in, after re-create the rule, will you start from 101?

Yisaer:
> > > Did you consider delete and then create a rule with the same id? Will this clean the old stats?
> > 
> > 
> > We thought it as the same rule.
> 
> Will the metrics increasing from the last state. For example, the old instance has 100 records in, after re-create the rule, will you start from 101?

yes

PR #2716, opened 2024-03-18
feat: introduce rule metrics

Yisaer:
introduce rule metrics:
1. `kuiper_rule_count` to record count of the running and stop rule
2. `kuiper_rule_status` to record the rule status by ruleID

image


internal/server/server.go
```diff
@@ -351,6 +352,33 @@ func runScheduleRuleChecker(exit <-chan struct{}) {
 	runScheduleRuleCheckerByInterval(d, exit)
 }
 
+func handleAllRuleStatusMetrics(rs []ruleWrapper) {
+	if conf.Config != nil && conf.Config.Basic.Prometheus {
```
ngjaying, line 356: This seems like setting in initialization. Do the metrics updated in runtime when start/stop rules?

Yisaer: The metircs updated in every interval (10s)

ngjaying (approved):

PR #2727, opened 2024-03-22
feat: let IntegerLiteral Value as int64

Yisaer:


internal/binder/function/funcs_misc.go
```diff
@@ -249,16 +251,22 @@ func registerMiscFunc() {
 		fType: ast.FuncTypeScalar,
 		exec: func(ctx api.FunctionContext, args []interface{}) (interface{}, bool) {
 			var v0 float64
-			if v1, ok := args[0].(int); ok {
+			switch v1 := args[0].(type) {
```
ngjaying, line 254: Just use cast function here

ngjaying (approved):

PR #2730, opened 2024-03-25
feat: support modify sqlSource field index value

Yisaer:


ngjaying (commented):
I prefer a general state read/write API, not specific to SQL index value.

extensions/sources/sql/ext/sql.go
```diff
@@ -41,6 +42,9 @@ type sqlsource struct {
 	Query sqlgen.SqlQueryGenerator
 	// The db connection instance
 	db *sql.DB
+
+	RuleID     string
```
ngjaying, line 46: Actually, these information are already in the context. Is it possible not to add another ref? I don't think the SQL source buisiness logic relates to the rule name, stream name.

ngjaying (changes requested):

extensions/sources/sql/ext/sql.go
Yisaer: updated.

internal/server/rule_state.go
```diff
@@ -0,0 +1,105 @@
...
+	if s != rule.RuleStarted {
+		return fmt.Errorf("rule %v should be running when modify state", ruleID)
+	}
+
+	req := &resetOffsetRequest{}
+	if err := cast.MapToStruct(param, req); err != nil {
+		return err
+	}
+	switch strings.ToLower(req.OffsetType) {
+	case "sql":
```
ngjaying, line 96: Not sure if this is needed? Because the type is already controlled in Topo

ngjaying (approved):

PR #2736, opened 2024-03-26
fix: replace password when create rule

Yisaer:


internal/server/rest_test.go
```diff
@@ -95,6 +96,7 @@ func (suite *RestTestSuite) SetupTest() {
 	r.HandleFunc("/data/import", configurationImportHandler).Methods(http.MethodPost)
 	r.HandleFunc("/data/import/status", configurationStatusHandler).Methods(http.MethodGet)
 	r.HandleFunc("/connection/websocket", connectionHandler).Methods(http.MethodGet, http.MethodPost, http.MethodDelete)
+	r.HandleFunc("/metadata/sinks/{name}/confKeys/{confKey}", sinkConfKeyHandler).Methods(http.MethodDelete, http.MethodPut)
```
ngjaying, line 99: Already have a similar API defined in meta_init.go

ngjaying (changes requested):

internal/server/rest_test.go
Yisaer: This is in test

ngjaying (approved):

PR #2741, opened 2024-03-27
fix: check rule exists before create

Yisaer:


internal/server/rule_manager.go
```diff
@@ -85,6 +85,10 @@ func createRule(name, ruleJson string) (string, error) {
 		return "", fmt.Errorf("invalid rule json: %v", err)
 	}
 
+	if exists := ruleProcessor.ExecExists(r.Id); exists {
+		return "", fmt.Errorf("rule %v already exists", r.Id)
+	}
+
 	// Validate the topo
```
ngjaying, line 92: How about move deleteRule to defer as it is a clean up action?

internal/server/rule_manager.go
```diff
@@ -75,16 +75,25 @@ func (rr *RuleRegistry) Delete(key string) (*rule.RuleState, bool) {
 	return result, ok
 }
 
-func createRule(name, ruleJson string) (string, error) {
+func createRule(name, ruleJson string) (id string, err error) {
 	var rs *rule.RuleState = nil
-	var err error = nil
+	defer func() {
```
ngjaying, line 80: This is wrong, should move this to line 105 which the new rule state has been created. Put here may delete the running rule registry

ngjaying (changes requested):

Issue #2590, opened 2024-01-26
Why do the stdev and var aggregate functions not ignore null values like sum, min, max, and avg ?


ngjaying:
### Discussed in https://github.com/lf-edge/ekuiper/discussions/2589

<div type='discussions-op-text'>

<sup>Originally posted by **EscanorUt** January 26, 2024</sup>
Hello,

I've noticed that while using eKuiper, the stdev and var aggregate functions seem to consider null values in their calculations, unlike other aggregate functions like sum, min, max, and avg which ignore null values. Can anyone shed light on the reason behind this behavior and whether there are any workarounds or alternative approaches to handle null values with stdev and var functions in eKuiper

Thank you</div>

BNNARAJ:
Hello @ngjaying sir , 
I am Prabal Pratap Singh Rathore , second - year student of Btech in Artificial Intelligence and Data Science. I am good at python and several libraries with experience in ML and DL with keras, tensorflow and Pytorch. I want to look into this issue , I am currently exploring eKuiper , So please assign this Good First Issue to me .

BNNARAJ:
we have to see how this function stdev and var are implemented internally because aggregate functions exclude null values for calculation but for these function involves stastical calculation that can be mislead by null values. So, please can you navigate to the directory or code file to see how they are present in code.

ngjaying:
@BNNARAJ Sorry for the late response. The functions are in https://github.com/lf-edge/ekuiper/blob/master/internal/binder/function/funcs_agg.go. Actually, you can do a search in the codebase to find it next time.

BNNARAJ:
Thank you sir

BNNARAJ:
Hello @ngjaying sir, 
In the definition of the stddev and var function, we can see the cast function which is reconstructing the float64slice and removing null values "float64Slice, err := cast.ToFloat64Slice(arg0, cast.CONVERT_SAMEKIND, cast.IGNORE_NIL)", so it looks like it does take null values but while calculating it is reconstructing the input array while removing null values by using cast.IGNORE_NIL . 
As I am a beginner please ensure that I am following the correct understanding and approach.

ngjaying:
Hi @BNNARAJ,
Maybe it was fixed. Could you try to add a ut case to confirm that? If that's the case, we can push the test case as a PR to close this issue. Thanks!

BNNARAJ:
Hello @ngjaying  Sir ,
[...]

PR #2748, opened 2024-03-29
fix: ignore nil value in array convert

carlclone:


carlclone:
#2590

pkg/cast/cast.go
```diff
@@ -905,14 +913,22 @@ func ToUint64Slice(input interface{}, sn Strictness) ([]uint64, error) {
 	return result, nil
 }
 
-func ToFloat64Slice(input interface{}, sn Strictness) ([]float64, error) {
+func ToFloat64Slice(input interface{}, sn Strictness, anc ARRAY_NIL_CONVERT) ([]float64, error) {
```
ngjaying, line 916: Shall we make similar changes to other ToXXXSlice functions?

carlclone: Those methods don't seem to affect the aggregation function. Should I modified them ? maybe in a new pull request .

ngjaying (approved):
Thank you @carlclone

Issue #2726, opened 2024-03-22
Add 'type' and 'format' to list api of '/streams' and '/tables'


oucb:
**What would you like to be added/modified**:
Add 'type' and 'format' to list api of '/streams' and '/tables'
Response example: [{name: 'xxx', type: 'mqtt', format: 'json'}, ...]

**Why is this needed**:
Convenient for more detailed information of stream, table

PR #2750, opened 2024-03-31
feat: add 'format' and 'type' fields in list apis

wfnuser:
Instead of returning name of tables and streams, we can return xsql.StreamInfo directly.

### DIFF
--- a/internal/server/rest.go
+++ b/internal/server/rest.go
@@ -35,6 +35,7 @@ import (
 	"github.com/lf-edge/ekuiper/internal/conf"
 	"github.com/lf-edge/ekuiper/internal/pkg/httpx"
 	"github.com/lf-edge/ekuiper/internal/pkg/store"
+	"github.com/lf-edge/ekuiper/internal/processor"
 	"github.com/lf-edge/ekuiper/internal/server/middleware"
 	"github.com/lf-edge/ekuiper/internal/topo/planner"
 	"github.com/lf-edge/ekuiper/internal/trial"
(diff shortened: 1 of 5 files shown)


wfnuser:
I've been following your project with great interest recently. I hope this PR can address issue #2726 .
I'm not sure this is the best way to do it. Since I'm still getting familiar with this project. I hope some one will help me to merge this pr.

The major change is that instead of returning stream names we returning unmarshaled StreamInfo directly in this pr. Since the Unmarshal need to be done anyway, we won't introduce any additional cpu overhead. Though we need to change the stream format we store in db.

wfnuser:
And if there are more issues you think can help me to get familiar with this project. Please let me know, I'd like to contribute.

wfnuser:
It seems there are some unit test failures, I will check it some day next week.

Yisaer:
@wfnuser I tended to add param to separte the returning value for `GET /streams` and `GET /tables` in order not to break the REST API. I think you can take an example in https://github.com/lf-edge/ekuiper/pull/2756

wfnuser:
> @wfnuser I tended to add param to separte the returning value for `GET /streams` and `GET /tables` in order not to break the REST API. I think you can take an example in #2756

How about we simply add a new api for detail and keep the original api behavior? like `GET /streamInfos` `GET /tableInfos`

Yisaer:
> seems a little bit wierd to me to add a param control if we return detailed

That's ok to me.

Yisaer:
> > @wfnuser I tended to add param to separte the returning value for `GET /streams` and `GET /tables` in order not to break the REST API. I think you can take an example in #2756
> 
> @Yisaer How about we simply add a new api for detail and keep the original api behavior unchanged? Considering path like `GET /streamInfos` `GET /tableInfos` It seems a little bit wierd to me to add a param control if we return detailed info.

That's ok to me.

wfnuser:
@Yisaer Updated. I add new apis for detailed info. But instead of parse the stream again in a for loop. I prefer to use the result unmarshaled from db. Though, I'm not sure if there are some other concerns.

Yisaer (dismissed):

internal/server/rest.go
```diff
@@ -423,6 +426,35 @@ func pingHandler(w http.ResponseWriter, _ *http.Request) {
...
+			kind = ast.StreamKindLookup
+		} else {
+			kind = ""
+		}
+	}
+	if kind != "" {
+		content, err = streamProcessor.ShowTableInfo(kind)
+	} else {
+		content, err = streamProcessor.ShowStreamInfo(st)
+	}
```
Yisaer, line 450: I think the users who upgraded eKuiper from old version can not use this api as here we read data from the database.

I think we can use `streamProcessor.DescStream` here to get the info we want.

Yisaer (changes requested):
ditto.

internal/server/rest.go
wfnuser: Get. I will revert the change the db schema. But I think DescStream might still be avoided by adding a new structure to return stream detail, I will change that later.

wfnuser: Sorry. My thought was wrong. If we don't change the db schema, we have to descstream. And the code is updated.
But I'm wondering if some times in the future you have to change the database schema, how to deal with that?
I'm not quite familiar with SQLite, but I thought if we design the structure carefully, maybe it will always be forward compatible.

Like, if we only append new fields to the xsql.StreamInfo{}, the old data can still be serializable? I don't know if it is possible. Some database can do that automatically I guess.

wfnuser:
@ngjaying There is a flaky tests failure again. Can you retrigger the tests for me, thanks a lot.
BTW, shall we fix this flaky failure in CI? Is it unavoidable?

Yisaer (approved):

ngjaying (changes requested):
Need to modify docs as well

wfnuser:
@ngjaying  Add the docs.

ngjaying (approved):

ngjaying:
Thans you @wfnuser

Issue #2714, opened 2024-03-15
Add Gzip/Brotli compression to REST sink


webfrank:
**What would you like to be added/modified**:

Add Gzip/Brotli, or any other compatible compression algorithm, to REST sink.

**Why is this needed**:

When sending large JSON/XML payload, compressing POST would benefit in terms of latency.

yai-dev:
@ngjaying Hi there, is this issue still open and i saw there has releated pr has been closed, can i create pr for this issue?

ngjaying:
@yai-dev I think @carlclone is working on this now. @carlclone Can I assign this to you?

yai-dev:
> @yai-dev I think @carlclone is working on this now. @carlclone Can I assign this to you?

okay, i see this issue has been opened for a long time (about 1 month at least) and this feature is not complex, i will look another issue to be fixing, if @carlclone has long time to not respond, i will try to working on this.

carlclone:
yes , feel free to contribute  , i don't have enough time recently , you can continue this issue

yai-dev:
> yes , feel free to contribute , i don't have enough time recently , you can continue this issue

OK, thx pal

PR #2753, opened 2024-04-02
feat: rest sink compress

carlclone:


carlclone:
#2714

internal/io/http/rest_sink_test.go
```diff
@@ -149,7 +154,8 @@ func TestRestSink_Apply(t *testing.T) {
 				ContentType: "application/x-www-form-urlencoded;param=value",
 				Body:        `ab=hello2`,
 			}},
-		}, {
+		},
```
wfnuser, line 157: Are these lint format results or just some irrelevant commits?

internal/io/http/rest_sink.go
```diff
@@ -16,21 +16,23 @@ package http
...
 	"fmt"
+	"github.com/pingcap/failpoint"
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
 
-	"github.com/pingcap/failpoint"
-
+	"github.com/lf-edge/ekuiper/internal/compressor"
```
wfnuser, line 25: I guess there are some lint problems here.

internal/io/http/rest_sink.go
```diff
@@ -40,7 +42,17 @@ func (ms *RestSink) Validate(props map[string]interface{}) error {
...
 func (ms *RestSink) Configure(ps map[string]interface{}) error {
 	conf.Log.Infof("Initialized rest sink with configurations %#v.", ps)
-	return ms.Validate(ps)
+	err := ms.Validate(ps)
+	if err != nil {
+		return err
+	}
+	if ms.config.Compression != "" {
+		ms.compressor, err = compressor.GetCompressor(ms.config.Compression)
+		if err != nil {
```
wfnuser, line 51: I guess the compressor.GetCompressor error message is clear enough. Though, no harm to make it more specific.

internal/io/http/rest_sink_test.go
```diff
@@ -565,3 +597,9 @@ func TestRestSinkMethod(t *testing.T) {
 	err := rs.Validate(map[string]interface{}{"method": "head"})
 	require.Error(t, err)
 }
+
+func TestRestSinkConfigure(t *testing.T) {
```
wfnuser, line 601: Better to make this test name clearly showing it's a failure test. I guess the intention is to test unsupported compression method.

internal/io/http/rest_sink.go
carlclone: thanks , fixed

internal/io/http/rest_sink.go
carlclone: fixed

internal/io/http/rest_sink_test.go
carlclone: fixed

wfnuser (approved):

ngjaying:
@carlclone The FVT tests fail. Do you need help on that?

internal/io/http/rest_sink.go
```diff
@@ -64,6 +79,13 @@ func (ms *RestSink) collectWithUrl(ctx api.StreamContext, item interface{}, desU
 		return fmt.Errorf("rest sink decode data error: %v", err)
 	}
 
+	if ms.compressor != nil {
+		decodedData, err = ms.compressor.Compress(decodedData)
+		if err != nil {
+			return err
+		}
+	}
+
```
ngjaying, line 88: Http header needs to be set when using compression like `req.Header.Set("Content-Encoding", "gzip")`

ngjaying (changes requested):

internal/io/http/rest_sink.go
carlclone: the Content-Encoding header seems to only set in the response rather than request

carlclone:
will create a new pr , this may not being considerate enough

PR #2755, opened 2024-04-02
feat: support rule stop by error metrics 

Yisaer:


internal/server/server.go
```diff
@@ -372,15 +372,27 @@ func handleAllRuleStatusMetrics(rs []ruleWrapper) {
...
 		for _, r := range rs {
 			id := r.rule.Id
-			isRunning := r.state == rule.RuleStarted
-			if isRunning {
+			switch r.state {
+			case rule.RuleStarted:
 				runningCount++
-			} else {
+				v = promMetrics.RuleRunning
+			case rule.RuleStopped:
```
ngjaying, line 382: You can have a case with multiple conditions to avoid code duplication. `case rule.RuleStopped, rule.RuleTerminated, rule.RuleWait:`

ngjaying (changes requested):

internal/server/promMetrics/metrcis.go
```diff
@@ -59,12 +59,17 @@ func SetRuleStatusCountGauge(isRunning bool, count int) {
...
 
-func SetRuleStatus(ruleID string, isRunning bool) {
-	v := 0
-	if isRunning {
-		v = 1
-	}
-	RuleStatusGauge.WithLabelValues(ruleID).Set(float64(v))
+type RuleStatusMetricsValue int
+
+const (
```
ngjaying, line 64: Can we move this to another package? It does not make sense for server to refer to this package, especially prometheus can be turned off.

ngjaying (changes requested):

ngjaying (approved):

PR #2781, opened 2024-04-17
feat: support async task framework

Yisaer:


pkg/async/async.go
```diff
@@ -0,0 +1,228 @@
+// Copyright 2024 EMQ Technologies Co., Ltd.
```
ngjaying, line 1: Shall we move this into internal/pkg

ngjaying (approved):

PR #2784, opened 2024-04-18
feat: introduce ctx for data import api

Yisaer:
introduce ctx for data import api in order to support cancel in async api in future.

### DIFF
--- a/internal/plugin/js/export_test.go
+++ b/internal/plugin/js/export_test.go
@@ -19,6 +19,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	"github.com/lf-edge/ekuiper/internal/topo/context"
 )
 
 func TestFullLifecycle(t *testing.T) {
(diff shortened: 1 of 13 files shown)


internal/plugin/js/export_test.go
```diff
@@ -54,7 +56,7 @@ func TestFullLifecycle(t *testing.T) {
 		"testScript": "{\"id\":\"testScript\",\"description\":\"Test script\",\"script\":\"function testScript() { return 'Hello, new!'; }\",\"isAgg\":false}",
 		"invalid":    "{\"id\":\"novalid\",\"description\":\"invalid script\",\"script\":\"function novalid() { return 'Hello, invalid!'; }\",\"isAgg\":false}",
 	}
-	errMap := m.PartialImport(invalidSet)
+	errMap := m.PartialImport(context.Background(), invalidSet)
```
ngjaying, line 59: Why use context.Background?

ngjaying (approved):

PR #2813, opened 2024-04-26
feat: support separate schema layer

Yisaer:


ngjaying (commented):
Can we move this to v2. There is no need to change 1.x for now

PR #2815, opened 2024-04-26
feat: separate schema layer from decoder

Yisaer:


internal/converter/converter.go
```diff
@@ -55,7 +55,7 @@ func GetOrCreateConverter(options *ast.Options) (c message.Converter, err error)
 		if options.RuleID == "" || options.StreamName == "" {
 			return json.GetConverter()
 		}
-		return json.NewFastJsonConverter(options.RuleID, options.StreamName, options.Schema, options.IsWildCard, options.IsSchemaLess), nil
+		return json.NewFastJsonConverter(options.Schema, options.IsWildCard, options.IsSchemaLess), nil
```
ngjaying, line 58: Is it possible to move isWildcard out? Converter should not take care of that?

ngjaying (approved):

PR #2817, opened 2024-04-28
feat: support async import api

Yisaer:


internal/server/rest.go
```diff
@@ -184,6 +184,8 @@ func createRestServer(ip string, port int, needToken bool) *http.Server {
 	r.HandleFunc("/data/import", configurationImportHandler).Methods(http.MethodPost)
 	r.HandleFunc("/data/import/status", configurationStatusHandler).Methods(http.MethodGet)
 	r.HandleFunc("/connection/websocket", connectionHandler).Methods(http.MethodGet, http.MethodPost, http.MethodDelete)
+	r.HandleFunc("/async/task", asyncTaskHandler).Methods(http.MethodGet, http.MethodPost)
```
ngjaying, line 187: How about data/import/async? Or async/data/import? By this way, each task can have static body and easy to understand.

Yisaer: updated.

ngjaying (approved):

PR #2839, opened 2024-05-10
fix: revise interval config

Yisaer:


internal/conf/conf.go
```diff
@@ -62,7 +62,7 @@ type SinkConf struct {
 	MaxDiskCache         int    `json:"maxDiskCache" yaml:"maxDiskCache"`
 	BufferPageSize       int    `json:"bufferPageSize" yaml:"bufferPageSize"`
 	EnableCache          bool   `json:"enableCache" yaml:"enableCache"`
-	ResendInterval       int    `json:"resendInterval" yaml:"resendInterval"`
+	ResendInterval       string `json:"resendInterval" yaml:"resendInterval"`
```
ngjaying, line 65: Is it possible to parse into time.Duration directly?

deploy/chart/ekuiper/values.yaml
```diff
@@ -160,7 +160,7 @@ kuiperConfig:
       # to enable the checkpoint mechanism
       qos: 0
       # The interval in millisecond to run the checkpoint mechanism.
-      checkpointInterval: 300000
+      checkpointInterval: 300s
```
ngjaying, line 163: Comment has to be changed. Is it possible to compatible with int, which means default to millisecond if no duration unit.

internal/conf/conf.go
Yisaer: no

deploy/chart/ekuiper/values.yaml
Yisaer: Compatible with int will make it too complex as configuration.

ngjaying (approved):

PR #2846, opened 2024-05-12
ci(lint): bump `golangci-lint-action` to v6

xjasonlyu:
This PR mainly updates `golangci-lint-action` to v6, and also automatically specifies the go version from `go.mod` for further convenience.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -18,12 +18,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Set up Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
         with:
-          go-version: '1.22'
+          go-version-file: 'go.mod'
 
       - name: Check licenses
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yaml
```diff
@@ -18,12 +18,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
ngjaying, line 21: Shall we update all checkout action in all workflows?

xjasonlyu: Yea, but it's up to you. These two versions are compatible.

.github/workflows/lint.yaml
```diff
@@ -18,12 +18,12 @@ jobs:
...
 
       - name: Set up Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
         with:
-          go-version: '1.22'
+          go-version-file: 'go.mod'
```
ngjaying, line 26: This is handy, don't need to define the version everywhere

xjasonlyu: Yes, and we can use this at other workflows too.

Rory-Z (approved):

Issue #2521, opened 2024-01-02
Support parquet format


ngjaying:
**What would you like to be added/modified**:
Add parquet format for source/sink

**Why is this needed**:
Parquet is a columnar storage file format optimized for use with big data processing frameworks.  Parquet is widely used in big data processing.

PR #2847, opened 2024-05-13
feat: support parquet file source

forsaken628:
#2521


internal/io/file/parquet_reader.go
```diff
@@ -0,0 +1,92 @@
...
+
+type ParquetReader struct {
+	f          *os.File
+	pf         *parquet.File
+	groups     []parquet.RowGroup
+	curGroup   int
+	rowsReader parquet.Rows
+}
+
+func CreateParquetReader(filename string, config *FileSourceConfig, ctx api.StreamContext) (FormatReader, error) {
```
ngjaying, line 33: Recommend to put ctx as the first arg

internal/io/file/parquet_reader.go
```diff
@@ -0,0 +1,92 @@
...
+	var pr ParquetReader
```
ngjaying, line 34:
```
var (
  pr ParquetReader
  err error
)
```

ngjaying (changes requested):
Is the parquet mod big? Can you check how big the binary size will increase? Thanks.

forsaken628:
> Is the parquet mod big? Can you check how big the binary size will increase? Thanks.

before
24M kuiper
38M kuiperd
after
27M kuiper
41M kuiperd
There is another option, but I haven't compared it before.
https://pkg.go.dev/github.com/apache/arrow/go/v17/parquet

ngjaying:
> > Is the parquet mod big? Can you check how big the binary size will increase? Thanks.
> 
> linux before 24M kuiper 38M kuiperd after 27M kuiper 41M kuiperd There is another option, but I haven't compared it before. https://pkg.go.dev/github.com/apache/arrow/go/v17/parquet

Great, thanks! If anyone complians about the size, we can make it optional by build tag later.

ngjaying (approved):

PR #2850, opened 2024-05-15
feat: support redis/influx ping

Yisaer:


internal/io/redis/lookup.go
```diff
@@ -44,6 +44,16 @@ type lookupSource struct {
 	cli *redis.Client
 }
 
+func (s *lookupSource) Ping(dataSource string, props map[string]interface{}) error {
+	err := s.Configure(dataSource, props)
+	defer func() {
+		if s.cli != nil {
+			s.cli.Close()
```
ngjaying, line 51: Why close?

ngjaying (changes requested):

internal/io/redis/lookup.go
ngjaying: but the ping seems not happen. Is it inside configure?

ngjaying (approved):

Issue #2525, opened 2024-01-04
Gracefully stop rules when exit eKuiper


ngjaying:
Currently, eKuiper just exit immediately. Need to let rule clean up, especially when using file sink.

PR #2862, opened 2024-05-20
fix: support wait operators close

Yisaer:


contract/api/ctx.go
```diff
@@ -45,6 +45,7 @@ type StreamContext interface {
 	GetOpId() string
 	GetInstanceId() int
 	GetRootPath() string
+	GetRuleWaitGroup() *sync.WaitGroup
```
ngjaying, line 48: Looks like this is an internal signal, do not need to export here? Just pass into node when creating topo/planning.

ngjaying (changes requested):

internal/conf/conf.go
```diff
@@ -356,6 +357,10 @@ func InitConf() {
 		}
 	}
 
+	if time.Duration(Config.Basic.GracefulShutdownTimeout) < 1 {
```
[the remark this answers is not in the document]
Yisaer: it's expected.

contract/api/ctx.go
Yisaer: updated.

ngjaying (approved):

ngjaying:
Closes #2525

PR #2867, opened 2024-05-22
feat: support manage connection

Yisaer:


ngjaying (changes requested):

internal/io/connection/connection_pool.go
```diff
@@ -0,0 +1,271 @@
...
+var ConnectionRegister map[string]RegisterConnection
+
+func init() {
+	ConnectionRegister = map[string]RegisterConnection{}
+	ConnectionRegister["mock"] = createMockConnection
+}
+
+var isTest bool
+
+type Connection interface {
```
ngjaying, line 40: Suggest to move this interface and the register methods to pkg/modules. This will decouple future connection extension from the pool implementation. Basically, this internal package should not contain any public interface (possible implemented by extended features)

internal/io/connection/connection_pool.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+type ConnectionMeta struct {
+	ID    string         `json:"id"`
+	Typ   string         `json:"typ"`
+	Props map[string]any `json:"props"`
+	conn  Connection     `json:"-"`
+}
+
+type mockConnection struct {
```
ngjaying, line 236: Put this into _test file?

ngjaying (changes requested):

internal/io/connection/connection_pool.go
Yisaer: mockConnection should be remained for the test in other packages.

internal/io/mqtt/client/client.go
```diff
@@ -224,6 +221,27 @@ func validateConfig(props map[string]any) (*ConnectionConfig, error) {
...
+func CreateAnonymousConnection(ctx api.StreamContext, props map[string]any) (*Connection, error) {
+	cli, err := CreateClient(ctx, "", props)
+	if err != nil {
+		return nil, err
+	}
+	return cli, nil
+}
+
+const (
+	dataSourceProp = "datasource"
```
ngjaying, line 233: datasource prop is topic prop, why separate them?

ngjaying (changes requested):

ngjaying (approved):

PR #2872, opened 2024-05-23
fix: bump alpine version

Yisaer:


deploy/docker/Dockerfile
```diff
@@ -21,7 +21,7 @@ WORKDIR /go/kuiper
 
 RUN make build_with_edgex
 
-FROM alpine:3.16
+FROM alpine:3.19.1
```
ngjaying, line 24: Shall we use the latest 3.20

ngjaying (approved):

PR #2880, opened 2024-05-28
feat: support store configuration in kv

Yisaer:


etc/kuiper.yaml
```diff
@@ -57,8 +57,8 @@ basic:
     # 0 indicates unlimited
     maxConnections: 0
   rulePatrolInterval: 10s
-  # cfgStorageType indicates the storage type to store the config, support "file","sqlite" and "fdb"
-  cfgStorageType: file
+  # cfgStorageType indicates the storage type to store the config, support "kv"
```
ngjaying, line 60: If this is no use, just delete and hide it.

Yisaer: updated.

ngjaying: I mean, shall we just remove this configuration from kuiper.yaml? Which means we do not encourage users to set this spec.

ngjaying (changes requested):

ngjaying (approved):

PR #2896, opened 2024-06-03
fix: sql source reconnect will check connection

Yisaer:


ngjaying (commented):
- Unrecoverable connection error should stop the rule
- Recoverable connection error should retry until connected

ngjaying (approved):

PR #2898, opened 2024-06-04
feat: support portable plugin management

Yisaer:


internal/plugin/portable/runtime/plugin_ins_manager.go
```diff
@@ -300,27 +348,30 @@
 				if ins.ctrlChan != nil {
 					_ = ins.ctrlChan.Close()
 				}
-				p.deletePluginIns(pluginMeta.Name)
 			}
 			ins.process = nil
 			ins.Unlock()
+			p.deletePluginIns(pluginMeta.Name)
```
ngjaying, line 354: If the plugin instance is deleted, is the plugin still in the status list?

ngjaying (approved):

PR #2906, opened 2024-06-05
feat: support connection status api

Yisaer:


internal/server/connection.go
```diff
@@ -31,6 +31,16 @@ type ConnectionRequest struct {
 	Props map[string]interface{} `json:"props"`
 }
 
+func connectionsStatusHandler(w http.ResponseWriter, r *http.Request) {
```
ngjaying, line 34: Will it run a long time? Is it a problem?

ngjaying (approved):

PR #2933, opened 2024-06-17
feat: support sql source

Yisaer:


internal/io/sql/source.go
```diff
@@ -0,0 +1,208 @@
...
+	ctx.GetLogger().Infof("Closing sql source connector url:%v", s.conf.DBUrl)
+	if s.conn != nil {
+		id := fmt.Sprintf("%s-%s-%s-sql-source", ctx.GetRuleId(), ctx.GetOpId(), s.conf.DBUrl)
+		connection.DetachConnection(ctx, id, s.props)
+		s.conn.DetachSub(ctx, s.props)
+	}
+	return nil
+}
+
+func (s *SQLSourceConnector) Subscribe(ctx api.StreamContext, ingest api.TupleIngest, ingestError api.ErrorIngest) error {
```
ngjaying, line 104: Try to implement PullTupleSource. The source  can only focus on how to pull, do not need to take care about the ticker.

ngjaying (changes requested):

pkg/dbpool/pool.go
```diff
@@ -0,0 +1,153 @@
...
+		return count
+	}
+	return 0
+}
+
+func (dp *dbPool) getOrCreate(url string) (*sql.DB, error) {
+	hiddenURL, _ := hidden.HiddenURLPasswd(url)
+	dp.Lock()
+	defer dp.Unlock()
+	db, ok := dp.pool[url]
```
ngjaying, line 68: It looks like the same logic is generalized and implemented in the connection pool. Can we merge this into that?

ngjaying (approved):

PR #2966, opened 2024-07-01
feat: support rest sink

Yisaer:


internal/io/http/rest_sink.go
```diff
@@ -0,0 +1,121 @@
...
+	return err
+}
+
+func (r *RestSink) collect(ctx api.StreamContext, data map[string]any) error {
+	logger := ctx.GetLogger()
+	headers, err := r.parseHeaders(ctx, data)
+	if err != nil {
+		return err
+	}
+	bodyType, err := ctx.ParseTemplate(r.config.BodyType, data)
```
ngjaying, line 68: Template is now parsing in transform op and pass to sink as "dynamic property". Check mqtt sink line 91about how to use that.

ngjaying (changes requested):

ngjaying (approved):

PR #2968, opened 2024-07-01
feat: support httppush source

Yisaer:


internal/io/http/httppush_source.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (h *HttpPushSource) Provision(ctx api.StreamContext, configs map[string]any) error {
+	cfg := &PushConf{
+		Method:       http.MethodPost,
+		ContentType:  "application/json",
+		BufferLength: 1024,
+	}
+	err := cast.MapToStruct(configs, cfg)
+	if err != nil {
+		return err
```
ngjaying, line 51: UT should cover the provision errors

internal/io/http/httppush_source.go
```diff
@@ -0,0 +1,102 @@
...
+	httpserver.UnregisterEndpoint(h.conf.DataSource)
+	return nil
+}
+
+func (h *HttpPushSource) Connect(ctx api.StreamContext) error {
+	t, err := httpserver.RegisterEndpoint(h.conf.DataSource, h.conf.Method)
+	if err != nil {
+		return err
+	}
+	h.ch = pubsub.CreateSub(t, nil, fmt.Sprintf("%s_%s_%d", ctx.GetRuleId(), ctx.GetOpId(), ctx.GetInstanceId()), h.conf.BufferLength)
```
ngjaying, line 78: Do not use pubsub now. The current implementation cannot start two stream with the same endpoint.. Try to comply to connection pool framework: each endpoint is a "connection". When doing connect, get the connection from the pool. Checkout neuron/mqtt for example.

ngjaying (changes requested):

internal/io/http/httpserver/data_server.go
```diff
@@ -0,0 +1,139 @@
...
+		return ch, nil
+	} else {
+		ch = make(chan *xsql.Tuple, 1024)
+		m.endpoint[endpoint] = ch
+	}
+	m.Unlock()
+	m.router.HandleFunc(endpoint, func(w http.ResponseWriter, r *http.Request) {
+		defer r.Body.Close()
+		ma := make(map[string]interface{})
+		err := json.NewDecoder(r.Body).Decode(&ma)
```
ngjaying, line 96: Better to return bytes only, thus this source can work with all kinds of format which already supported

internal/io/http/httpserver/data_server.go
```diff
@@ -0,0 +1,139 @@
...
+		if err != nil {
+			handleError(w, err, "Fail to decode data")
+			return
+		}
+		dataCh := m.GetCh(endpoint)
```
ngjaying, line 101: dataCh cannot broadcast to multiple sources.  How about provide a subscribe func with a callback func passed in like MQTT client? When receive data, broadcast to all subscribers.

ngjaying (changes requested):

internal/io/http/httppush_source.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return err
+	}
+	hc, ok := c.(*httpserver.HttpPushConnection)
+	if !ok {
+		return fmt.Errorf("connection isn't httppushConnection")
+	}
+	h.sourceID = fmt.Sprintf("%s_%s_%v", ctx.GetRuleId(), ctx.GetOpId(), ctx.GetInstanceId())
+	h.topic = hc.GetTopic()
+	ch := pubsub.CreateSub(h.topic, nil, h.sourceID, 1024)
```
ngjaying, line 84: Move this to subscribe?

internal/io/http/httppush_source.go
```diff
@@ -0,0 +1,113 @@
...
+	h.ch = ch
+	go func(ctx api.StreamContext) {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case v := <-h.ch:
+				data, ok := v.([]byte)
+				if !ok {
```
ngjaying, line 97: will it happen? report error?

ngjaying (approved):

Issue #2981, opened 2024-07-04
Bad variable reference causes a NPE


zhang943:
**Environment**:

- eKuiper version : 1.13.4 

**What happened and what you expected to happen**:
In line 258, a bad variable "err" is referenced, it should actually be "err1". The same error is also found in the functions sourceConfKeyHandler and connectionConfKeyHandler.

https://github.com/lf-edge/ekuiper/blob/add08deaff414eec7b83f5bbb43587dc061685a4/internal/server/meta_init.go#L245-L267

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

ngjaying:
@zhang943 Thanks for reporting this. Since you have already locate the problem in the source code, is it possible to create a PR?

PR #2984, opened 2024-07-04
fix: fix the bad variable reference

zhang943:
fix https://github.com/lf-edge/ekuiper/issues/2981


ngjaying:
@zhang943 Please check the comment. What do you think? Thanks.

PR #2997, opened 2024-07-09
fix(docs): fix guide error about join func

zhenjun115:


ngjaying:
@zhenjun115 Sorry, I was wrong. Actually, eKuiper rule supports two different format to define business logic: SQL and graph. The stream and scan table join is supported in SQL rule, but not in the graph rule yet. So the doc is correct.

zhenjun115:
👌

PR #3000, opened 2024-07-10
feat: support rule trial

Yisaer:


internal/topo/node/sink_node.go
```diff
@@ -66,12 +66,13 @@ func newSinkNode(ctx api.StreamContext, name string, rOpt def.RuleOption, eoflim
 
 func (s *SinkNode) Exec(ctx api.StreamContext, errCh chan<- error) {
 	s.prepareExec(ctx, errCh, "sink")
+	err := s.sink.Connect(ctx)
```
ngjaying, line 69: why move this out?

ngjaying (approved):

PR #3011, opened 2024-07-15
feat: support load connections env

Yisaer:


ngjaying (commented):
Is it possible to unify all env settings including both normal conf and connection conf together?

ngjaying (approved):

PR #3015, opened 2024-07-15
fix: remove create connection from critical area

Yisaer:


ngjaying (commented):
We may want to distinguish between connect once (waitOnce()?) and auto reconnect (wait()?). In source, connect need to wait until connected. In sink, connect only need to wait once.

extensions/impl/sql/sink.go
```diff
@@ -106,24 +107,28 @@ func (s *SQLSinkConnector) Provision(ctx api.StreamContext, configs map[string]a
...
 
 func (s *SQLSinkConnector) Close(ctx api.StreamContext) error {
 	ctx.GetLogger().Infof("Closing sql sink connector url:%v", s.config.DBUrl)
-	if s.conn != nil {
-		id := s.config.DBUrl
-		connection.DetachConnection(ctx, id, s.props)
-		s.conn.DetachSub(ctx, s.props)
+	id := s.config.DBUrl
+	connection.DetachConnection(ctx, id, s.props)
+	conn, _ := s.cw.Wait()
```
ngjaying, line 129: Why not use s.conn?

internal/io/neuron/sink.go
```diff
@@ -81,7 +83,14 @@ func (s *sink) Ping(ctx api.StreamContext, props map[string]interface{}) error {
 }
 
 func (s *sink) Connect(ctx api.StreamContext) error {
-	cli, err := connect(ctx, s.cc.Url, s.props)
+	ctx.GetLogger().Infof("Connecting to neuron")
```
ngjaying, line 86: Why not using the connect function?

internal/server/rest.go
```diff
@@ -647,6 +647,9 @@ func ruleHandler(w http.ResponseWriter, r *http.Request) {
 			handleError(w, err, "Delete rule error", logger)
 			return
 		}
+		// wait resource clear
+		time.Sleep(time.Second)
```
ngjaying, line 651: Sleep too long, why needing this?

ngjaying (approved):

PR #3031, opened 2024-07-21
fix: delete useless line in echo.go

578223592:


ngjaying (commented):
This is not useless. It is a common pattern in golang to check if a struct implements an interface in complile time.

578223592:
> This is not useless. It is a common pattern in golang to check if a struct implements an interface in complile time.

thanks.this is the first time that i listen the complile check tech beacase of my poor go skill,and i learn it.

PR #3033, opened 2024-07-22
feat: restore kafka

Yisaer:


extensions/impl/kafka/source.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+	readerConfig := kConf.GetReaderConfig()
+	conf.Log.Infof("topic: %s, brokers: %v", readerConfig.Topic, readerConfig.Brokers)
+	readerConfig.Dialer = &kafkago.Dialer{
+		Timeout:       10 * time.Second,
+		DualStack:     true,
+		TLS:           tlsConfig,
+		SASLMechanism: mechanism,
+	}
+	reader := kafkago.NewReader(readerConfig)
```
ngjaying, line 131: Move connection to connect func. Provision is supposed to be light-weight. It will be called in planner. Connect will only be called when the rule is run.

extensions/impl/kafka/sink.go
```diff
@@ -0,0 +1,245 @@
...
+		err := k.collect(ctx, data)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (k *KafkaSink) collect(ctx api.StreamContext, data map[string]any) error {
+	ds, err := json.Marshal(data)
```
ngjaying, line 154: This is not correct. The data can be other formats than JSON.

ngjaying (changes requested):

PR #3041, opened 2024-07-23
feat: restore kafka

Yisaer:


extensions/sinks/kafka/kafka.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kafka
+
+import (
+	"github.com/lf-edge/ekuiper/contract/v2/api"
+	"github.com/lf-edge/ekuiper/v2/extensions/impl/kafka"
+)
+
```
ngjaying, line 21: Sink json file is missing

extensions/impl/kafka/source.go
```diff
@@ -0,0 +1,274 @@
...
+	k.reader = reader
+	err := k.reader.SetOffset(kafkago.LastOffset)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (k *KafkaSource) Subscribe(ctx api.StreamContext, ingest api.BytesIngest, ingestError api.ErrorIngest) error {
+	go func() {
```
ngjaying, line 156: No need to start a new go routine, subscribe is already called asyncly

ngjaying (approved):

PR #3042, opened 2024-07-23
feat: restore import export plugins

Yisaer:


ngjaying (approved):
The v1 import/export also need to restore plugin support

PR #3063, opened 2024-07-28
feat: prometheus view rule cpu usage

Rookiecom:
By introducing the feature of observing cpu usage in a specified time window provided by the cpuprofile library, we can now observe the cpu usage of each rule of ekuiper through prometheus.

To use this feature, you need to set prometheus and enableResourceProfiling in the configuration file to true.

Then, run ekuiper and use [test tool](https://github.com/Yisaer/kuiper-nexmark-test) as shown below:

image

You can observe rule cpu usge in prometheus as shown below:

image

### DIFF
--- a/internal/server/rule_init.go
+++ b/internal/server/rule_init.go
@@ -229,23 +229,54 @@ func scheduleCronRule(now time.Time, options *def.RuleOption) (bool, error) {
 	return false, nil
 }
 
-func startCPUProfiling(ctx context.Context) error {
-	if err := cpuprofile.StartProfilerAndAggregater(ctx, time.Duration(1000)*time.Millisecond); err != nil {
+type Profiler interface {
+	StartCPUProfiler(context.Context, time.Duration) error
+	EnableWindowAggregator(int)
+	GetWindowData() cpuprofile.DataSetAggregateMap
+}
+
+type ekuiperProfile struct{}
+
+func (e *ekuiperProfile) StartCPUProfiler(ctx context.Context, t time.Duration) error {
+	return cpuprofile.StartCPUProfiler(ctx, t)
+}
+
+func (e *ekuiperProfile) EnableWindowAggregator(window int) {
+	cpuprofile.EnableWindowAggregator(window)
+}
+
+func (e *ekuiperProfile) GetWindowData() cpuprofile.DataSetAggregateMap {
+	return cpuprofile.GetWindowData()
+}
+
+func StartCPUProfiling(ctx context.Context, cpuProfile Profiler) error {
+	cpuProfile.EnableWindowAggregator(30)
+	if err := cpuProfile.StartCPUProfiler(ctx, 1000*time.Millisecond); err != nil {
 		return err
 	}
-	receiveChan := make(chan *cpuprofile.DataSetAggregate, 1024)
-	cpuprofile.RegisterTag("rule", receiveChan)
 	go func(ctx context.Context) {
+		ticker := time.NewTicker(15 * time.Second)
+		defer ticker.Stop()
 		for {
 			select {
 			case <-ctx.Done():
 				return
-			case data := <-receiveChan:
-				// TODO: support query in future
-				conf.Log.Debugf("cpu profile data: %v", data)
+			case <-ticker.C:
+				data := cpuProfile.GetWindowData()
+				if data == nil {
+					continue
+				}
+				ruleUsage, ok := data["rule"]
+				if !ok {
+					continue
+				}
+				for labelValue, t := range ruleUsage.Stats {
+					promMetrics.SetRuleCPUUsageGauge(labelValue, t)
+				}
 			}
 		}
 	}(ctx)
+
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


Rookiecom:
> Build packages / build (1.22.1, linux/amd64, debian) (pull_request) Failing

I call the newly added function at internal/server/promMetrics/prom_metrcis.go at intern/server.go, but prom_metrice.go can't be compiled when the tag of go build is core.So what should I do?

Rookiecom:
> > Build packages / build (1.22.1, linux/amd64, debian) (pull_request) Failing
> 
> I call the newly added function at internal/server/promMetrics/prom_metrcis.go at intern/server.go, but prom_metrice.go can't be compiled when the tag of go build is core.So what should I do?

just add func SetRuleCPUUsageGauge to core_metrics.go

internal/server/rule_init.go
```diff
@@ -230,22 +231,34 @@ func scheduleCronRule(now time.Time, options *def.RuleOption) (bool, error) {
...
 func startCPUProfiling(ctx context.Context) error {
-	if err := cpuprofile.StartProfilerAndAggregater(ctx, time.Duration(1000)*time.Millisecond); err != nil {
+	cpuprofile.EnableWindowAggregator(30)
+	if err := cpuprofile.StartCPUProfiler(ctx, time.Duration(1000)*time.Millisecond); err != nil {
 		return err
 	}
-	receiveChan := make(chan *cpuprofile.DataSetAggregate, 1024)
-	cpuprofile.RegisterTag("rule", receiveChan)
 	go func(ctx context.Context) {
+		ticker := time.NewTicker(time.Duration(15) * time.Second)
```
ngjaying, line 239: Use 15 * time.Second

internal/server/rule_init.go
```diff
@@ -230,22 +231,34 @@ func scheduleCronRule(now time.Time, options *def.RuleOption) (bool, error) {
...
+				data := cpuprofile.GetWindowData()
+				if data == nil {
+					continue
+				}
+				ruleUsage, ok := data["rule"]
+				if !ok {
+					continue
+				}
+				for labelValue, t := range ruleUsage.Stats {
+					fmt.Println(labelValue, t)
```
ngjaying, line 255: Remove this? At least don't use fmt.Printxx

ngjaying (changes requested):
The coverage is not enough? It looks like UT is not easy to add. Could you try to add a FVT test case to cover the changes? Take a look at fvt/server_test.go as an example.

internal/server/rule_init.go
Rookiecom: sorry, i forgot to remove it(for debug)🥺

Rookiecom:
> The coverage is not enough? It looks like UT is not easy to add. Could you try to add a FVT test case to cover the changes? Take a look at fvt/server_test.go as an example.

Sure, I will take care of this shortly.

ngjaying:
For this FVT error, in .github/workflows/run_test_case.yaml, try to add MQTT service.

ngjaying:
@Rookiecom This is not performance test. In the UT, we only need to cover the functionality, like assert the return body format. You don't even need to feed data if you can get a CPU usage result.

Rookiecom:
@ngjaying I'm really doing a functional test. Because cpuprofile library only parse the profileData with label, I can only get  an empty map if I don't feed data.If only need to detect whether a non-nil map is returned, it is not necessary to feed data?

ngjaying:
> @ngjaying I'm really doing a functional test. Because cpuprofile library only parse the profileData with label, I can only get an empty map if I don't feed data.If only need to detect whether a non-nil map is returned, it is not necessary to feed data?

Yeah, that's a valid scenario. We can cover the empty map scenario at the beginning. It would already cover most of the code, correct?

Rookiecom:
@ngjaying There is nothing I can do about that goroutine which writes metrics regularly

ngjaying:
> @ngjaying There is nothing I can do about that goroutine which writes metrics regularly

OK，thanks. You can take a look at the mock timer timex package. Or we can merge this one, and I'll try later to cover this.

Rookiecom:
> You can take a look at the mock timer timex package

yeah, I have had a look at mock timex package, but I don't know how to mock ticker, or you do it(I learn ?)

internal/server/rule_init.go
```diff
@@ -229,23 +229,54 @@ func scheduleCronRule(now time.Time, options *def.RuleOption) (bool, error) {
...
+
+func StartCPUProfiling(ctx context.Context, cpuProfile Profiler) error {
+	cpuProfile.EnableWindowAggregator(30)
+	if err := cpuProfile.StartCPUProfiler(ctx, 1000*time.Millisecond); err != nil {
 		return err
 	}
-	receiveChan := make(chan *cpuprofile.DataSetAggregate, 1024)
-	cpuprofile.RegisterTag("rule", receiveChan)
 	go func(ctx context.Context) {
+		ticker := time.NewTicker(15 * time.Second)
```
ngjaying, line 258: Use timex.GetTicker here. Then in the test, you can use timex.Add() to mock time proceeding.

Rookiecom: I have two question:

1. will timex affect program's performance?
2. although we can use timex.Add() to mock time proceeding, how to test code other than ticker？

ngjaying: 1. No.   2. Once time proceed, the code inside ticker will be executed.

ngjaying:
@Rookiecom I'll merge this for now. We can add more tests and docs in other PRs.

PR #3091, opened 2024-08-09
fix: fix routine exit

Yisaer:


internal/topo/node/concurrent.go
```diff
@@ -77,7 +91,10 @@ func merge(ctx api.StreamContext, node *defaultSinkNode, sendInterval time.Durat
 	}
 }
 
-func distribute(ctx api.StreamContext, node *defaultSinkNode, numWorkers int, workerChans []chan any) {
+func distribute(ctx api.StreamContext, node *defaultSinkNode, numWorkers int, workerChans []chan any, cancelWorker context.CancelFunc) {
+	defer func() {
+		cancelWorker()
```
ngjaying, line 96: It seems this cannot guarantee distribute exit before worker. Because worker all listens to the original ctx.Done()

PR #3094, opened 2024-08-12
feat: support tracing framework

Yisaer:


contract/api/ctx.go
```diff
@@ -49,6 +49,8 @@ type StreamContext interface {
 	WithMeta(ruleId string, opId string, store Store) StreamContext
 	WithInstance(instanceId int) StreamContext
 	WithCancel() (StreamContext, context.CancelFunc)
+	SetRuleTracer(enabled bool)
```
ngjaying, line 52: Tracing can be changed during runtime? How about name it EnableTracer?

contract/api/ctx.go
```diff
@@ -49,6 +49,8 @@ type StreamContext interface {
...
+	IsRuleTraceEnabled() bool
```
ngjaying, line 53: Name it IsTraceEnabled?

ngjaying (approved):

PR #3100, opened 2024-08-15
fix: use prepared stmt in sql

Yisaer:


ngjaying (commented):
Do we need to change the SQL plugin as well?

ngjaying (approved):

PR #3102, opened 2024-08-15
fix: fix scheduleSignal

Yisaer:


ngjaying:
No need to handle state transit in schedule, it is all controlled by rule state itself.

PR #3130, opened 2024-08-22
fix(sql): backport CVE-2024-43406

Crispy-fried-chicken:
Here is a vulnerability which is fixed in the master branch (https://github.com/lf-edge/ekuiper/commit/1a9c745649438feaac357d282959687012b65503) but is not fixed in the branch of master-1.13, maybe it should be backported?

### DIFF
--- a/internal/pkg/store/sql/sqlKv.go
+++ b/internal/pkg/store/sql/sqlKv.go
@@ -89,10 +89,15 @@ func (kv *sqlKvStore) Set(key string, value interface{}) error {
 func (kv *sqlKvStore) Get(key string, value interface{}) (bool, error) {
 	result := false
 	err := kv.database.Apply(func(db *sql.DB) error {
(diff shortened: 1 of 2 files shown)


ngjaying:
We only maintain the latest version (1.14.x now) as usual. Recommend to upgrade instead of stay in 1.13. Any reason to port to 1.13.x?

ngjaying:
If you're still using 1.13.x and willing to maitain issues like this, we are more than happy to merge and publish new release. Thank you. Please check the DCO details link to sign off.

ngjaying:
> I see here is an error for build-for-android, what does it mean? I don't modify the code it mentioned, how can I fix it?

It fails due to dependency version updates. Just ignore it now.

ngjaying:
@Crispy-fried-chicken  The build-for-android fail can be ignored now. However, the UT error should be fixed. Please kindly take a look.

ngjaying (approved):

PR #3144, opened 2024-08-27
fix: fix metrics for batch operator

Yisaer:


internal/topo/node/batch_op.go
```diff
@@ -150,6 +149,7 @@ func (b *BatchOp) send(ctx api.StreamContext) {
 		panic("shouldn't send message when empty")
 	})
 	b.handleTraceEmitTuple(ctx, b.buffer)
+	b.statManager.IncTotalRecordsOut()
```
ngjaying, line 152: better move after broadcast

ngjaying (approved):

Issue #3080, opened 2024-08-07
panic when restart a mqtt source rule


TKthink:
**Environment**:

- eKuiper version (e.g. `1.3.0`): 1.14.1、v2alpha.10
- Hardware configuration (e.g. `lscpu`): armv7 also try x86_64
- OS (e.g. `cat /etc/os-release`):custom linux 
- Others:

**What happened and what you expected to happen**:
panic when restart a mqtt source rule.
expected:DO NOT actively panic unless during the program initialization phase

**How to reproduce it (as minimally and precisely as possible)**:
make a mqtt source with clientid,then make only one rule using this mqtt source,and finally switch the rule repeatedly,after few times switching,it will panic

**Anything else we need to know?**:
panic info

panic: Failed to subscribe topic /lab/test not currently connected and ResumeSubs not set

goroutine 1189 [running]:
github.com/lf-edge/ekuiper/internal/io/mqtt.(*Connection).onConnect(0xc0006220c0, {0x2781a08, 0xc0000a8a08})
	C:/Users/xxx/Desktop/demo project/ekuiper/internal/io/mqtt/connection.go:66 +0x291
created by github.com/eclipse/paho%2emqtt%2egolang.(*client).startCommsWorkers in goroutine 966
	C:/Users/xxx/Desktop/demo project/ekuiper/vendor/github.com/eclipse/paho.mqtt.golang/client.go:617 +0xb85

TKthink:
https://github.com/lf-edge/ekuiper/blob/0cef7657f254b24ad32192f8692d05e77d06cc4c/internal/io/mqtt/connection_pool.go#L31
this code. Why get the connectionSelector value as clientid instead of getting the clientid as clientid? get the connectionSelector always get "",it makes connectionPool always have nothing

ngjaying:
> https://github.com/lf-edge/ekuiper/blob/0cef7657f254b24ad32192f8692d05e77d06cc4c/internal/io/mqtt/connection_pool.go#L31
> 
> 
> this code. Why get the connectionSelector value as clientid instead of getting the clientid as clientid? get the connectionSelector always get "",it makes connectionPool always have nothing

This is eKuiper level connection selector for the connection sharing between rules. It is different from mqtt client id.

TKthink:
> @TKthink Could you provide the rule spec and describe how to `swtich` the rule repeatedly?

here is the rule:
{
    "id": "rule_dc3e",
    "name": "",
    "triggered": true,
[...]

PR #3149, opened 2024-08-28
fix: fix mqtt client panic

Yisaer:


internal/io/mqtt/client/client.go
```diff
@@ -69,10 +69,7 @@ func (conn *Connection) onConnect(_ pahoMqtt.Client) {
 	conn.connected.Store(true)
 	conn.logger.Infof("The connection to mqtt broker is established")
```
ngjaying, line 70: Do not need to handle subscription now. Looks like we can remove conn.subscriptions and all codes that manipulate it.

ngjaying (changes requested):

internal/io/mqtt/client/client.go
```diff
@@ -68,20 +66,10 @@ func (conn *Connection) Publish(topic string, qos byte, retained bool, payload a
...
-	}
 }
 
 func (conn *Connection) onConnectLost(_ pahoMqtt.Client, err error) {
 	conn.connected.Store(false)
-	e := fmt.Errorf("The connection to mqtt broker is lost due to %v", err)
-	for _, info := range conn.subscriptions {
-		info.ErrHandler(e)
-	}
 	conn.logger.Infof("%v", err)
```
ngjaying, line 73: delete this line

ngjaying (changes requested):

ngjaying (approved):

ngjaying:
Closes #3080

PR #3163, opened 2024-09-03
feat: revise connection status

Yisaer:


ngjaying:
Close because in another refactor PR, this fix is already covered

PR #3167, opened 2024-09-03
feat: support trace rest api

Yisaer:


ngjaying:
Do not need api trace for now, so close.

ngjaying (approved):

PR #3168, opened 2024-09-03
fix: fix mqtt sink conf validate

Yisaer:


internal/io/mqtt/client/client.go
```diff
@@ -162,6 +162,10 @@ func handleToken(token pahoMqtt.Token) error {
 }
 
 func ValidateConfig(props map[string]any) (*ConnectionConfig, error) {
+	_, ok := props["connectionSelector"].(string)
```
ngjaying, line 165: How about other types of client？

ngjaying (approved):

PR #3176, opened 2024-09-06
feat: support local span storage

Yisaer:


internal/pkg/store/setup.go
```diff
@@ -68,5 +71,13 @@ func Setup(config definition.Config) error {
 		return err
 	}
 	extStateStores = s
-	return nil
+	db, err := sqldb.BuildSqliteStore(config, "trace.db")
```
ngjaying, line 74: Only create when enable local cache?

internal/pkg/store/setup.go
```diff
@@ -68,5 +71,13 @@ func Setup(config definition.Config) error {
...
+	if err != nil {
+		return err
+	}
+	TraceStores = db
+	return TraceStores.Apply(func(db *sql.DB) error {
```
ngjaying, line 79: What happen if db already exists?

ngjaying (approved):

PR #3193, opened 2024-09-12
feat: revise rule trace api

Yisaer:


internal/server/rest_test.go
```diff
@@ -89,6 +89,8 @@ func (suite *RestTestSuite) SetupTest() {
 	r.HandleFunc("/rules/{name}/topo", getTopoRuleHandler).Methods(http.MethodGet)
 	r.HandleFunc("/rules/{name}/reset_state", ruleStateHandler).Methods(http.MethodPut)
 	r.HandleFunc("/rules/{name}/explain", explainRuleHandler).Methods(http.MethodGet)
+	r.HandleFunc("/rules/{name}/trace/start", enableRuleTraceHandler).Methods(http.MethodPost)
```
ngjaying, line 92: Is it better to move this to fvt directly. Don't need to sync the duplicate routing code in the test

internal/topo/context/default.go
```diff
@@ -39,12 +39,12 @@ const (
...
-	instanceId       int
-	ctx              context.Context
-	err              error
-	enableRuleTracer bool
+	ruleId         string
+	opId           string
+	instanceId     int
+	ctx            context.Context
+	err            error
+	isTraceEnabled *isRuleTraceEnabledWrapper
```
ngjaying, line 47: Why need to use a wrapper? Just want to add the lock. Is it possible to use atomic.bool to keep it simple？

ngjaying (approved):

PR #3196, opened 2024-09-13
feat: support query rule trace

Yisaer:


internal/server/rule_manager.go
```diff
@@ -277,10 +277,18 @@ func (rr *RuleRegistry) GetAllRulesWithStatus() ([]map[string]any, error) {
 		} else {
 			str = rule.StateName[s]
 		}
+		trace := false
```
ngjaying, line 280: Is it better to get trace state together with other state in line 274?

ngjaying (approved):

PR #3221, opened 2024-09-24
fix: warning when config not in connection.yaml

Yisaer:


internal/conf/yaml_config_ops.go
```diff
@@ -501,7 +501,7 @@ func NewConfigOperatorFromConnectionStorage(pluginName string) (ConfigOperator,
 			return nil, fmt.Errorf("file content is not right: %v", plgCnfs)
 		}
 	} else {
-		return nil, fmt.Errorf("not find the target connection type: %s", c.pluginName)
+		Log.Warnf("connection type %s have no config in connection.yaml", pluginName)
```
ngjaying, line 504: Do we need to print anything here? Just remove else?

ngjaying (approved):

PR #3253, opened 2024-09-27
fix: fix driver import

Yisaer:


extensions/impl/sql/client/dburl.go
```diff
@@ -40,7 +41,7 @@ func ParseDBUrl(urlstr string) (string, string, error) {
 func ParseDriver(url string) (string, error) {
 	u, err := dburl.Parse(url)
 	if err != nil {
-		return "", fmt.Errorf("parse driver err:%v", err)
+		return "", fmt.Errorf("parse driver err:%v, known drivers: %v", err, driver.KnownBuildTags())
```
ngjaying, line 44: Print out known drivers seems confusing. They are not the current built-in drivers

ngjaying (approved):

PR #3271, opened 2024-10-03
build: Dockerfile-alpine-python and Makefile build

vkaramanis:
**Motivation:**
When running Edgex in secure mode it uses securitty-bootstrapper which has been compiled in Alpine Linux. 
Thus, using any other docker image than ekuiper-alpine fails because it cannot run the aforementioned binary.

**Suggested solution:**
Make a eKuiper python Alpine based docker image to be able to use python based plugins

**Tests:**
Tested and running successfully using ekuiper v1.14.4

### DIFF
--- a/deploy/docker/Dockerfile-alpine-python
+++ b/deploy/docker/Dockerfile-alpine-python
@@ -21,7 +21,7 @@ WORKDIR /go/kuiper
 
 RUN make build_with_edgex
 
-FROM python:3.12.7-alpine3.20
+FROM python:3.12-alpine
 
 # Set environment vars
 ENV MAINTAINER="emqx.io" \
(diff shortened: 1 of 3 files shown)


deploy/docker/Dockerfile-alpine-python
```diff
@@ -0,0 +1,54 @@
...
+ARG GO_VERSION=1.23.1
+FROM ghcr.io/lf-edge/ekuiper/base:$GO_VERSION-alpine AS builder
+
+COPY . /go/kuiper
+
+WORKDIR /go/kuiper
+
+RUN make build_with_edgex
+
+FROM python:3.12.7-alpine3.20
```
Rory-Z, line 24:
```suggestion
FROM python:3.12-alpine
```
I think it's better than `python:3.12.7-alpine3.20`, because we can use the latest patch version.

Rory-Z:
And please update `build-docker-images` jobs in `.github/workflows/build_packages.yaml`

ngjaying:
@vkaramanis Thanks. Please also sign-off your commit, following this link https://github.com/lf-edge/ekuiper/pull/3271/checks?check_run_id=31035835843

Rory-Z (approved):

PR #3284, opened 2024-10-11
feat: support trace strategy

Yisaer:
support trace strategy `head`/`always`

1. `head` will check whether the source request has trace parent if enabled
2. `always`will always trace if enabled


ngjaying (changes requested):
Need docs for strategy setting.

ngjaying (approved):

PR #3308, opened 2024-10-18
fix: fix import connection configuration

Yisaer:


internal/meta/yamlConfigMeta.go
```diff
@@ -768,6 +769,9 @@ func LoadConfigurations(configSets YamlConfigurationSet) YamlConfigurationSet {
 			continue
 		}
 	}
+	if err := connection.ReloadNamedConnection(); err != nil {
```
ngjaying, line 772: How does this deal with the already loaded connections?

ngjaying (approved):

PR #3322, opened 2024-10-23
feat: support inc agg window

Yisaer:


internal/binder/function/funcs_agg.go
```diff
@@ -379,3 +379,26 @@ func registerAggFunc() {
 		check: returnNilIfHasAnyNil,
 	}
 }
+
+func registerIncAggFunc() {
+	builtins["inc_count"] = builtinFunc{
+		fType: ast.FuncTypeScalar,
+		exec: func(ctx api.FunctionContext, args []interface{}) (interface{}, bool) {
+			key := fmt.Sprintf("%v", ctx.GetFuncId())
```
ngjaying, line 387: Why do Sprintf, seems unnecessary

internal/topo/node/window_inc_agg_op.go
```diff
@@ -0,0 +1,183 @@
...
+					co.setIncAggWindow(ctx)
+				}
+				name := "dim_"
+				ve := &xsql.ValuerEval{Valuer: xsql.MultiValuer(row, fv, &xsql.WildcardValuer{Data: row})}
+				for _, d := range co.Dimensions {
+					r := ve.Eval(d.Expr)
+					if _, ok := r.(error); ok {
+						continue
+					} else {
+						name += fmt.Sprintf("%v,", r)
```
ngjaying, line 122: Think about faster string concatenation because this is in the hot path

internal/topo/node/window_inc_agg_op.go
```diff
@@ -0,0 +1,183 @@
...
+
+func (co *CountWindowIncAggOp) setIncAggWindow(ctx api.StreamContext) {
+	co.currWindow = &IncAggWindow{
+		DimensionsIncAggRange: make(map[string]*IncAggRange),
+	}
+}
+
+func (co *CountWindowIncAggOp) newIncAggRange(ctx api.StreamContext) *IncAggRange {
+	fstore, _ := state.CreateStore("incAggWindow", 0)
+	fctx := topoContext.Background().WithMeta(ctx.GetRuleId(), ctx.GetOpId(), fstore)
```
ngjaying, line 143: Why use backgroud context? Any reason to not link to the passed in ctx?

ngjaying (changes requested):

ngjaying (approved):

PR #3378, opened 2024-11-14
feat: validate confkey

Yisaer:


internal/server/meta_init.go
```diff
@@ -290,6 +291,10 @@ func connectionConfKeyHandler(w http.ResponseWriter, r *http.Request) {
 	case http.MethodDelete:
 		err = meta.DelConnectionConfKey(pluginName, confKey, language)
 	case http.MethodPut:
+		if err := validate.ValidateID(confKey); err != nil {
```
ngjaying, line 294: Is it possible that the confKey contain invalid character like /  e.g mqtt/key1?

PR #3392, opened 2024-11-21
feat: support merge tokens into body

Yisaer:


internal/io/http/httppull_source.go
```diff
@@ -75,7 +76,8 @@ func doPull(ctx api.StreamContext, c *ClientConf, lastMD5 string) ([]map[string]
 	if err != nil {
 		return nil, "", err
 	}
-	resp, err := httpx.Send(ctx.GetLogger(), c.client, c.config.BodyType, c.config.Method, c.config.Url, headers, []byte(c.config.Body))
+	body := writeTokenIntoBody(c.tokens, []byte(c.config.Body))
```
ngjaying, line 79: This is not to hard coded body. Just to support using template in body like using that in url, bodytype etc.

ngjaying (changes requested):

ngjaying (approved):

PR #3443, opened 2024-12-13
feat: support io metrics

Yisaer:


ngjaying:
Looks like these are common metrics. No need to add them in the specific io node. So close for now.

extensions/impl/kafka/sink.go
```diff
@@ -19,18 +19,22 @@ import (
...
 	"github.com/segmentio/kafka-go/sasl"
 
 	"github.com/lf-edge/ekuiper/v2/internal/pkg/util"
+	"github.com/lf-edge/ekuiper/v2/metrics"
 	"github.com/lf-edge/ekuiper/v2/pkg/cast"
 	"github.com/lf-edge/ekuiper/v2/pkg/cert"
 )
 
 type KafkaSink struct {
+	ruleID         string
```
ngjaying, line 36: Is it possible to get it from context directly. Seems no need to save in the struct.

ngjaying (changes requested):

ngjaying (approved):

PR #3445, opened 2024-12-16
fix: converter lose precision

Yisaer:
goos: darwin
goarch: amd64
pkg: github.com/lf-edge/ekuiper/v2/internal/converter/json
cpu: Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
BenchmarkFloatParse
BenchmarkFloatParse-4   	 1000000	      1044 ns/op
PASS

goos: darwin
goarch: amd64
pkg: github.com/lf-edge/ekuiper/v2/internal/converter/json
cpu: Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
BenchmarkIntParse
BenchmarkIntParse-4   	 1763830	       661.6 ns/op
PASS

### DIFF
--- a/internal/converter/json/converter.go
+++ b/internal/converter/json/converter.go
@@ -17,6 +17,7 @@ package json
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/lf-edge/ekuiper/contract/v2/api"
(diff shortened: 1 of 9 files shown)


internal/converter/json/converter.go
```diff
@@ -454,3 +447,22 @@ func getType(t *ast.JsonStreamField) string {
 		return t.Type
 	}
 }
+
+func extractNumber(v *fastjson.Value) (any, error) {
+	if !isFloat64(v.String()) && !conf.IsTesting {
```
ngjaying, line 452: Will this slow down the decode a lot?

ngjaying (changes requested):
This is just

ngjaying (approved):

PR #3451, opened 2024-12-18
fix: fix sync cache context

Yisaer:


internal/topo/node/cache_op.go
```diff
@@ -105,10 +109,13 @@ func (s *CacheOp) Exec(ctx api.StreamContext, errCh chan<- error) {
...
 					s.span = nil
 					s.onProcessEnd(ctx)
-					l := int64(len(s.input) + s.cache.CacheLength)
+					a := int64(len(s.input))
+					b := s.cache.CacheLength
+					l := a + int64(b)
 					if s.currItem != nil {
 						l += 1
 					}
+					conf.Log.Printf("cache op buffer length %v, chan %v, cacheLength %v", l, a, b)
```
ngjaying, line 118: use ctx.GetLogger().Infof ?

ngjaying (changes requested):

ngjaying (approved):

PR #3464, opened 2024-12-24
feat: support config portable timeout

Yisaer:


etc/kuiper.yaml
```diff
@@ -140,6 +140,8 @@ portable:
   pythonBin: python
   # control init timeout. If the init time is longer than this value, the plugin will be terminated.
   initTimeout: 60s
```
ngjaying, line 142: Suggest to keep the previous default value

ngjaying (changes requested):

ngjaying (approved):

PR #3474, opened 2024-12-25
feat: bump go to 1.23.4

Yisaer:


ngjaying (changes requested):
Need to split to 2 PRs. The first one update build_base_image.yaml. After merge, trigger the new base image. Then push other changes as the 2nd PR.

PR #3477, opened 2024-12-25
feat: add alias reference calculate swtich

Yisaer:


internal/pkg/def/rule.go
```diff
@@ -47,6 +47,14 @@ type RuleOption struct {
 type PlanOptimizeStrategy struct {
 	EnableIncrementalWindow bool `json:"enableIncrementalWindow,omitempty" yaml:"enableIncrementalWindow,omitempty"`
 	EnableAliasPushdown     bool `json:"enableAliasPushdown,omitempty" yaml:"enableAliasPushdown,omitempty"`
+	EnableAliasRefCal       bool `json:"enableAliasRefCal,omitempty" yaml:"enableAliasRefCal,omitempty"`
```
ngjaying, line 50: Is it better to name it disableAliasRefCal to keep the default value false?

ngjaying (approved):

PR #3506, opened 2025-01-14
modify nil != nil bug

maosenlin:
internal/server/rpc_plugin.go
line: 42

code current:
// define according to the build tag
err = t.doRegister(pt, p)
if err != nil {
return err
}
if err != nil {
return fmt.Errorf("Create plugin error: %s", err)
} else {
*reply = fmt.Sprintf("Plugin %s is created.", p.GetName())
}

code modified:
err = t.doRegister(pt, p)
if err != nil {
return fmt.Errorf("Create plugin error: %s", err)
} else {
*reply = fmt.Sprintf("Plugin %s is created.", p.GetName())
}


ngjaying:
@maosenlin Thanks. Please follow the link in DCO check to pass it. Basically, you need to sign the commit.

PR #3617, opened 2025-03-13
feat(planner): rule option check

Yisaer:
Rule option disableBufferFullDiscard is not allowed when source is shared


internal/topo/planner/planner.go
```diff
@@ -344,6 +344,18 @@ func CreateLogicalPlan(stmt *ast.SelectStatement, opt *def.RuleOption, store kv.
 	return createLogicalPlan(stmt, opt, store)
 }
 
+func checkSharedSourceOption(streams []*streamInfo, opt *def.RuleOption) error {
```
ngjaying, line 347: Seems this function is not used?

ngjaying (approved):

PR #3637, opened 2025-03-25
feat: support connection metrics

Yisaer:


ngjaying (approved):

ngjaying (commented):
Periodic pull connection status may have overhead. Why not just update the connection status metrics in the connection status change handler?

PR #3651, opened 2025-04-02
fix: check kafka batch config

Yisaer:


extensions/impl/kafka/sink.go
```diff
@@ -161,6 +161,9 @@ func (k *KafkaSink) Provision(ctx api.StreamContext, configs map[string]any) err
 	if err != nil {
 		return err
 	}
+	if k.kc.BatchSize == 0 && k.kc.LingerInterval == 0 {
```
ngjaying, line 164: How about just report error?

Yisaer: `batchSize:0 ` and `lingerInterval:0` will make batch op can't send the msg, and the `batchSize:0` will be sent in the props by neuronex by default if it is not set .

ngjaying (approved):

PR #3671, opened 2025-04-15
feat: support eof tuple when rule stop

Yisaer:


internal/topo/node/source_node.go
```diff
@@ -138,6 +158,9 @@ func (m *SourceNode) traceStart(ctx api.StreamContext, meta map[string]any, tupl
 }
 
 func (m *SourceNode) ingestAnyTuple(ctx api.StreamContext, data any, meta map[string]any, ts time.Time) {
+	if m.recvFinNotify.Load() {
```
ngjaying, line 161: Just stop ingesting seems incorrect. The state will save til the last checkpoint barrior, but the source may already stopped ingest before receiving the barrior.

PR #3689, opened 2025-05-06
feat: support batch request

### LINKED ISSUE #2746: REST API to delete multiple/all rules
**What would you like to be added/modified**:
I propose the addition of a new endpoint that accepts a list of rule IDs to be deleted in a single API call. This would allow users to perform bulk deletions without having to make multiple individual requests, thereby reducing network overhead and improving system performance.

**Why is this needed**:
As a user of ekuiper, I often find the need to efficiently delete multiple rules through a REST API. Currently, the system only supports the deletion of individual rules through the API. To enhance the usability and efficiency of ekuiper, it would be beneficial to introduce a new REST API endpoint that allows for the simultaneous deletion of multiple rules.

Yisaer:
Closed #2746

### DIFF
--- a/docs/en_US/api/restapi/overview.md
+++ b/docs/en_US/api/restapi/overview.md
@@ -44,10 +44,9 @@ POST http://localhost:9081/batch/req
         "path": "/streams/demobatch"
     }
 ]
-```
 
-response
-```json
+Response
+
 [
     {
         "code": 201,
--- a/docs/zh_CN/api/restapi/overview.md
+++ b/docs/zh_CN/api/restapi/overview.md
@@ -42,10 +42,9 @@ POST http://localhost:9081/batch/req
         "path": "/streams/demobatch"
     }
 ]
-```
 
-response
-```json
+Response
+
 [
     {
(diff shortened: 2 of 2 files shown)


internal/server/rest_test.go
```diff
@@ -1004,3 +1006,29 @@ func (suite *RestTestSuite) TestWaitStopRule() {
 	require.True(suite.T(), end.Sub(now) >= 300*time.Millisecond)
 	waitAllRuleStop()
 }
+
```
ngjaying, line 1009: Suggest not to add new case to this file any more because it needs a lot of duplicate code to mock. Just add a test case in fvt

ngjaying (commented):
Please add the doc.

internal/server/batch_init.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2025 EMQ Technologies Co., Ltd.
```
ngjaying, line 1: This is not a init file, rename it or move the code into rest.go.

PR #3710, opened 2025-05-23
feat: add rule cpu metrics panel

Yisaer:


ngjaying (changes requested):
Suggest to merge the if condition

PR #3721, opened 2025-05-27
feat: support rule tags

Yisaer:


internal/server/rule_tag.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func rulesTagsHandler(w http.ResponseWriter, r *http.Request) {
+	tagsReq := &RuleTagRequest{Tags: []string{}}
+	if err := json.NewDecoder(r.Body).Decode(&tagsReq); err != nil {
+		handleError(w, err, "decode body error", logger)
+		return
+	}
+	res := make([]string, 0)
+	rss := registry.list()
```
ngjaying, line 159: registry only has running rules. The list should get from the DB.

ngjaying (approved):

PR #3727, opened 2025-06-03
feat: support eof barrier

Yisaer:


internal/topo/checkpoint/responder.go
```diff
@@ -54,6 +54,10 @@ func (re *ResponderExecutor) TriggerCheckpoint(checkpointId int64) error {
 	barrier := &Barrier{
 		CheckpointId: checkpointId,
 		OpId:         name,
+		IsEof:        isEof,
+	}
+	if st, ok := re.task.(SourceTask); ok {
```
ngjaying, line 59: Is this correct? Always notify close regardless of the isEof signal.

ngjaying:
Will create another PR later

PR #3755, opened 2025-06-27
feat: support median function

Yisaer:


ngjaying (changes requested):
Add more ut & doc as well

internal/binder/function/funcs_agg.go
```diff
@@ -25,6 +26,33 @@ import (
...
+				return int64(0), true
+			}
+			switch arg0[0].(type) {
+			case float64:
+				f64s, err := cast.ToFloat64Slice(arg0, cast.CONVERT_SAMEKIND, cast.IGNORE_NIL)
+				if err != nil {
+					return err, false
+				}
+				return median(f64s), true
+			case int64:
```
ngjaying, line 43: Must cover int type as well

ngjaying (changes requested):

ngjaying (approved):

PR #3757, opened 2025-06-30
feat: support state window

Yisaer:


abdelhameedhamdy:
I am trying to create a rule using REST service, with statewindow()
I got the following response:

```
{
  "error": 2201,
  "message": "v1 window op didn't support state window, use v2 window"
}
```
knowing that the image used is :
 `lfedge/ekuiper:v2.3.0-beta.2`

PR #3767, opened 2025-07-09
fix(kv): prepare ts query

odaysec:
https://github.com/lf-edge/ekuiper/blob/163809ee5191827223759fd820883a361ba3d274/internal/pkg/store/sql/sqlTs.go#L163-L163

fix the issue, the SQL query should be constructed using parameterized queries or prepared statements. However, since SQL drivers typically do not allow parameterized table names, the `table` parameter must be validated explicitly to ensure it contains only safe characters (e.g., alphanumeric characters and underscores). This validation should be implemented in the `isValidTableName` function. additionally, the `getLast` function should avoid constructing the query string dynamically. Instead, it should use a safe and validated table name.

If a database query (such as an SQL or NoSQL query) is built from user-provided data without sufficient sanitization, a malicious user may be able to run commands that exfiltrate, tamper with, or destroy data stored in the database.

### DIFF
--- a/internal/pkg/store/sql/sqlTs.go
+++ b/internal/pkg/store/sql/sqlTs.go
@@ -39,12 +39,12 @@ func createSqlTs(database Database, table string) (*ts, error) {
 		last:     getLast(database, table),
 	}
 	err := store.database.Apply(func(db *sql.DB) error {
-		query := fmt.Sprintf("CREATE TABLE IF NOT EXISTS '%s'('key' INTEGER PRIMARY KEY, 'val' BLOB);", table)
+		query := fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s ('key' INTEGER PRIMARY KEY, 'val' BLOB);", table)
 		stmt, err := db.Prepare(query)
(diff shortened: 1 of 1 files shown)


odaysec:
Signed-off-by: Zeroday BYTE <github@zerodaysec.org>

ngjaying:
@odaysec Thanks. The sign off is not successful. Please try to follow this link https://github.com/lf-edge/ekuiper/pull/3767/checks?check_run_id=45605925176 to sign off

ngjaying:
@odaysec The change looks good. Do you encounter any problems to sign off?

ngjaying:
@odaysec I think you'll need to add the signoff to your commit instead of a comment in the PR. Thanks.

ngjaying:
> Signed-off-by: Zeroday BYTE [github@zerodaysec.org](mailto:github@zerodaysec.org)

I've added a sign-off to this commit and listed myself as a co-author. Please let me know if you think this is inappropriate.

Yisaer (approved):

PR #3785, opened 2025-07-17
fix gob not register map[string]time.Time{} bug

lidihao:
fix gob not register map[string]time.Time{} bug , when use event time.


ngjaying:
@lidihao Please follow the DCO check link to do the sign off. Thanks.

ngjaying:
@lidihao you'll need to set the merge target to master and update the sign-off. Could you please help to create another PR? Appreciate your help!

PR #3795, opened 2025-07-22
fix(checkpoint): gob register map[string]time.time

### LINKED ISSUE #3780: gob: type not registered for interface: map[string]time.Time
time="2025-07-14T09:29:30Z" level=info msg="Cannot save checkpoint 1752485370612 due to storage error: save checkpoint err: gob: type not registered for interface: map[string]time.Time" file="checkpoint/coordinator.go:300" rule=msgPushAckCntAndByteCntRule

When using event time and setting the stream's QoS to 1, this error will occur.

ngjaying:
Closes #3780

### DIFF
--- a/internal/topo/node/window_op.go
+++ b/internal/topo/node/window_op.go
@@ -88,6 +88,7 @@ const (
 func init() {
 	gob.Register([]xsql.EventRow{})
 	gob.Register([]map[string]interface{}{})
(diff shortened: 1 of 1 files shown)


ngjaying:
@lidihao Create this based on your commit and make you the co-author. Thanks.

PR #3803, opened 2025-07-25
feat: support eval aggfunc in where condition

Yisaer:


internal/topo/operator/agg_func_operator_test.go
```diff
@@ -0,0 +1,121 @@
...
+		switch aggFunc := node.(type) {
+		case *ast.Call:
+			if aggFunc.FuncType == ast.FuncTypeAgg {
+				newAggFunc := &ast.Call{
+					Name:     aggFunc.Name,
+					FuncType: aggFunc.FuncType,
+					Args:     aggFunc.Args,
+					FuncId:   aggFunc.FuncId,
+				}
+				name := fmt.Sprintf("agg_ref_%v", index)
```
ngjaying, line 99: How abou $$agg_ref to avoid possibiit of conflict? And may be extract the constant here.

PR #3814, opened 2025-08-01
fix: slice memory allocation with excessive size value

ptrgits:
https://github.com/lf-edge/ekuiper/blob/efbad99264b77152582253290cea7734d65095b9/internal/topo/node/dedup_trigger_op.go#L53-L53

https://github.com/lf-edge/ekuiper/blob/efbad99264b77152582253290cea7734d65095b9/internal/topo/node/node.go#L212-L212

Using untrusted input to allocate slices with the built-in `make` function could lead to excessive memory allocation and potentially cause the program to crash due to running out of memory. This vulnerability could be exploited to perform a denial-of-service attack by consuming all available server resources.

fix this issue, we need to ensure that the value of `options.BufferLength` used in the `make(chan interface{}, options.BufferLength)` call is within a safe, predefined range. The best way to do this is to define a reasonable maximum buffer length constant (e.g., `maxBufferLength := 1024`), and clamp `options.BufferLength` to this maximum (and a minimum of 1) before using it in the `make` call. This can be done directly in the `NewDedupTriggerNode` constructor, as this is where the allocation occurs. No changes to the function signature are required, and this approach preserves existing functionality while preventing excessive memory allocation.

### DIFF
--- a/internal/topo/node/dedup_trigger_op.go
+++ b/internal/topo/node/dedup_trigger_op.go
@@ -48,9 +48,16 @@ func NewDedupTriggerNode(name string, options *def.RuleOption, aliasName string,
 	if aliasName != "" {
 		aname = aliasName
 	}
+	const maxBufferLength = 1024
+	bufferLength := options.BufferLength
+	if bufferLength < 1 {
+		bufferLength = 1
+	} else if bufferLength > maxBufferLength {
+		bufferLength = maxBufferLength
+	}
 	return &DedupTriggerNode{
 		defaultSinkNode: &defaultSinkNode{
-			input: make(chan interface{}, options.BufferLength),
+			input: make(chan interface{}, bufferLength),
 			defaultNode: &defaultNode{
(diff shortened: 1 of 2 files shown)


ngjaying:
@ptrgits Thanks for bring up this. Could you please sign off by following the DCO check link: https://github.com/lf-edge/ekuiper/pull/3814/checks?check_run_id=47183593727

kreeksec:
Signed-off-by: pTr <itsecbadges.andri@gmail.com>

ngjaying:
> Signed-off-by: pTr [itsecbadges.andri@gmail.com](mailto:itsecbadges.andri@gmail.com)

Sign off in the PR comment does not work. Check this link https://github.com/lf-edge/ekuiper/pull/3814/checks?check_run_id=47183593727

ngjaying:
@ptrgits  Thanks for the contribution. But there are still a lot of fail checks. Are you able to fix them recently? Feel free to add me if you need help.

PR #3864, opened 2025-09-15
fix: fix mqtt session config

Yisaer:


internal/io/mqtt/v4client/client.go
```diff
@@ -135,8 +135,8 @@ func (c *Client) Unsubscribe(_ api.StreamContext, topic string) error {
 	return nil
 }
 
-func (c *Client) Disconnect(_ api.StreamContext) {
-	c.cli.Disconnect(1)
+func (c *Client) Disconnect(ctx api.StreamContext) {
+	c.cli.Disconnect(1000)
```
ngjaying, line 139: Why change this?

PR #3910, opened 2025-10-29
feat(io): add rollingSize option to file sink

e7217:
Hello,

## Description

Adds support for file rolling based on file size in the file sink. This allows users to rotate files when they reach a specified size threshold, in addition to the existing time-based and count-based rolling options.

## Motivation

Currently, the file sink only supports rolling by:
- Time interval (`rollingInterval`)
- Number of records (`rollingCount`)

However, users may want to limit file sizes to prevent excessively large files, which is a common requirement for log rotation and file management systems.

### Example Configuration
```json
{
  "file": {
    "path": "/data/logs/app.log",
    "rollingSize": 10485760,  // 10MB
    "rollingCount": 0,
    "rollingInterval": 0
  }
}
```

Thank you for making such a lightweight tool.

### DIFF
--- a/internal/io/file/watch_test.go
+++ b/internal/io/file/watch_test.go
@@ -90,6 +90,7 @@ func TestWatchDir(t *testing.T) {
 	r := &WatchWrapper{f: &Source{}}
 	go func() {
 		time.Sleep(100 * time.Millisecond)
+		timex.Set(654321)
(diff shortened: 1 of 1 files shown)


ngjaying:
LGTM! This looks good and aligns perfectly with our roadmap.

Issue #3491, opened 2025-01-03
Add Support for Key, Partition Offset, and Headers in Kafka Source Plugin


wduczkowski:
Currently, the Kafka source plugin does not retrieve key, partition, offset  or header information from Kafka messages, even though the underlying library ([github.com/segmentio/kafka-go](https://github.com/segmentio/kafka-go)) already supports these parameters. This omission prevents users from accessing important metadata that can be crucial for debugging, routing, or processing logic.

Proposed Solution:

Extend the existing Kafka source plugin to include the retrieval of:

- Message Key 
- Partition
- Offset
- Headers 
- Topic

Update any existing data structures or interfaces within the plugin to store and return these additional metadata fields.
Ensure that backward compatibility is preserved if the plugin is already being used without this metadata.

Example Reference: Below is a sample code snippet using kafka-go that demonstrates how to read the  topic, partition, offset, key, value and headers:

```go
import (
	"context"
	"fmt"
	"log"
	"os"
	"strings"

	kafka "github.com/segmentio/kafka-go"
)

func getKafkaReader(kafkaURL, topic, groupID string) *kafka.Reader {
	brokers := strings.Split(kafkaURL, ",")
	return kafka.NewReader(kafka.ReaderConfig{
		Brokers:  brokers,
		GroupID:  groupID,
		Topic:    topic,
		MinBytes: 10e3, // 10KB
		MaxBytes: 10e6, // 10MB
	})
}

func main() {
	kafkaURL := os.Getenv("kafkaURL")
	topic := os.Getenv("topic")
	groupID := os.Getenv("groupID")

	reader := getKafkaReader(kafkaURL, topic, groupID)
	defer reader.Close()

	fmt.Println("start consuming ... !!")
	for {
		m, err := reader.ReadMessage(context.Background())
		if err != nil {
			log.Fatalln(err)
		}
		fmt.Printf("message at topic:%v partition:%v offset:%v  %s = %s  headers:%v\n",
			m.Topic, m.Partition, m.Offset, string(m.Key), string(m.Value), m.Headers)
	}
}

```
Event on kafka topic:
```
key: dev000001
```
```
header : {
     "app.id": "best-app-ever"
}
```
```
value: { "temperature":32.7063, "createTime":1735896706186, "correlationId":"bb889204-4ad7-4652-b18a-bc465592302d"} 
```
output:

```
[...]

PR #3923, opened 2025-11-11
feat: add monitoring for kafka message metadata

Elia-Renzoni:
This PR introduces monitoring for Kafka message metadata to improve observability 
and traceability of message processing.

Related Issue: #3491

### DIFF
--- a/extensions/impl/kafka/metrics.go
+++ b/extensions/impl/kafka/metrics.go
@@ -51,8 +51,8 @@ var (
 		Namespace: "kuiper",
 		Subsystem: "kafka_source",
 		Name:      "messages_counter",
-		Help:      "Kafka messages processed per rule, labeled by topic, partition, key and headers",
-	}, []string{LblTopic, LblPartition, LblKey, LblHeaders, metrics.LblRuleIDType, metrics.LblOpIDType})
+		Help:      "Kafka messages processed per rule, labeled by topic and partition",
+	}, []string{LblTopic, LblPartition, metrics.LblRuleIDType, metrics.LblOpIDType})
 
 	KafkaSourceGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "kuiper",
--- a/extensions/impl/kafka/sink.go
+++ b/extensions/impl/kafka/sink.go
@@ -48,8 +48,6 @@ const (
 	LblBytes     = "bytes"
 	LblOffset    = "offset"
 	LblTopic     = "topic"
-	LblKey       = "key"
-	LblHeaders   = "headers"
 	LblPartition = "partition"
 )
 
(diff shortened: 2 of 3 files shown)


extensions/impl/kafka/metrics.go
```diff
@@ -47,6 +47,27 @@ var (
...
+		Help:      "counter of key occurrences",
+	}, []string{LblKey, metrics.LblRuleIDType, metrics.LblOpIDType})
+
+	KafkaSourceHeadersCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "kuiper",
+		Subsystem: "kafka_source",
+		Name:      "headers_counter",
+		Help:      "counter of header occurrences",
+	}, []string{LblHeaders, metrics.LblRuleIDType, metrics.LblOpIDType})
+
```
Yisaer: I think this can be one counter with different labels.

Elia-Renzoni:
@Yisaer 
I’ve updated the code. Is it ready to merge, or do I need to make any more changes? Thanks!

extensions/impl/kafka/metrics.go
```diff
@@ -47,6 +47,13 @@ var (
 		Help:      "counter of Kafka Source IO",
 	}, []string{metrics.LblType, metrics.LblRuleIDType, metrics.LblOpIDType})
 
+	KafkaSourceMessageCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "kuiper",
+		Subsystem: "kafka_source",
+		Name:      "messages_counter",
+		Help:      "Kafka messages processed per rule, labeled by topic, partition, key and headers",
+	}, []string{LblTopic, LblPartition, LblKey, LblHeaders, metrics.LblRuleIDType, metrics.LblOpIDType})
```
Yisaer: I think expose `key/header` in the metrics may not be a proper design, this may cause high dimension issue.

Elia-Renzoni: @Yisaer 
Thanks for the feedback! I’ve made the requested changes, I think this approach should work better now. Let me know what you think!
We could also consider keeping only the topic and removing the partition if that makes more sense.

Yisaer:
@Elia-Renzoni What's the purpose for adding `topic/partition` label in the metrics as it's already been defined in the stream properties.

Elia-Renzoni:
@Yisaer Thanks for the quick response!
The main reason, I guess, was that it made things easier to work with directly from the monitoring system, without having to cross-reference the stream properties every time. But if this introduces redundancy or isn’t aligned with the expected design, we can definitely remove them, let me know what you prefer.

ngjaying:
@Elia-Renzoni Thanks for the contribution. We think this may not help the users so we decide to close this.

PR #3946, opened 2025-12-11
feat(plugin): sql plugin support DuckDB

L-607:


ngjaying:
go.sum should be committed alongside go.mod updates.

ngjaying:
@L-607 Looks like some dependency needs to be installed to build in this DuckDB driver. Check the build package failures.

L-607:
> @L-607 Looks like some dependency needs to be installed to build in this DuckDB driver. Check the build package failures.

Looks like should add lib for debian and alpine，need to modify the Dockerfile; I'll try it.
https://github.com/lf-edge/ekuiper/tree/master/.github/dockerfile

PR #4020, opened 2026-04-08
docs: fix conf api example

piepor:
Previous example led to parsing error. This commit fixes the curl http put request, letting the user simply copy and paste it to create a demo configuration key.


ngjaying (approved):
LGTM, thanks

PR #4042, opened 2026-05-13
fix(planner): cancel partial topo on plan failure

ngjaying:
This PR fixes a critical resource leak where shared sub-topologies (e.g., MQTT shared connections) would leak references and output channels if a rule failed during the building phase (e.g., invalid actions or converters).

### Changes:
- **Topology Cancellation**: Added a deferred \	p.Cancel()\ in \planner.createTopo\ to ensure that any partially built topology is cleaned up if buildOps or buildActions fails.
- **Sub-topology Cleanup**: Added a deferred \srcSubtopo.Close()\ in \splitSource\ to release shared sub-topology references if subsequent operator creation fails.
- **Variable Shadowing Fix**: Fixed a bug in \splitSource\ where the deferred cleanup missed errors due to block-scoped variable shadowing.
- **Regression Tests**: Added tests in \planner_source_test.go\ and \subtopo_test.go\ to verify that shared sub-topology references are correctly released and pool entries are removed when planning fails at various stages.

### DIFF
--- a/internal/topo/planner/planner.go
+++ b/internal/topo/planner/planner.go
@@ -290,6 +290,17 @@ func createTopo(rule *def.Rule, lp LogicalPlan, mockSourcesProp map[string]map[s
 	tp.SetStreams(streamsFromStmt)
 	tp.SetSinkSchema(schema)
 
+	// Cancel the partial topo on any planning error so that shared connection/stream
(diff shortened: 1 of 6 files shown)


ngjaying:
Fixed the error return tracking by using named return variables, and fixed the comment typo.

Yisaer (approved):

PR #4043, opened 2026-05-13
fix(topo): refactor pool to avoid close race

ngjaying:
## Summary
- Fixed a critical ABA race condition in sub-topology management where rules could be stuck with "zombie" sub-topologies.
- Refactored sub-topology lifecycle management to be centrally handled by the pool, ensuring atomic state transitions.

## Why
- The previous implementation used a two-phase lock release in `SrcSubTopo.Close()` that opened a window for new rules to acquire a sub-topology while it was being destroyed.
- This led to "dead connection" states where a rule's source would never produce data because it held a reference to a closed sub-topology.
- The root cause was an inverted lock order (sub-topology then pool) that forced the split lock release to avoid ABBA deadlocks.

## Changes In This PR
- **Architecture Refactor**: Moved all pool-lifecycle logic (reference removal, cancel, pool eviction, and chained close) into `CloseSubTopo` in `internal/topo/subtopo_pool.go`.
- **Atomic State Transitions**: `CloseSubTopo` now acquires the pool lock first, then the sub-topology lock, ensuring that the decision to destroy and the removal from the pool are one atomic step.
- **Improved Isolation**: `SrcSubTopo` no longer reaches into pool internals; its `Close()` method now delegates to the pool manager.
- **Enhanced Testing**: Added regression tests to verify that rule cancellation correctly releases sub-topology references and that no zombie states are created during concurrent access.

## Notes
- This fix eliminates the "dead connection" issue while maintaining a safe lock ordering that prevents deadlocks.

### DIFF
--- a/internal/topo/subtopo.go
+++ b/internal/topo/subtopo.go
@@ -140,34 +140,11 @@ func (s *SrcSubTopo) Open(ctx api.StreamContext, parentErrCh chan<- error) {
 	}
 }
 
-// Close is different from main topo because this will run multiple times.
-// Close ref rule will run close subtopo
+// Close satisfies the node.DataSourceNode interface. All pool-lifecycle logic
+// (reference counting, pool eviction, chained close) lives in CloseSubTopo so
+// that subtopo_pool.go is the single owner of the pool lock and map.
 func (s *SrcSubTopo) Close(ctx api.StreamContext) {
-	s.Lock()
-	isStop, isDestroy := s.removeRef(ctx)
-	if isDestroy { // destroy this subtopo
-		// The cleanup logic is moved to CloseSubTopo
-		// It will unlock inside CloseSubTopo
-	} else {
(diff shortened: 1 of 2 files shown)


ngjaying:
I've updated the PR to address the feedback:

1. **Schema Leak Fixed**: Modified CloseSubTopo to call schemaLayer.Detach even on the isDestroy path. This ensures RemoveRuleSchema is always called for the global registry, preventing stale entries and memory leaks.
2. **Safe Type Assertion**: Replaced the hard assertion ctx.(*kctx.DefaultContext) with a safe ok-checked version to prevent panics in non-standard context scenarios.
3. **Lock Contention**: Decided to maintain the full scope of the pool lock during CloseSubTopo. This ensures absolute atomicity for the 'decide to destroy' and 'evict from pool' steps, which is necessary to robustly eliminate the ABA race condition. Given this is a control-plane operation, the safety and simplicity of the current locking model outweigh the benefits of micro-optimizing contention here.

Yisaer (approved):

PR #4049, opened 2026-06-04
refactor: use strconv.FormatFloat

Elia-Renzoni:
### Summary
This PR fixes an issue identified by Copilot in the recently merged PR (#4025).
The ```NumberLiteral``` struct implemented the ```String()``` method using ```fmt.Sprintf("%f")```. As a result, calling ```String()``` produced a float representation with trailing zeros.

For example, a number such as 55.340 would be printed as 55.340000.

### Changes
This change replaces fmt.Sprintf with strconv.FormatFloat, ensuring a more accurate representation of float literals.


ngjaying (approved):
LGTM, thanks

PR #4055, opened 2026-06-25
test: add error check after cacheOp.Exec

piepor:
Previously, any error sent by cacheOp.Exec(ctx, errCh) to errCh was ignored; this change explicitly checks the error and handles it gracefully.

### DIFF
--- a/internal/topo/node/cache_op_test.go
+++ b/internal/topo/node/cache_op_test.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/lf-edge/ekuiper/v2/internal/conf"
 	"github.com/lf-edge/ekuiper/v2/internal/pkg/def"
@@ -119,6 +120,11 @@ func TestCacheRun(t *testing.T) {
 	index := 0
 	errCh := make(chan error)
 	cacheOp.Exec(ctx, errCh)
+	select {
+	case err := <-errCh:
+		require.NoError(t, err)
+	case <-time.After(100 * time.Millisecond):
+	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			var r any


ngjaying (commented):
I don’t think this fixed delay is useful here. `TestCacheRun` already validates correctness through the processed metrics and expected outputs. Since `cacheOp.Exec` starts async work,
  checking `errCh` once after 100ms only samples one moment in time and does not catch errors emitted later while the test is actually exercising the operator.

  It also makes the test slower without adding a deterministic readiness signal. What specific failure is this intended to fix? If we want to validate async operator errors, the waits/
  receives in the test should select on `errCh` while waiting for processing/output. Otherwise I’d prefer to remove this block.

piepor:
The problem I was trying to fix is  a segmentation fault `panic: runtime error: invalid memory address or nil pointer dereference` during TestCacheRun: if cacheOp.Exec fails due to en error in InitStore (failure mode in my case) s.prepareExec is skipped and the StatManager is not created. 

```go
func (s *CacheOp) Exec(ctx api.StreamContext, errCh chan<- error) {
	if len(s.outputs) > 1 {
		infra.DrainError(ctx, fmt.Errorf("cache op should have only 1 output but got %+v", s.outputs), errCh)
	}
	s.cache.SetupMeta(ctx)
	if err := s.cache.InitStore(ctx); err != nil {
		infra.DrainError(ctx, fmt.Errorf("cache op init store error:%v", err), errCh)
		return
	}
	s.prepareExec(ctx, errCh, "op")
	go func() {
        ...
  }()
}
```

Consequently, cacheOp.statManager.GetMetrics() does not exists: hence the nil pointer issue when it gets called during the tests in TestCacheRun:

```go
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			var r any
			for index < tt.sendUntil {
				cacheOp.input <- tuples[index]
				index++
			}
			timex.Add(100 * time.Millisecond)
			for { // wait until all processed
				processed := cacheOp.statManager.GetMetrics()[2]
				if processed == int64(tt.sendUntil) {
					break
				}
				time.Sleep(10 * time.Millisecond)
			}
      ... 
      }
```

I agree the actual test has an unnecessary delay: perhaps we could check for the two synchronous errors of  cacheOp.Exec (caused by len(outputs) > 1 and err := s.cache.InitStore) without delay and then requiring that cacheOp.statManager does exits before running tests in the loop?

Yisaer (approved):

ngjaying:
Closing this. The 100ms errCh poll cannot catch the InitStore failure (unbuffered channel + non-blocking DrainError drops the error before the select runs), and it does not observe errors emitted later while the test is waiting on metrics/output. It only adds latency to the happy path.

TestCacheRun already validates correctness via processed metrics and expected tuples. If we want a clearer failure when setup skips prepareExec, buffer errCh and require.NotNil on statManager after Exec (and optionally select on errCh in the wait loop). Happy to review that as a follow-up.

PR #4067, opened 2026-07-16
fix(sql): reject unsafe dynamic field names

Yisaer:
## Summary

- Reject an SQL sink row before it reaches the database when `fields` is unset and an incoming map key could change SQL syntax.
- Apply the check to the normal insert path and the row-kind insert, update, and delete paths.
- Keep explicitly configured `table`, `fields`, and `keyField` text unchanged so existing database-specific quoting continues to work.
- Document the new dynamic-field restriction and compatibility behavior in English and Chinese.

## Why

When `fields` is not configured, the SQL sink uses incoming map keys directly as column names. A publisher could therefore put SQL syntax in a key and change the generated statement.

Trying to quote every identifier in the sink is not a safe compatibility fix: SQL Server brackets, MySQL backticks, Oracle database links, case folding, and other database rules differ. This PR separates the two input classes instead. Untrusted dynamic keys are validated, while trusted SQL configuration keeps its existing semantics.

## Changes In This PR

- Accept dynamic field names only when they match `[A-Za-z_][A-Za-z0-9_]*`.
- Return an error for the whole row when any generated field name is invalid; do not filter or partially use the map keys.
- Preserve configured identifier text in the generated INSERT, UPDATE, and DELETE statements.
- Add regression tests for crafted keys and for existing SQL Server, Oracle, and MySQL identifier syntax.
- Update the English and Chinese SQL sink documentation with the validation rule, batch behavior, error behavior, and configured-field semantics.

## Notes

- Impact is limited to the SQL sink when `fields` is unset.
- Dynamic names containing spaces, punctuation, non-ASCII characters, or a leading digit now fail with a clear error. Users who need database-specific identifier syntax can continue to configure `fields` explicitly.
- Provision-time validation of configured `table`, `fields`, and `keyField` is intentionally outside this PR and will be handled separately after defining its compatibility contract.
- Verified with `go test ./extensions/impl/sql/... -count=1`, `go test -race ./extensions/impl/sql -count=1`, `go vet ./extensions/impl/sql`, local `golangci-lint run ./extensions/impl/sql/...`, and `git diff --check`.


ngjaying:
- [P1] Use dialect-specific identifier quoting — /home/ngjaying/repo/ekuiper/extensions/impl/sql/sink.go:103-104
    When the sink connects through the supported MySQL driver with its default SQL mode, double quotes delimit strings rather than identifiers, so statements such as INSERT INTO
    "t" ("a") ... fail with a syntax error. Since MySQL is a documented and prominently sampled target, identifier quoting must depend on the parsed driver (for example, backticks for
    MySQL) rather than universally using double quotes.

  - [P2] Preserve qualification when quoting table names — /home/ngjaying/repo/ekuiper/extensions/impl/sql/sink.go:355-355
    When table is schema-qualified, such as public.events, quoting the entire value generates "public.events", which refers to one identifier containing a dot rather than table events in
    schema public. Configurations that previously used qualified table names will therefore fail; each qualified component needs separate validation and quoting.

  - [P2] Handle filtering that removes every key — /home/ngjaying/repo/ekuiper/extensions/impl/sql/sink.go:341-347
    When fields is unset and a nonempty message contains only rejected names, extractKeys returns an empty but non-nil slice, so the callers still execute SQL like INSERT INTO ... ()
    values ();. This either fails repeatedly on PostgreSQL/SQLite or may insert a default-valued row on MySQL; return an error or skip the message when no usable keys remain.

ngjaying (changes requested):
Breaks MySQL

extensions/impl/sql/sink.go
```diff
@@ -324,13 +376,19 @@ func (s *SQLSinkConnector) extractKeys(item map[string]any) []string {
 	}
 	keys := make([]string, 0, len(item))
 	for k := range item {
-		keys = append(keys, k)
+		if isValidIdentifier(k) {
```
ngjaying: **[suggestion]** The change installs two parallel defenses—`isValidIdentifier` silent filtering in `extractKeys`, plus dialect-aware `quoteIdentifier`/`quoteTableName` at every SQL build site—when correct quoting alone already prevents the stated attack (metacharacters become part of a delimited identifier; embedded quote chars are doubled). The filter is the expensive half of the design: it pulls in a package-level regexp, forces three new empty-key branches, and changes runtime behavior for any legitimate key that is not a bare ASCII identifier (hyphens, leading digits, non-ASCII names, etc.). It is also applied inconsistently: configured `fields` / `keyField` / `table` are never validated, only quoted. That asymmetry signals the wrong abstraction boundary rather than a deliberate security model.

**Suggestion:** Prefer the code-judo simplification—delete `validIdentifierPattern` / `isValidIdentifier` and the empty-key skip paths, and keep dialect-aware quoting as the single identifier defense. If an allowlist is still desired for defense-in-depth, apply it uniformly (including `fields`/`keyField`/`table` at `Provision`) and fail closed with an error instead of dropping keys. Either way, one mechanism should own identifier safety so build sites only quote and assemble SQL.

extensions/impl/sql/sink.go
```diff
@@ -172,18 +218,19 @@ func (s *SQLSinkConnector) Collect(ctx api.StreamContext, item api.MessageTuple)
 func (s *SQLSinkConnector) collect(ctx api.StreamContext, item map[string]any) (err error) {
 	if len(s.config.RowKindField) < 1 {
 		keys := s.extractKeys(item)
```
ngjaying: **[bug]** When `fields` is unset and validation removes every key, `collect` / `collectList` / `save` log a warning and `return nil`. That is treated as a successful sink write: no error is surfaced to the rule, no exception metric is recorded, and no retry occurs. Partial filtering is worse in a different way—invalid keys are dropped inside `extractKeys` with no log at all, so an INSERT/UPDATE can succeed with a silent subset of columns. For a security fix this fail-open path both hides attack traffic and turns non-malicious naming mismatches into silent data loss.

**Suggestion:** Fail closed: return a clear error when no usable columns remain (and ideally when any untrusted key fails validation, if the filter is kept). At minimum, never report success for a skipped write, and log each dropped key. Better, remove the filter (Issue 1) so this branch is unnecessary; empty input is already rejected by `getValuesByKeys` via `"data is empty"`.

extensions/impl/sql/sink.go
```diff
@@ -93,6 +99,41 @@ func quoteSQLString(s string) string {
 	return "'" + strings.ReplaceAll(s, "'", "''") + "'"
 }
 
+// quoteIdentifier wraps a SQL identifier in dialect-appropriate quotes and escapes
+// embedded quote characters. This prevents SQL injection through attacker-controlled
+// column/table names.
+func quoteIdentifier(driver, identifier string) string {
```
ngjaying: **[bug]** There are no new unit tests covering identifier quoting, dialect selection, schema-qualified tables, or injection payloads. Existing sink tests are behind `//go:build mysql_integration_test` and only exercise happy-path inserts/updates with benign column names. A regression that reintroduces raw key interpolation, breaks MySQL backtick mode, or mishandles `"public.events"` would not be caught by CI’s default test set. For a SQL-injection fix this is a correctness gap, not a coverage nit.

**Suggestion:** Add build-tag-free table-driven tests for `quoteIdentifier`, `quoteTableName`, `identifierQuoteChar`, and `buildInsertSQL` / UPDATE-DELETE string assembly. Include cases such as: `a); DROP TABLE t;--`, embedded quote characters, MySQL vs postgres quote chars, `public.events`, and (if filtering remains) invalid-key rejection/error behavior. Assert on the generated SQL string so the security contract is locked without a live DB.

extensions/impl/sql/sink.go
```diff
@@ -93,6 +99,41 @@ func quoteSQLString(s string) string {
...
+func quoteTableName(driver, table string) string {
+	parts := strings.Split(table, ".")
+	for i, p := range parts {
+		parts[i] = quoteIdentifier(driver, p)
+	}
+	return strings.Join(parts, ".")
+}
+
+// identifierQuoteChar returns the SQL identifier quoting character for a given driver.
+func identifierQuoteChar(driver string) string {
```
ngjaying: **[suggestion]** Driver handling is threaded as a free-floating `driver string` through package functions and call sites (`buildInsertSQL(driver, ...)`, `quoteIdentifier(drv, key)`, local `drv := s.config.driver` in UPDATE/DELETE), but `identifierQuoteChar` only special-cases `"mysql"` (the `"sqlserver"/"mssql"` arms are identical to `default`). Unlike `lookupSource.buildGen`, the switch does not `strings.ToLower` the driver name, so any non-lowercase driver value silently falls into double-quote mode and can break MySQL again. The net abstraction is more surface area than behavior.

**Suggestion:** Collapse quoting onto `sqlSinkConfig` methods that close over `c.driver` (e.g. `c.quoteID(name)`, `c.quoteTable()`), compare drivers case-insensitively, and reduce the switch to mysql-backtick vs default double-quote unless SQL Server bracket quoting is intentionally supported. That removes the extra parameters and repeated `drv` locals without changing behavior.

ngjaying (changes requested):
## Summary

This PR hardens the SQL sink against identifier injection by (1) dialect-aware quoting of table/column names and (2) silently filtering message keys that fail `^[A-Za-z_][A-Za-z0-9_]*$` when `fields` is unset. Dialect quoting (especially MySQL backticks) and `quoteTableName` correctly address the earlier review feedback and are the right core fix. The dominant remaining problem is structural: the allowlist filter is largely redundant with correct quoting, introduces silent data-loss behavior (including success-on-skip), and is applied inconsistently only on the dynamic-key path—while the security-sensitive change still has no unit tests that lock in injection resistance.

## Issue counts by severity

- bugs: 2
- suggestions: 2
- nits: 0

extensions/impl/sql/sink.go
```diff
@@ -93,6 +94,34 @@ func quoteSQLString(s string) string {
 	return "'" + strings.ReplaceAll(s, "'", "''") + "'"
 }
 
+// quoteIdentifier wraps a SQL identifier in dialect-appropriate quotes and escapes
+// embedded quote characters. This prevents SQL injection through attacker-controlled
+// column/table names.
+func (c *sqlSinkConfig) quoteIdentifier(identifier string) string {
+	q := c.identifierQuoteChar()
+	return q + strings.ReplaceAll(identifier, q, q+q) + q
```
ngjaying: **[P1] Preserve unquoted identifier case semantics**

When targeting Oracle, ordinary unquoted schema objects are stored uppercase, but this now turns configurations such as `table: events` and field `id` into exact-case identifiers `"events"` and `"id"`, causing ORA-00942/ORA-00904. PostgreSQL mixed-case configurations regress similarly because unquoted names previously folded lowercase; normalize names according to each dialect's unquoted rules before delimiting them.

extensions/impl/sql/sink.go
```diff
@@ -93,6 +94,34 @@ func quoteSQLString(s string) string {
...
+	return strings.Join(parts, ".")
+}
+
+// identifierQuoteChar returns the SQL identifier quoting character for the configured driver.
+func (c *sqlSinkConfig) identifierQuoteChar() string {
+	switch strings.ToLower(c.driver) {
+	case "mysql":
+		return "`"
+	default:
+		return "\""
```
ngjaying: **[P1] Use the correct delimiter for every supported driver**

When eKuiper is built with supported drivers such as `mymysql`, Hive, or Spanner, this fallback emits double quotes even though those dialects require backticks for identifiers. For example, `ParseDriver` returns `mymysql`, so a default MySQL server receives `INSERT INTO "t" ("a") ...` and rejects every write because double quotes delimit strings unless `ANSI_QUOTES` is enabled.

ngjaying (changes requested):
The injection mitigation introduces runtime regressions for the default Oracle backend and several supported optional drivers. Identifier quoting must preserve each dialect's case-folding and delimiter semantics.

ngjaying (changes requested):
The common Oracle case-folding regression is fixed, but valid Oracle table references that worked before this PR are still rewritten into different identifiers.

[P1] Preserve existing Oracle table-reference semantics

- An already quoted table such as `"MixedCase"` was previously emitted unchanged, but is now uppercased and quoted again as `"""MIXEDCASE"""`, so the existing case-sensitive object is no longer addressable. Oracle DBMS_ASSERT.ENQUOTE_NAME preserves input that is already quoted.
- A remote reference such as `events@remote` was previously valid Oracle syntax, but now becomes `"EVENTS@REMOTE"`. That names one local identifier instead of appending the required `@dblink` qualifier.

Please parse Oracle table references rather than using only `strings.Split(table, ".")`: recognize `.` and the first `@` only outside quoted identifiers, validate and preserve already-quoted components, and validate/preserve the database-link qualifier separately. For ordinary valid identifiers, keeping them unquoted lets Oracle perform its native database-character-set case folding and avoids trying to reproduce that behavior with Go `strings.ToUpper`.

These are regressions relative to the base behavior, so Oracle compatibility is not yet preserved.

extensions/impl/sql/sink.go
```diff
@@ -93,6 +94,80 @@ func quoteSQLString(s string) string {
...
+// embedded quote characters. It also normalizes identifier case according to the
+// database's unquoted case-folding rules so that the quoted form matches what the
+// database would have stored for an unquoted identifier. If the identifier is already
+// quoted with the correct dialect quote character, it is preserved as-is to respect
+// the operator's explicit casing choice (e.g. Oracle "MixedCase"). This prevents
+// SQL injection through attacker-controlled column/table names.
+func (c *sqlSinkConfig) quoteIdentifier(identifier string) string {
+	q := c.identifierQuoteChar()
+	// If already quoted with this dialect's quote character, preserve as-is.
+	if len(identifier) >= 2 && identifier[0] == q[0] && identifier[len(identifier)-1] == q[0] {
```
ngjaying: [P0] Do not treat matching first/last quote bytes as proof that this is one valid quoted identifier. When `fields` is unset, `collect` passes attacker-controlled message keys here. For PostgreSQL/SQLite, a key such as `"x") VALUES ('safe'); DROP TABLE secret;--"` is returned verbatim and produces `INSERT INTO "t" ("x") VALUES ('safe'); DROP TABLE secret;--") values ('ignored');`. I verified that SQLite executes the stacked statement and drops `secret`. Parse and validate exactly one dialect-quoted identifier (including escaped internal delimiters), then requote it, or reject malformed quoted input. Add this quoted-payload case as a regression test.

extensions/impl/sql/sink.go
```diff
@@ -93,6 +94,80 @@ func quoteSQLString(s string) string {
...
+
+// quoteTableName splits a possibly schema-qualified table name (e.g. "public.events")
+// by dot and quotes each component with the dialect-appropriate quote character.
+// It also recognizes and preserves Oracle dblink syntax (table@remote), keeping the
+// @dblink suffix unquoted after the quoted identifier.
+func (c *sqlSinkConfig) quoteTableName(table string) string {
+	// Parse out Oracle dblink suffix (first @ not inside quotes).
+	idPart, dblink := splitDblink(table)
+
+	parts := strings.Split(idPart, ".")
```
ngjaying: [P1] This still splits `.` inside quoted identifiers. A valid table such as `"audit.v1"` or `"My.Schema"."Events"` is rewritten into different components, although it was previously emitted unchanged. There is a related cross-dialect problem: `splitDblink` tracks only double quotes and runs for every driver, so a MySQL name like `` `a@b` `` is split at `@`. Parse the table reference with the configured dialect delimiter, split `.` only outside quotes, and apply Oracle dblink handling only to `oracle`/`godror`.

extensions/impl/sql/sink.go
```diff
@@ -93,6 +94,80 @@ func quoteSQLString(s string) string {
...
+			}
+		}
+	}
+	return s, ""
+}
+
+// identifierQuoteChar returns the SQL identifier quoting character for the configured driver.
+func (c *sqlSinkConfig) identifierQuoteChar() string {
+	switch strings.ToLower(c.driver) {
+	case "mysql", "mymysql", "hive", "spanner":
```
ngjaying: [P1] At least one supported driver is still missing: Impala requires backticks for quoted identifiers, but `impala` falls through to the double-quote default. This makes generated INSERT/UPDATE/DELETE SQL invalid for the optional Impala backend. The repository registers an Impala SQL driver, and the official identifier documentation recommends always quoting generated identifiers with backticks: https://impala.apache.org/docs/build/plain-html/topics/impala_identifiers.html. Please audit the complete registered driver set rather than only the drivers already named in the prior thread, and add an Impala case here.

extensions/impl/sql/sql_quote_test.go
```diff
@@ -0,0 +1,494 @@
...
+			keyField: "a",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cfg := &sqlSinkConfig{
+				driver:   tt.driver,
+				KeyField: tt.keyField,
+			}
+			sqlStr := "UPDATE " + cfg.quoteTableName(tt.table) + " SET "
```
ngjaying: [P2] This test reconstructs the UPDATE statement inside the test and then checks that it contains outputs from the same helpers. It never calls `save` or another production SQL builder, so removing identifier quoting from the real UPDATE path would not fail this test. `TestBuildDeleteSQL` has the same problem. Exercise the actual production construction path (or extract pure UPDATE/DELETE builders used by both `save` and the tests) and assert the complete generated SQL, including malicious identifiers.

extensions/impl/sql/sql_quote_test.go
```diff
@@ -0,0 +1,494 @@
+// Copyright 2025 EMQ Technologies Co., Ltd.
```
ngjaying: [P3] Modified files must refresh the copyright year through 2026. This new file still says 2025; `sink.go` also remains at 2024-2025.

ngjaying (changes requested):
## Review summary

The latest update addresses the common Oracle case-folding and dblink cases, but this still must not merge. The new already-quoted shortcut reopens the original SQL injection path for attacker-controlled message keys. Quoted table components and the Impala dialect also remain incompatible, and the UPDATE/DELETE tests do not exercise production SQL construction.

Findings: P0: 1, P1: 2, P2: 1, P3: 1.

Both modified files must also refresh their copyright year to 2026: `extensions/impl/sql/sink.go` is still `2024-2025`, and `extensions/impl/sql/sql_quote_test.go` is still `2025`.

PR #4092, opened 2026-08-03
fix(planner): skip oauth section in findTemplateProps

ngjaying:
### Motivation

fix one bug for REST sink with oauth

### Changes

- findTemplateProps was recursively scanning all sink props including the oauth section. This caused `Bearer {{.access_token}}` from OAuth headers to be passed to TransformOp as a data template, where it would be evaluated against message data (which lacks access_token) instead of the token data.

- Skip the oauth key during template scanning so OAuth token templates are resolved only by the auth client, not the transform pipeline.

### Verification

- Unit test added: TestFindTemplates/skip_oauth


ngjaying:
Closing - this fix addresses the template scanning issue, not the reported bug 3 (unknown data type RawTuple). Will submit the correct fix separately.

PR #4100, opened 2026-08-05
feat(rule): add disable option for sink actions

Yisaer:
## Summary

  Add a common `enable` option for rule sink actions.

  When a sink action sets `"enable": false`, the planner skips constructing that sink pipeline. The option defaults to enabled when omitted.

  ## Changes

  - Support `enable` in SQL rule `actions`.
  - Support `enable` for graph rule sink nodes.
  - Skip disabled sinks before connection config overwrite and sink provisioning.
  - Return an error when no sink action is enabled.
  - Validate `enable` as a boolean value.
  - Document the new common sink property in EN/ZH docs.
  - Add planner tests for disabled sinks, all-disabled sinks, and invalid `enable` values.


docs/en_US/guide/sinks/overview.md
```diff
@@ -68,9 +68,10 @@ Each action can define its own properties. There are several common properties:
...
+| enable               | bool: true                           | Whether to enable this sink action. If set to false, the rule planner skips this sink and does not create its sink pipeline. At least one sink action in a rule must be enabled.                                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
```
ngjaying: I suggest to use disable, so that the default can be false and to backward compatible with rules that has not set this property

ngjaying (changes requested):

internal/topo/planner/planner_sink.go
```diff
@@ -30,31 +31,59 @@ import (
...
+				continue
+			}
+			enabledCount++
+			props, err = conf.OverwriteByConnectionConf(name, props)
 			if err != nil {
 				return err
 			}
 			sinkName := fmt.Sprintf("%s_%d", name, i)
-			cn, err := SinkToComp(tp, name, sinkName, copyProps(props), rule, streamCount, schema)
+			cn, err := SinkToComp(tp, name, sinkName, props, rule, streamCount, schema)
```
ngjaying: Still need copyProps here

ngjaying (changes requested):

ngjaying (approved):

PR #4103, opened 2026-08-05
feat(func): add acc_collect accumulator function

Yisaer:
### What                                                                                                                                                                          
                                                                                                                                                                                     
   Add `acc_collect` accumulator function that collects target field values from streaming events into a `[]interface{}` slice.                                                      
                                                                                                                                                                                     
   ### Behavior                                                                                                                                                                      
                                                                                                                                                                                     
   - Consistent interface with the existing `acc_*` function family: supports both `PARTITION BY` (3-arg) and `WHEN` conditional (5-arg) modes                                       
   - **nil values are skipped**, consistent with `acc_count`                                                                                                                         
   - Zero value is an empty slice `[]interface{}{}` (reset via `accReset`)                                                                                                           
   - No collection when `validData=false`                                                                                                                                            
                                                                                                                                                                                     
   ### Example                                                                                                                                                                       
                                                                                                                                                                                     
   ```sql                                                                                                                                                                            
   SELECT acc_collect(a) OVER (PARTITION BY key) FROM stream                                                                                                                         
 ```                                                                                                                                                                                 
                                                                                                                                                                                     
 With inputs 1, 2, nil, 3, the outputs are [1], [1,2], [1,2], [1,2,3].                                                                                                               
                                                                                                                                                                                     
 ```sql                                                                                                                                                                              
   SELECT acc_collect(a) OVER (WHEN a > 0) FROM stream                                                                                                                               
 ```                                                                                                                                                                                 
                                                                                                                                                                                     
 Uses onBegin / onReset to control the accumulation window.


fvt/ruletest_test.go
```diff
@@ -125,3 +125,114 @@ func (s *RuletestTestSuite) TestRuletestMockSourceUnnestKeepProjectedFields() {
...
+	s.T().Logf("Total results: %d", len(results))
+	for i, r := range results {
+		formatted, _ := json.MarshalIndent(r, "", "  ")
+		s.T().Logf("Result[%d]: %s", i, string(formatted))
+	}
+
+	// Basic sanity checks
+	s.Require().NotEmpty(results, "should have at least one output (cycle end)")
+	for _, r := range results {
+		s.Require().NotEmpty(r["acc_collected"], "acc_collected should not be empty")
```
ngjaying, line 236: Should test the output, not just not empty

ngjaying (changes requested):

fvt/ruletest_test.go
Yisaer: Good catch, fixed. Now asserting the exact output for both charge cycles:

- Cycle 1: flag_c=1754100107, acc_collected=[{metric:50,et:1754100101},{metric:60,et:1754100105},{metric:60,et:1754100107}]
- Cycle 2: flag_c=1754100118, acc_collected=[{metric:30,et:1754100112},{metric:40,et:1754100116}]

ngjaying (approved):

Issue #4099, opened 2026-08-04
feat: support collect filter inside window operators


ngjaying:
## Summary

Support a **collect filter** inside window operators so that `WHERE` conditions on event-driven windows (state/count/sliding/event-time) only buffer matching rows instead of all rows, reducing peak memory from O(all rows) to O(matching rows) while keeping the window's control logic (begin/emit/trigger/count) able to see every row.

### Problem scenario

A common state window rule:

```sql
SELECT collect(*) AS charge_data
FROM vehicle_status
WHERE soc % 10 = 0 AND changed_col(true, soc % 10 = 0)
GROUP BY statewindow(charge_status = 'charging', charge_status = 'discharging')
```

Expected semantics: partition by charge status, collect only the data points where `soc % 10` changes.

### Current behavior

`WHERE` is pushed below the window (via predicate push-down), so it runs **before** the window:

```
Project
  └── Window (statewindow)
        └── Filter (WHERE)        ← before the window
              └── AnalyticFuncs (changed_col)
                    └── DataSource
```

This causes two problems:

1. **Correctness bug:** The state window's `begin`/`emit` conditions need to see *every* row to detect `charge_status` transitions. `WHERE` filters out rows where `soc % 10 != 0` — including the very row that flips `charge_status` to `discharging` — so the window never opens/closes. (Partially addressed by #4098, which moves the `WHERE` to run after the window emit, but that is only a correctness fix and does not save memory.)

2. **Memory waste:** Even with `FILTER(WHERE)` or a post-window `WHERE`, the window scanner buffers the entire charging session (potentially thousands of rows) and filters down to ~10 rows only at emit time. Peak memory = O(all rows).

### Root cause

All eKuiper window operators share a "buffer then trigger" architecture:

```
row arrives -> [buffer to scanner] -> [trigger/control logic] -> emit scans buffer
```

Control logic (begin/emit/trigger/gap) reads the **current row**, while the buffer is purely the "data to emit". But the current mechanism does not distinguish "control rows" from "collect rows" — it either buffers everything, or pre-filters everything (dropping control rows too).

## Proposal: unified filter dichotomy

**A filter is either before the window (pre-filter) or a collect filter inside the window. There is no third option.**

This reuses the existing `WindowPlan.PushDownPredicate` branch split:

| Window type | Branch | Filter location | Semantics |
|---|---|---|---|
[...]

PR #4107, opened 2026-08-16
feat(windows): add internal filter operators

Elia-Renzoni:
### Summary
This PR introduces an internal **collect filter** inside event-driven window operators (`state`, `count`, `sliding`, and `event-time`). 

Instead of buffering every row into memory and filtering at emit time the window now evaluates the `WHERE` condition upon row arrival and only buffers matching rows. Control logic (`begin`, `emit`, `trigger`, `count`) still evaluates every arriving row to maintain correct execution semantics.

By assigning `WHERE` predicates directly to `WindowPlan.collectCondition`, the standalone `FilterPlan` node is eliminated automatically via `PushDownPredicate`. 

**Key benefit:** Reduces peak memory consumption from **O(all rows)** to **O(matching rows)** without breaking window state/transition semantics or requiring any SQL syntax changes.

Related Issue: #4099


Elia-Renzoni:
The failure of TestSingleSQLWithEventTime doesn't seem related to this patch. However, the JMX test failure is caused by it. I'll investigate further and update the failing test.

ngjaying:
@Elia-Renzoni We had just fixed lint errors due to lint version update. Now you can rebase on the latest master to resolve those repo lint errors.

internal/topo/planner/windowPlan.go
```diff
@@ -96,14 +106,16 @@ func (p *WindowPlan) PushDownPredicate(condition ast.Expr) (ast.Expr, LogicalPla
 	// before it; otherwise rows that trigger a state change could be filtered
 	// out and the window would never open/close.
 	if p.wtype == ast.COUNT_WINDOW || p.wtype == ast.SLIDING_WINDOW || p.wtype == ast.STATE_WINDOW {
-		return condition, p
+		p.collectCondition = condition
```
Elia-Renzoni: @ngjaying 
I think the failing test (TestRuleAggInWhere which contains the following rule: ```SELECT * FROM sim5 where a > avg(a) group by countwindow(2)```) reflects an edge case I hadn't considered. Currently, the WHERE filter sits inside the window, but an aggregation function needs a collection of messages to operate on. So, if the condition contains a FuncTypeAgg call, the filter should not be pushed inside the window. I'm proposing a function like this, to check if a condition contains an AggFunc:
```Go
func containsAggregationFuncs(filterCond ast.Expr) bool {  
    var match = false  
    ast.WalkFunc(filterCond, func(n ast.Node) bool {  
        switch f := n.(type) {  
        case *ast.Call:  
            if f.FuncType == ast.FuncTypeAgg {  
                match = true  
                return false  
            }  
        }  
        return true  
    })  
    return match  
}
```
and using it as a guard at the top of PushDownPredicate, like this:
```Go
if containsAggregationFuncs(condition) {  
    return condition, p  
}
```
What do you think?

ngjaying:
We have fixed the merge tool install problem in the ci. Try to rebase on the latest master

ngjaying (dismissed):
Requesting changes: the P0 below (agg-referencing WHERE pushed below its data dependency, verified end-to-end: base returns rows, this branch returns empty) must be fixed before merge. The other two are required cleanups in the same files.

internal/topo/planner/windowPlan.go
ngjaying, line 109: **P0: this pushdown moves the predicate below its data dependency and breaks aggregate-referencing WHERE.**

In standard SQL, WHERE indeed cannot see aggregates — but eKuiper deliberately deviates here:

1. `aggFuncChecker` (`analyzer.go`) only forbids nested aggregates, `NoAggFunc` inside aggregate queries, and aggregates in GROUP BY. Aggregates in WHERE are allowed;
2. `rewrite_planner.go:43 RewriteAggFunctionInWhere` is dedicated machinery for exactly this shape: it extracts aggregate calls from `Condition` into `AggFunc` output fields `$$agg_ref_N` and rewrites the original site to `bypass(...)`;
3. The old plan spells this out (see the pre-change expectation in `TestExplainAggInWhere`): `Project > Filter(a > bypass($$agg_ref_0)) > AggFunc > Window` — the Filter sits above AggFunc, where the reference is bound.

This PR makes `PushDownPredicate` swallow the condition unconditionally and return `nil` (eliminating `FilterPlan`), while `AggFuncPlan` has no `PushDownPredicate` of its own to intercept it. The new plan is `Project > AggFunc > Window(collectCondition)`, and collect is evaluated per input row inside the window, where `$$agg_ref_0` can never be bound. `collectConditionMatch` returns `(false, nil)` for unbound fields, so rows are silently dropped with no error.

Verified end-to-end with the same rule (`CountWindow(2)`):

```sql
SELECT * FROM demoE2 WHERE temp > avg(temp) GROUP BY CountWindow(2)
```

- base: returns `{temp:27.5}` (first window (27.5, 25.5), avg=26.5)
- this branch: returns empty (reproduces the regression)

Suggested fix: refuse to absorb predicates referencing downstream-produced fields (`$$agg_ref_*` / `bypass(...)`-wrapped; general principle: never absorb anything that cannot be bound at the new location) in this `PushDownPredicate`, leaving that part to the upstream `FilterPlan`. Predicates over pure input columns keep pushing down, so the memory goal is unaffected. Also please cover the SQL above with a runtime regression test — `TestExplainAggInWhere` only asserts plan shape and cannot catch this class of issue.

internal/topo/node/window_op.go
```diff
@@ -366,7 +368,22 @@ func (o *WindowOperator) execProcessingWindow(ctx api.StreamContext, inputs []xs
 			switch d := data.(type) {
 			case xsql.EventRow:
 				o.handleTraceIngestTuple(ctx, d)
-				inputs = append(inputs, d)
+
+				var filterMatch bool
```
ngjaying, line 372: **P2.1: restructure this branch around the invariant below instead of just adding a comment.**

There is exactly one correct behavior here (it is also the contract stated in the `TestCountWindowCollectCondition` comment): **a COUNT window's trigger count must observe every row, and narrowing happens only at emit time in `nextCountWindow`**; only non-COUNT windows filter at ingest. The current `var filterMatch bool` + two `append` sites reads as if COUNT could double-append, plus the `var err` shadowing. Make the branches mutually exclusive so the three facts are pinned down:

```go
if o.window.Type == ast.COUNT_WINDOW {
    // Count trigger counts every row; narrowing happens at emit in nextCountWindow.
    inputs = append(inputs, d)
} else if match, err := collectConditionMatch(ctx, d, o.window.CollectCondition, o.name); err != nil {
    o.onError(ctx, err) // eval error: report and drop the row
} else if match {
    inputs = append(inputs, d)
}
```

Key points: COUNT appends unconditionally; eval error = report + drop (spelled out so nobody later "fixes" it); normal mismatch = silently drop.

internal/topo/node/window_op.go
```diff
@@ -728,6 +765,15 @@ func (o *WindowOperator) scan(inputs []xsql.EventRow, triggerTime time.Time, ctx
 	}
 	log.Debugf("window %s triggered for %d tuples", o.name, len(inputs))
 
+	if o.window.CollectCondition != nil && len(rowContent) == 0 {
```
ngjaying, line 768: **P2.2: remove both empty-window skips and keep the original emit behavior.**

This gate and the `len(tsets.Content) == 0 → continue` in `execProcessingWindow` (~L452) are new "skip instead of emit when nothing matched" behavior. Neither is required for the memory goal — not buffering non-matching rows already saves the memory. **Skipping emission changes trigger semantics**: downstream window count, timing, and `records_out_total` all change (the forced `TestEventWindow` metric change `5→2` is the evidence).

Requirement: delete both gates so trigger timing and emission cardinality stay exactly as before (narrowed content is the point of the feature; empty windows still emit).

ngjaying (changes requested):
Requesting changes: the P0 below (agg-referencing WHERE pushed below its data dependency, verified end-to-end: base returns rows, this branch returns empty) must be fixed before merge. The other two are required cleanups in the same files.

Elia-Renzoni:
> Requesting changes: the P0 below (agg-referencing WHERE pushed below its data dependency, verified end-to-end: base returns rows, this branch returns empty) must be fixed before merge. The other two are required cleanups in the same files.

Thanks for the review. I'll address the requested changes and push an update within this week.

internal/topo/planner/windowPlan.go
Elia-Renzoni: Addressed here: 3da08de0fb3de4cc098e19e066ac356f1a676db1

internal/topo/node/window_op.go
Elia-Renzoni: Addressed here: 3da08de0fb3de4cc098e19e066ac356f1a676db1

internal/topo/node/window_op.go
Elia-Renzoni: Addressed here: 3da08de0fb3de4cc098e19e066ac356f1a676db1

internal/topo/planner/windowPlan.go
Elia-Renzoni: I've also updated the failing window test here: 3e3d65b15587141b94ff32ee0c65ea234b143ad1

internal/topo/node/window_op.go
Elia-Renzoni: I've also updated the failing window test here: 3e3d65b15587141b94ff32ee0c65ea234b143ad1

internal/topo/node/window_op.go
Elia-Renzoni: I've also updated the failing window test here: 3e3d65b15587141b94ff32ee0c65ea234b143ad1

PR #4110, opened 2026-08-20
feat(sink): add IoTDB predefined sink plugin

CritasWang:
## What

Add an IoTDB predefined sink plugin that writes data into Apache IoTDB using the native Thrift RPC client. It supports both the tree model and the table model.

## Changes

- feat(sink): add IoTDB predefined sink plugin
  - tree model: device/isAligned properties, aligned time-series support
  - table model: database/table/columnCategories properties
  - plugin descriptor, registration, English and Chinese docs, and unit tests
  - register the new sink docs in docs/directory.json
- fix(iotdb): harden sink writers and review fixes
  - use `*SessionPool` and guard `close()` so closing before a successful connect no longer panics
  - make auto-generated timestamps strictly increase per writer to avoid same-millisecond collisions
  - extract shared `fillTablet` helper
  - reject INT32/FLOAT overflow in `convertValue`
  - support IPv6 host literals in `splitAddr`
  - add tests for close-without-connect, monotonic timestamps, and overflow handling

## Verification

- `go test ./extensions/impl/iotdb/...` passes locally.
- `gofmt` / `gofumpt` / `gci` (repository settings) pass on changed Go files.
- `markdownlint` passes on changed docs, and the docs directory check passes after registration.
- Plugin compiles with `go build -trimpath --buildmode=plugin -o plugins/sinks/Iotdb.so extensions/sinks/iotdb/*.go`.

### DIFF
--- a/extensions/impl/iotdb/iotdb.go
+++ b/extensions/impl/iotdb/iotdb.go
@@ -20,7 +20,6 @@ import (
 	"github.com/lf-edge/ekuiper/contract/v2/api"
 
 	"github.com/lf-edge/ekuiper/v2/pkg/cast"
-	"github.com/lf-edge/ekuiper/v2/pkg/errorx"
 	"github.com/lf-edge/ekuiper/v2/pkg/model"
 )
 
@@ -86,7 +85,7 @@ func (s *iotdbSink) collect(ctx api.StreamContext, data any) error {
 	}
 	if err := s.writer.write(ctx, maps); err != nil {
 		ctx.GetLogger().Errorf("iotdb sink write error: %v", err)
-		return errorx.NewIOErr(fmt.Sprintf("iotdb sink fails to send out the data: %v", err))
+		return err
 	}
 	ctx.GetLogger().Debug("insert data into iotdb success")
 	return nil
--- a/extensions/impl/iotdb/iotdb.go
+++ b/extensions/impl/iotdb/iotdb.go
@@ -1,0 +1,131 @@
+// Copyright 2026 Timecho Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package iotdb
+
+import (
+	"fmt"
+
+	"github.com/lf-edge/ekuiper/contract/v2/api"
+
+	"github.com/lf-edge/ekuiper/v2/pkg/cast"
+	"github.com/lf-edge/ekuiper/v2/pkg/errorx"
+	"github.com/lf-edge/ekuiper/v2/pkg/model"
+)
+
+// writer is the strategy interface implemented by tree/table specific writers.
+type writer interface {
+	connect(ctx api.StreamContext, conf *iotdbConfig) error
+	write(ctx api.StreamContext, data []map[string]any) error
+	close() error
+}
+
+// iotdbSink is the IoTDB sink that delegates writes to a model-specific writer.
+type iotdbSink struct {
+	conf   iotdbConfig
+	writer writer
+}
+
+func (s *iotdbSink) Provision(ctx api.StreamContext, props map[string]any) error {
+	s.conf = iotdbConfig{}
+	if err := cast.MapToStruct(props, &s.conf); err != nil {
+		return fmt.Errorf("error configuring iotdb sink: %s", err)
+	}
+	s.conf.applyDefaults()
+	if err := s.conf.validate(); err != nil {
+		return err
+	}
+	switch s.conf.Model {
+	case modelTree:
+		s.writer = &treeWriter{}
+	case modelTable:
+		s.writer = &tableWriter{}
+	default:
+		// unreachable: validate already rejected other values
+		return fmt.Errorf("unsupported model %q", s.conf.Model)
+	}
+	return nil
+}
+
+func (s *iotdbSink) Connect(ctx api.StreamContext, sch api.StatusChangeHandler) error {
+	err := s.writer.connect(ctx, &s.conf)
+	if err != nil {
+		sch(api.ConnectionDisconnected, err.Error())
+		return err
+	}
+	sch(api.ConnectionConnected, "")
+	return nil
+}
+
+func (s *iotdbSink) Collect(ctx api.StreamContext, item api.MessageTuple) error {
+	return s.collect(ctx, item.ToMap())
+}
+
+func (s *iotdbSink) CollectList(ctx api.StreamContext, items api.MessageTupleList) error {
+	return s.collect(ctx, items.ToMaps())
+}
+
+func (s *iotdbSink) collect(ctx api.StreamContext, data any) error {
+	maps, err := toMapList(data)
+	if err != nil {
+		return err
+	}
+	if len(maps) == 0 {
+		return nil
+	}
+	if err := s.writer.write(ctx, maps); err != nil {
+		ctx.GetLogger().Errorf("iotdb sink write error: %v", err)
+		return errorx.NewIOErr(fmt.Sprintf("iotdb sink fails to send out the data: %v", err))
+	}
+	ctx.GetLogger().Debug("insert data into iotdb success")
+	return nil
+}
+
+func (s *iotdbSink) Close(ctx api.StreamContext) error {
+	ctx.GetLogger().Infof("iotdb sink close")
+	if s.writer != nil {
+		return s.writer.close()
+	}
+	return nil
+}
+
+// toMapList normalizes data into []map[string]any.
+func toMapList(data any) ([]map[string]any, error) {
+	switch dd := data.(type) {
+	case map[string]any:
+		return []map[string]any{dd}, nil
+	case []map[string]any:
+		return dd, nil
+	default:
+		return nil, fmt.Errorf("iotdb sink needs map or []map, but received unsupported data %T", dd)
+	}
+}
+
+// GetSink returns a new IoTDB sink instance.
+func GetSink() api.Sink {
+	return &iotdbSink{}
+}
+
+// Info implements model.SinkInfoNode to declare sink capabilities.
+func (s *iotdbSink) Info() model.SinkInfo {
+	return model.SinkInfo{
+		HasFields: true,
+	}
+}
+
+// interface compliance check
+var (
+	_ api.TupleCollector  = &iotdbSink{}
+	_ model.SinkInfoNode = &iotdbSink{}
+)
--- a/extensions/impl/iotdb/iotdb.go
+++ b/extensions/impl/iotdb/iotdb.go
@@ -126,6 +126,6 @@ func (s *iotdbSink) Info() model.SinkInfo {
 
 // interface compliance check
 var (
-	_ api.TupleCollector  = &iotdbSink{}
+	_ api.TupleCollector = &iotdbSink{}
 	_ model.SinkInfoNode = &iotdbSink{}
 )
--- a/extensions/impl/iotdb/iotdb_test.go
+++ b/extensions/impl/iotdb/iotdb_test.go
@@ -435,6 +435,7 @@ func TestConvertValue(t *testing.T) {
 		{"int to TIMESTAMP", 1700000000000, "TIMESTAMP", int64(1700000000000), false},
 		{"unsupported type", 1, "UNKNOWN", nil, true},
 		{"invalid bool source", "notabool", "BOOLEAN", nil, true},
+		{"int32 overflow errors", int64(1) << 40, "INT32", nil, true},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
@@ -519,4 +520,37 @@ func TestGetSink(t *testing.T) {
 	assert.NotNil(t, sink)
 	_, ok := sink.(*iotdbSink)
 	assert.True(t, ok)
+}
+
+func TestWriterCloseWithoutConnect(t *testing.T) {
+	// close() before a successful connect() must not panic even though the
+	// underlying session pool was never initialized.
+	assert.NotPanics(t, func() {
+		tw := &treeWriter{}
+		assert.NoError(t, tw.close())
+	})
+	assert.NotPanics(t, func() {
+		tbw := &tableWriter{}
+		assert.NoError(t, tbw.close())
+	})
+}
+
+func TestNextTimestamp(t *testing.T) {
+	// explicit timestamps are used verbatim and do not advance the auto cursor
+	var last int64
+	ts, err := nextTimestamp(map[string]any{"ts": 100}, "ts", &last)
+	assert.NoError(t, err)
+	assert.Equal(t, int64(100), ts)
+	assert.Equal(t, int64(0), last)
+
+	// auto timestamps strictly increase within a batch even within the same millisecond
+	last = 0
+	t1, err := nextTimestamp(map[string]any{}, "", &last)
+	assert.NoError(t, err)
+	t2, err := nextTimestamp(map[string]any{}, "", &last)
+	assert.NoError(t, err)
+	t3, err := nextTimestamp(map[string]any{}, "", &last)
+	assert.NoError(t, err)
+	assert.Greater(t, t2, t1)
+	assert.Greater(t, t3, t2)
 }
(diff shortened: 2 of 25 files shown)


ngjaying:
Thanks for contributing! Could you please fix the lint errors?

CritasWang:
Fixed the lint failures.

- Applied gofumpt formatting in `internal/binder/function`, `internal/topo/node`, `internal/topo/operator`, and `internal/topo/planner` files.
- Removed the three dead stores flagged by `staticcheck` (`SA4006`).
- Updated the PR title so the commit-message length check passes.

Locally, `golangci-lint run` now reports `0 issues`. The new workflow runs for this head are awaiting approval again, so they will run once approved.

CritasWang:
> Thanks for contributing! Could you please fix the lint errors?

fixed

ngjaying:
## Review

Well-structured PR - follows the existing `influx2` sink pattern (strategy `writer` interface, `Provision`/`Connect`/`Collect`/`Close`, `Info()` with `HasFields:true`). The hardening commit shows good care (nil-guarded `close()`, monotonic timestamps, INT32/FLOAT overflow). I verified the IoTDB `Tablet` API - `SetTimestamp`/`SetValueAt` do **not** touch `RowSize`, so the manual `tablet.RowSize++` in `fillTablet` is **correct and required** (serialization slices to `[0:RowSize]`). The `index++` removals in `planner_sink.go`/`planner_source.go` are safe (the value is never read afterward - true SA4006).

### Should fix

1. **`model` default is advertised but not applied.** `iotdb.json` and both docs declare `default: "tree"`, but `config.go` `applyDefaults()` never sets `c.Model`. A user who omits `model` (following the docs) gets `model must be either "tree" or "table"`. Add `if c.Model == "" { c.Model = "tree" }` to `applyDefaults()` (or drop the default from the descriptor/docs).

2. **Unvalidated identifier in DDL (injection surface).** `table_writer.go` builds `fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s", conf.Database)` - `conf.Database` is user-supplied rule config, which on many deployments is submitted via REST by non-admin users. Validate it against an identifier pattern (e.g. `^[A-Za-z_][A-Za-z0-9_]*$`) before interpolating; IoTDB may accept `;`-separated statements.

### Worth considering

3. **FLOAT overflow guard has a hole.** In `types.go` `convertValue` FLOAT branch: `if !math.IsInf(f, 0) && math.IsInf(float64(float32(f)), 0)` - if `f` is already ±Inf (e.g. a string like `"1e400"` or a direct Inf), the check is skipped and `float32(±Inf)` is written. Drop the `!math.IsInf(f,0) &&` guard so the `IsInf(float64(float32(f)),0)` test catches both. Also add a FLOAT-overflow test - `TestConvertValue` only covers INT32 overflow today.

4. **Coverage gap on the hot path.** `table_writer.go` (2.13%) / `tree_writer.go` (16.67%) - `connect`/`write`/`fillTablet` are untested. A unit test for `fillTablet` using a real `client.NewTablet` (no server needed) would assert `RowSize`, timestamps, values, and null-bitmap handling, exercising `convertValue`+`nextTimestamp` together.

5. **IPv6 claim untested.** The commit advertises IPv6 support via `net.SplitHostPort`, but `TestSplitAddr` has no `[::1]:6667` case - add one.

### Nits

6. `tree_writer` uses `pool.PutBack(session)` while `table_writer` uses `session.Close()` - both are correct (`SessionPool` vs `TableSessionPool` have different return APIs; `PooledTableSession.Close()` calls `PutBack` internally), but a one-line comment would save the next reader a library dive.

7. `nodeUrls` "overrides addr" per docs, but `splitAddr(conf.Addr)` still runs unconditionally - so `addr` must stay valid in cluster mode. Either skip `splitAddr` when `NodeUrls` is set, or soften the docs wording.

8. `validate()` mutates the config (upper-casing `DataTypes`/`ColumnCategories`, stripping `root.`). It works since it runs once in `Provision`, but separating normalization from validation would be cleaner.

The pre-existing lint fixes (`funcs_agg.go`, `funcs_math.go`, `window_op.go`, `field_processor.go`, the `SA4006` dead stores) are behavior-neutral and fine; only process-wise they'd ideally live in a separate PR - acceptable here since master CI was red.

CritasWang:
Addressed review items 1-7 in `7d8d4958`:

- Applied the documented default of `model=tree` when `model` is omitted.
- Validated table-model database names against `^[A-Za-z_][A-Za-z0-9_]*$` before DDL interpolation, with an injection-style regression test.
- Rejected both finite FLOAT overflow and pre-existing positive/negative infinity.
- Added a real `client.NewTablet` unit test covering `RowSize`, explicit and monotonic auto timestamps, value conversion, and null-bitmap behavior.
- Added the `[::1]:6667` IPv6 case.
- Documented why tree sessions use `PutBack` while pooled table sessions use `Close`.
- Made `nodeUrls` authoritative, so cluster mode no longer parses or requires a valid `addr`.

I left item 8 unchanged because normalization/validation separation is a non-blocking structural refactor and would broaden this correctness/security patch.

Local validation:

- `go test -race ./extensions/impl/iotdb`
- `go test -count=3 ./extensions/impl/iotdb`
- `go vet ./extensions/impl/iotdb ./extensions/sinks/iotdb`
- `go test ./internal/topo/planner`

The IoTDB implementation package now reports 55.5% statement coverage locally.

extensions/impl/iotdb/iotdb.go
```diff
@@ -0,0 +1,131 @@
...
+	maps, err := toMapList(data)
+	if err != nil {
+		return err
+	}
+	if len(maps) == 0 {
+		return nil
+	}
+	if err := s.writer.write(ctx, maps); err != nil {
+		ctx.GetLogger().Errorf("iotdb sink write error: %v", err)
+		return errorx.NewIOErr(fmt.Sprintf("iotdb sink fails to send out the data: %v", err))
```
ngjaying: **[bug]** `collect` wraps every `writer.write` failure in `errorx.NewIOErr`, including non-IO errors from `convertValue` / `fillTablet` (INT32/FLOAT overflow, uncastable values, invalid timestamps). Sink retry (`resendInterval` / cache) treats IO errors as recoverable, so a poison row will be retried indefinitely instead of being dropped. Influx2 and SQL sinks only wrap the actual network/DB call.

**Suggestion:** Return conversion/schema errors as plain errors; wrap only session acquire and insert failures. Split `fillTablet` from the insert, or use a typed error the collector can inspect.

extensions/impl/iotdb/table_writer.go
```diff
@@ -0,0 +1,126 @@
...
+	poolConfig, err := conf.newPoolConfig()
+	if err != nil {
+		return err
+	}
+
+	pool := client.NewTableSessionPool(poolConfig, conf.PoolSize, int(conf.Timeout), 60000, false)
+	w.pool = &pool
+
+	// 自动创建数据库（table dialect 下的 CREATE DATABASE）
+	autoSession, err := w.pool.GetSession()
```
ngjaying: **[bug]** Table connect builds the pool with `PoolConfig.Database` set, then calls `GetSession()` to run `CREATE DATABASE IF NOT EXISTS`. Official IoTDB TableSessionPool examples create databases on a pool that does **not** specify `database`, then open a second pool with the name. The Go client sends `db` on `OpenSession`; docs say to create the database before constructing a pool with `Database` set. A missing catalog can make `GetSession` fail, so the advertised auto-create never runs.

**Suggestion:** Open a pool without `Database`, execute `CREATE DATABASE IF NOT EXISTS`, then either `USE` it or rebuild the pool with `Database` set. Fail `Connect` if CREATE fails (`IF NOT EXISTS` already covers "already exists").

extensions/impl/iotdb/table_writer.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+		if err := fillTablet(tablet, chunk, w.conf, &w.lastAuto); err != nil {
+			return err
+		}
+
+		session, err := w.pool.GetSession()
+		if err != nil {
+			return fmt.Errorf("failed to get iotdb table session: %w", err)
+		}
+		if err := session.Insert(tablet); err != nil {
```
ngjaying: **[bug]** The tree writer passes `sorted=false` so the client sorts timestamps before `InsertTablet` / `InsertAlignedTablet`. The table writer calls `session.Insert(tablet)` with no sort; `insertRelationalTablet` in iotdb-client-go v2.0.8 also does not sort. Unsorted explicit `tsFieldName` values (common with event time) can be rejected or stored incorrectly.

**Suggestion:** Call `tablet.Sort()` after `fillTablet` (or before `Insert`) on the table path so both models send ascending timestamps.

extensions/impl/iotdb/table_writer.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// 自动创建数据库（table dialect 下的 CREATE DATABASE）
+	autoSession, err := w.pool.GetSession()
+	if err != nil {
+		return fmt.Errorf("failed to get iotdb table session for database creation: %w", err)
+	}
+	createDBSQL := fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s", conf.Database)
+	if err := autoSession.ExecuteNonQueryStatement(createDBSQL); err != nil {
+		// 数据库可能已存在，记录 warning 但不阻断连接
+		ctx.GetLogger().Warnf("iotdb table writer: auto-create database %q failed (may already exist): %v", conf.Database, err)
```
ngjaying: **[suggestion]** `CREATE DATABASE IF NOT EXISTS` failures are only logged. Because the statement already includes `IF NOT EXISTS`, a remaining error is not "already exists" (auth, dialect, network). Connect still reports connected and later inserts fail more opaquely.

**Suggestion:** Return the CREATE error from `connect` unless you can detect a genuine already-exists status.

extensions/impl/iotdb/iotdb_test.go
```diff
@@ -0,0 +1,645 @@
...
+	got, err = toMapList(in)
+	assert.NoError(t, err)
+	assert.Equal(t, in, got)
+
+	// unsupported type
+	_, err = toMapList([]byte{1, 2, 3})
+	assert.Error(t, err)
+}
+
+func TestSinkClose(t *testing.T) {
```
ngjaying, line 786: **[suggestion]** Tests cover config validation, conversions, tablet fill, and close-without-connect, but not `collect`/`Connect` with a fake `writer`. The `writer` interface would make it cheap to assert that conversion errors are not IO errors and that empty batches are no-ops.

**Suggestion:** Inject a mock writer and cover collect success, conversion failure, and insert/IO failure paths.

extensions/impl/iotdb/tree_writer.go
```diff
@@ -0,0 +1,133 @@
...
+
+// extractTimestamp returns the timestamp for the given row. If tsFieldName is
+// empty or missing, the current time in milliseconds is returned.
+func extractTimestamp(row map[string]any, tsFieldName string) (int64, error) {
+	if tsFieldName == "" {
+		return time.Now().UnixMilli(), nil
+	}
+	v, ok := row[tsFieldName]
+	if !ok {
+		return time.Now().UnixMilli(), nil
```
ngjaying: **[suggestion]** When `tsFieldName` is set but the field is absent, the sink silently uses wall-clock time. Docs only describe that fallback for an unset `tsFieldName`. Mixed event-time and processing-time timestamps in one batch can overwrite IoTDB rows keyed by timestamp.

**Suggestion:** Error on a missing configured timestamp field, or document the per-row fallback explicitly.

extensions/impl/iotdb/table_writer.go
```diff
@@ -0,0 +1,126 @@
...
+
+	poolConfig, err := conf.newPoolConfig()
+	if err != nil {
+		return err
+	}
+
+	pool := client.NewTableSessionPool(poolConfig, conf.PoolSize, int(conf.Timeout), 60000, false)
+	w.pool = &pool
+
+	// 自动创建数据库（table dialect 下的 CREATE DATABASE）
```
ngjaying: **[suggestion]** Comments at this line and `config.go:149` restate the next statement in Chinese (`自动创建数据库`, `表模型的 database 不需要 root. 前缀，自动去除`). They do not add a non-obvious constraint.

**Suggestion:** Drop them, or keep a short English why (for example, why CREATE is attempted before writes).

ngjaying (commented):
## Summary

This PR adds a predefined Apache IoTDB sink with tree and table writers, plugin registration, docs, and solid config/type-conversion unit tests. The lifecycle, identifier checks for interpolated DDL, IPv6 `addr` parsing, and monotonic auto-timestamps look carefully iterated. Dominant risks are error classification (conversion failures retried as IO), table-model session setup that can prevent the advertised auto-create from running, and unsorted relational tablets.

## Issue counts by severity

- bugs: 3
- suggestions: 4
- nits: 0

extensions/impl/iotdb/iotdb.go
CritasWang: Fixed in 5834d14b. `iotdbSink.collect` now returns writer errors unchanged. The tree/table writers wrap only session-acquisition and insert failures as IO errors, while conversion, schema, and timestamp errors remain ordinary errors. Added `TestSinkCollectErrorClassification` with plain-error and IO-error cases.

extensions/impl/iotdb/tree_writer.go
CritasWang: Fixed in 5834d14b. When `tsFieldName` is configured, a missing field now returns an error; wall-clock timestamps are used only when `tsFieldName` is unset. The JSON descriptor and both English/Chinese docs were updated, with coverage in `TestFillTabletRejectsMissingTimestamp`.

extensions/impl/iotdb/table_writer.go
CritasWang: Addressed in 5834d14b. Removed the information-free Chinese comments and retained concise English comments where they explain the non-obvious pool lifecycle: bootstrapping before selecting a database and returning pooled sessions via `Close`/`PutBack`.

extensions/impl/iotdb/table_writer.go
CritasWang: Fixed in 5834d14b. Errors from `CREATE DATABASE IF NOT EXISTS` are now returned from `connect`; the bootstrap pool is closed and the connected status is not reported when the statement fails. `TestEnsureDatabase` covers the propagated DDL error.

extensions/impl/iotdb/iotdb_test.go
CritasWang: Addressed in 5834d14b. Added fake-writer tests for `Connect` success/failure and `collect` success, empty batches, ordinary conversion errors, and IO errors. Added fake table-session tests for database creation and relational insert error classification as well.

extensions/impl/iotdb/table_writer.go
CritasWang: Fixed in 5834d14b. Table connect now creates a bootstrap `TableSessionPool` with an empty `Database`, executes `CREATE DATABASE IF NOT EXISTS`, closes that pool, and then builds the pool with the target database before testing a session. `TestEnsureDatabase` covers the bootstrap execution and session failure paths.

extensions/impl/iotdb/table_writer.go
CritasWang: Fixed in 5834d14b. The table writer now calls `tablet.Sort()` after `fillTablet` and before `Insert`, keeping timestamps and column values aligned in ascending timestamp order. `TestTableWriterSortsTablet` verifies the reordered timestamps and values.

ngjaying (commented):
## Incremental review of 5834d14

Previous findings look actually fixed. No new issues on this follow-up.

- **IO vs conversion errors:** `collect` now returns writer errors unchanged. Tree/table wrap only session acquire and insert.
- **Table auto-create:** bootstrap pool with empty `Database`, `CREATE DATABASE IF NOT EXISTS` errors fail connect, then a second pool is opened with the target database.
- **Tablet sort:** `tablet.Sort()` runs after `fillTablet`; `TestTableWriterSortsTablet` checks timestamps and values.
- **Missing `tsFieldName`:** configured-but-absent field is an error; docs and `iotdb.json` match.
- **Tests:** fake writer/session coverage for connect, collect classification, DDL, insert, and sort.
- **Comments:** restating Chinese comments removed.

CritasWang:
Thanks for the incremental review. I rechecked the current PR head `5834d14b39e7fc768c1ac59e41d68faebed1af07` against each item in this review, and the implementation matches the stated fixes. There are no new code issues to address, so no additional code changes or empty commit are needed.

Local verification passes:
- `go test ./extensions/impl/iotdb ./extensions/sinks/iotdb ./internal/topo/planner`
- `go test -race ./extensions/impl/iotdb`
- `go vet ./extensions/impl/iotdb ./extensions/sinks/iotdb ./internal/topo/planner`

The PR head remains unchanged at `5834d14b`; the only currently failing check is the Codecov patch-coverage check.

Yisaer (approved):

PR #4112, opened 2026-08-31
feat: support acc_max_by and acc_map_agg

Yisaer:
### Background

  Support calculating the maximum battery temperature and its collection timestamp for each SOC range during a vehicle charging cycle, then reporting all SOC statistics
  when the vehicle powers off.

  ### Changes

  - Add the analytic function acc_max_by(value, compare_value)
      - Returns the value associated with the maximum compare_value.
      - Uses the latest value when compare_value is equal.

  - Add the analytic function acc_map_agg(key, value)
      - Accumulates values by key.
      - Updates the value when a key already exists.
      - Preserves the first-seen key order.
      - Returns an array of objects containing key and value.

  - Add Chinese and English documentation.
  - Add unit tests covering:
      - Maximum-value comparison in acc_max_by.
      - Updating the value when comparison values are equal.
      - Adding new keys in acc_map_agg.
      - Updating values for duplicate keys.

  - Add an FVT test covering the complete upstream/downstream flow:

  simulator
    → Rule 1: acc_max / acc_max_by
    → memory sink
    → memory stream
    → Rule 2: acc_map_agg + WHERE upload_flag = 1
    → memory sink


ngjaying:
Points Worth Flagging

1. Performance: accMapAggExec does a linear scan (for i := range s.Entries) to check whether a key already exists — O(n) — and every call also does a full PutState serialization write-back to the state store, so as Entries grows this becomes roughly O(n²) overall, with every single event writing the entire, ever-growing array to state (especially costly with disk-backed QoS=1). This could be a performance bottleneck on high-frequency streams. Using a map[string]int as a key→index side index would be more robust.
2. Minor code smell: accMaxByAccumulator.accFuncExec is a no-op method, with a comment literally saying "This method is not used" — it exists only to satisfy the accFunc interface so the code can reuse the generic extractAccArgs entry point. This hurts readability a bit; worth raising in review, though not a blocker.
3. Semantic consistency: acc_max returns float64(0) when there's no value, while acc_max_by returns nil in the same situation — this inconsistency has a real cost for users and should be called out explicitly in the docs, to avoid downstream code mistakenly treating nil as 0.

docs/en_US/sqls/functions/object_functions.md
```diff
@@ -121,6 +121,26 @@ example:
 object_construct("a", 1, "b", 2)
 ```
 
+## MAP_AGG_TO_ARRAY
```
ngjaying: **[bug]** `MAP_AGG_TO_ARRAY` was inserted between the `object_construct("a", 1, "b", 2)` example and its `result: {"a":1, "b":2}` block. `OBJECT_CONSTRUCT` now has no result, and `MAP_AGG_TO_ARRAY` appears to return `{"a":1, "b":2}`.

**Suggestion:** Keep the `object_construct` result with that function, and add `MAP_AGG_TO_ARRAY` as a complete section after it (with its own example/result).

docs/zh_CN/sqls/functions/object_functions.md
```diff
@@ -105,6 +105,26 @@ object_construct(key1, col, ...)
 object_construct("a", 1, "b", 2)
 ```
 
+## MAP_AGG_TO_ARRAY
```
ngjaying: **[bug]** Same insertion error in the Chinese object-functions page: `得到如下结果: {"a":1, "b":2}` now sits under `MAP_AGG_TO_ARRAY` instead of `OBJECT_CONSTRUCT`.

**Suggestion:** Restore the `object_construct` result to that section and give `MAP_AGG_TO_ARRAY` its own result example.

internal/binder/function/funcs_acc.go
```diff
@@ -141,6 +144,329 @@ func registerGlobalAggFunc() {
...
+	if !ok {
+		return nil, fmt.Errorf("valid data should be boolean")
+	}
+	v, err := ctx.GetState(key)
+	if err != nil {
+		return nil, err
+	}
+	status := &accStatus{}
+	if v != nil {
+		status = v.(*accStatus)
```
ngjaying: **[suggestion]** `handleAccMapAgg` duplicates `extractAccStatus` but uses `v.(*accStatus)` without a type check and never clears `status.Err`. `extractAccStatus` was just made safe for a mismatched stored type; this path still panics if checkpoint/restore or a reused state key yields another concrete type, and `accMapAggExec` later panics again if `status.Value` is not `*accMapAggStatus`.

**Suggestion:** Call `extractAccStatus`, then initialize/reset `*accMapAggStatus` when `status.Value` is nil or the wrong type, same as `handleAccMaxBy`.

internal/binder/function/funcs_obj.go
```diff
@@ -316,6 +353,52 @@ func registerObjectFunc() {
...
+		if !ok {
+			return nil, fmt.Errorf("each array item value should be an object")
+		}
+		converted := map[string]interface{}{keyName: keyValue}
+		for source, fieldValue := range value {
+			target := source
+			if renamed, ok := rename[source]; ok {
+				target = renamed
+			}
+			converted[target] = fieldValue
```
ngjaying: **[suggestion]** `mapAggToArray` writes the converted key first, then copies every value field (after rename) on top of the same map. A value field named `keyName`, or a rename whose target is `keyName`, silently overwrites the key. Map iteration order also makes two sources renamed to the same target non-deterministic.

**Suggestion:** Apply renames first, then set `converted[keyName] = keyValue` last (or skip/error on collisions with `keyName`).

internal/binder/function/funcs_acc.go
```diff
@@ -141,6 +144,329 @@ func registerGlobalAggFunc() {
...
+		for _, i := range currentIndex {
+			if _, ok := seen[i]; !ok {
+				current.Index = append(current.Index, i)
+			}
+		}
+		sort.Ints(current.Index)
+	}
+}
+
+// accMaxByStatus keeps the value x associated with the greatest y seen so far.
```
ngjaying: **[suggestion]** The comment restates the struct (`Value` plus `By`) using internal `x`/`y` names instead of explaining a non-obvious constraint.

**Suggestion:** Drop it, or replace with a short why (for example that equal `By` keeps the latest `Value`).

internal/binder/function/funcs_acc.go
```diff
@@ -141,6 +144,329 @@ func registerGlobalAggFunc() {
...
+	return status, nil
+}
+
+func accMaxByExec(ctx api.FunctionContext, value, by interface{}, valid bool, key string, status *accStatus, skip bool) {
+	if !valid || by == nil {
+		return
+	}
+	b, err := cast.ToFloat64(by, cast.CONVERT_SAMEKIND)
+	if err != nil {
+		status.Err = fmt.Errorf("acc_max_by y should be number: %w", err)
```
ngjaying: **[nit]** The error string says `acc_max_by y should be number`, leaking the internal `y` name. Callers and docs use `compare_value`.

**Suggestion:** Use `acc_max_by compare_value should be number`.

ngjaying (changes requested):

ngjaying (approved):

PR #4128, opened 2026-09-09
fix(sql): retry named connections after failure

Yisaer:
## Problem

  When the database referenced by a SQL `connectionSelector` is temporarily unreachable during rule startup, the initial SQL connection attempt returns the raw database driver
  error.

  The connection pool only retries errors marked as `errorx.IOErr`, so the failure is treated as permanent. The failed result is then cached in the `ConnWrapper`, causing subsequent
  requests using the same connection ID to return the same error. Creating a new connection is currently required to recover.

  ## Fix

  - Mark SQL connection and ping failures as retryable `errorx.IOErr` errors.
  - Close failed `sql.DB` instances instead of retaining them.
  - Improve nil safety in SQL `Reconnect()` and `Close()`.
  - Update the existing network error test to verify retryable error handling.
  - Add an embedded MySQL integration test to verify that the same named connection ID automatically reconnects after the database becomes available.

  ## Tests

  ```bash
  go test ./extensions/impl/sql/...

  go test -tags=mysql_integration_test ./extensions/impl/sql \
    -run 'TestSQLConnectionNetworkErrorIsRetryable|TestSQLNamedConnectionReconnectAfterStartupFailure'

  The integration test verifies that:

  1. A named SQL connection is created while the database is unavailable.
  2. The initial connection attempt fails.
  3. The embedded MySQL server is started.
  4. The original connection automatically reconnects without creating a new connection.


ngjaying (changes requested):
P1: Avoid leaving `s.db` nil after a failed runtime reconnect.

`Reconnect` now closes the current DB and sets `s.db = nil` before dialing a replacement. If that dial fails, the shared `SQLConnection` exposes a nil DB. SQL source, sink, and lookup callers dereference `GetDB()` without a nil check, so another rule sharing the named connection whose local `needReconnect` is still false can panic in `GetDB().Query/Exec`.

The previous implementation kept a non-nil (although closed) DB handle, so concurrent callers received an error instead of panicking. Please preserve a safe non-nil handle until replacement succeeds, or change DB access to return an error and update all callers. A regression test with two consumers sharing the connection would be useful.

ngjaying (approved):

PR #4135, opened 2026-09-10
fix(conn): clean up when removing connections

### LINKED ISSUE #4130: Cancel background retries when deleting a named connection
When a named connection is deleted or updated while its initial connection retry is still running, `dropNameConnection` removes it from the pool but does not cancel the retry goroutine because the `ConnWrapper` is not initialized yet.

The retry may later succeed and leave an orphan connection that can no longer be managed by its connection ID.

Deleting or updating a named connection should cancel any in-progress retries and release the associated resources.

Hanabi9248:
Fixes #4130.

Deleting or updating a named connection while its initial Dial is retrying removes it from the pool but leaves the retry running. A later successful Dial then creates an orphan connection.

Give each wrapper a cancellable context and cancel it on removal. The backoff now observes cancellation, and cleanup closes the connection even when an in-flight Dial finishes after removal. Cancellation is returned as an error; the parent context and replacement connection remain usable.

Tests cover deletion and replacement during retries and during a Dial that succeeds after removal. The original two deletion cases fail before the fix. With failpoints enabled, `go test -race ./pkg/connection -count=1` passes; the four regression cases also pass with `-race -count=20`. `go vet ./pkg/connection` passes. Tested on Windows with Go 1.27.1; full-repository lint and Linux FVT were not run.

### DIFF
--- a/pkg/connection/conn.go
+++ b/pkg/connection/conn.go
@@ -38,24 +38,29 @@ type ConnWrapper struct {
 	cancel      stdContext.CancelFunc
 }
 
-// close cancels pending retries and also closes a connection whose Dial finishes
-// after it has been removed from the pool.
-func (cw *ConnWrapper) close(ctx api.StreamContext) {
+// cancelAndClose cancels pending retries and closes the connection once Dial returns.
+// Dial must honor context cancellation for prompt cleanup; otherwise cleanup waits
+// for the in-flight call to finish.
+func (cw *ConnWrapper) cancelAndClose(ctx api.StreamContext) {
 	cw.cancel()
 	cleanup := func() {
-		<-cw.readCh
 		cw.l.RLock()
 		conn := cw.conn
 		cw.l.RUnlock()
 		if conn != nil {
-			conn.Close(ctx)
+			if err := conn.Close(ctx); err != nil {
+				ctx.GetLogger().Warnf("failed to close connection %s: %v", cw.ID, err)
+			}
 		}
 	}
 	select {
 	case <-cw.readCh:
 		cleanup()
 	default:
-		go cleanup()
+		go func() {
+			<-cw.readCh
+			cleanup()
+		}()
 	}
 }
 
--- a/pkg/connection/conn.go
+++ b/pkg/connection/conn.go
@@ -15,6 +15,7 @@
 package connection
 
 import (
+	stdContext "context"
 	"sync"
 	"sync/atomic"
 
@@ -34,6 +35,28 @@ type ConnWrapper struct {
 	l           syncx.RWMutex
 	readCh      chan struct{}
 	detachCh    chan struct{}
+	cancel      stdContext.CancelFunc
+}
+
+// close cancels pending retries and also closes a connection whose Dial finishes
+// after it has been removed from the pool.
+func (cw *ConnWrapper) close(ctx api.StreamContext) {
+	cw.cancel()
+	cleanup := func() {
+		<-cw.readCh
+		cw.l.RLock()
+		conn := cw.conn
+		cw.l.RUnlock()
+		if conn != nil {
+			conn.Close(ctx)
+		}
+	}
+	select {
+	case <-cw.readCh:
+		cleanup()
+	default:
+		go cleanup()
+	}
 }
 
 func (cw *ConnWrapper) setConn(conn modules.Connection, err error) {
@@ -63,10 +86,12 @@ func (cw *ConnWrapper) IsInitialized() bool {
 }
 
 func newConnWrapper(ctx api.StreamContext, meta *Meta) *ConnWrapper {
+	ctx, cancel := ctx.WithCancel()
 	cw := &ConnWrapper{
 		ID:       meta.ID,
 		readCh:   make(chan struct{}),
 		detachCh: make(chan struct{}),
+		cancel:   cancel,
 	}
 	go func() {
 		conn, err := createConnection(ctx, meta)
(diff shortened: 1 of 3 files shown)


pkg/connection/conn.go
```diff
@@ -34,6 +35,28 @@ type ConnWrapper struct {
 	l           syncx.RWMutex
 	readCh      chan struct{}
 	detachCh    chan struct{}
+	cancel      stdContext.CancelFunc
+}
+
+// close cancels pending retries and also closes a connection whose Dial finishes
+// after it has been removed from the pool.
+func (cw *ConnWrapper) close(ctx api.StreamContext) {
```
ngjaying, line 43: Suggestion: rename `close` — it shadows the builtin `close` and is easy to misread (`cw.close` vs `close(ch)`). `go vet` passes but readability suffers. `cancelAndClose` / `terminate` would better convey the \"cancel first, then wait for readCh, then Close\" semantics.

pkg/connection/conn.go
```diff
@@ -34,6 +35,28 @@ type ConnWrapper struct {
...
+	cw.cancel()
+	cleanup := func() {
+		<-cw.readCh
+		cw.l.RLock()
+		conn := cw.conn
+		cw.l.RUnlock()
+		if conn != nil {
+			conn.Close(ctx)
```
ngjaying, line 51: Two nits: 1) The outer `select { case <-readCh: cleanup() }` is redundant since `cleanup()` waits on `<-readCh` again — harmless because reading a closed channel is non-blocking, but a bit convoluted. Consider simplifying to \"close synchronously if ready, otherwise wait in a goroutine\". 2) The `conn.Close(ctx)` return value is ignored (as in the old code); consider at least logging it to help diagnose leaks after deletion.

pkg/connection/pool.go
```diff
@@ -263,12 +263,7 @@ func dropNameConnection(ctx api.StreamContext, selId string) error {
 	if err != nil {
 		return fmt.Errorf("drop connection %s failed, err:%v", selId, err)
 	}
-	if meta.cw.IsInitialized() {
-		conn, err := meta.cw.Wait(ctx)
-		if conn != nil && err == nil {
-			conn.Close(ctx)
-		}
-	}
+	meta.cw.close(ctx)
```
ngjaying, line 266: The named path is fixed here, but the anonymous path in `detachConnection` (~L348) still uses `close(detachCh)+Wait`: detaching during a retry leaves the `createConnection` goroutine running, and a later successful `Dial` becomes an orphan `conn` no longer in the map. Fine to keep this PR scoped to named connections, but suggest a TODO/comment noting the inconsistency or a follow-up fix.

pkg/connection/pool.go
```diff
@@ -421,7 +416,10 @@ func createConnection(connCtx api.StreamContext, meta *Meta) (modules.Connection
 			return err
 		}
 		return backoff.Permanent(err)
-	}, NewExponentialBackOff())
+	}, backoff.WithContext(NewExponentialBackOff(), connCtx))
```
ngjaying, line 419: `backoff.WithContext` only cancels the sleep interval between retries; it cannot interrupt an in-flight `Dial`. If a driver ignores `ctx.Done()`, the `cleanup` goroutine stays blocked on `<-readCh` until `Dial` returns (the test only passes because it manually closes `release`). Suggest documenting/commenting that `Dial` is expected to respect context cancellation, otherwise delayed release or a brief leak is still possible in production.

ngjaying (commented):
Overall direction LGTM — the approach to fixing the orphan-connection issue in #4130 is sound. Leaving a few inline nits on naming, cleanup redundancy, anonymous-connection consistency, and the Dial-cancellation dependency.

ngjaying (changes requested):

Hanabi9248:
Updated in bc1a644. Renamed the helper to `cancelAndClose`, removed the second channel receive on the synchronous path, and added a warning when `Close` fails. The helper comment now explains that prompt cleanup depends on `Dial` honoring cancellation. I left anonymous connections unchanged and added a TODO at that detach path.

The full connection package passes with failpoints and `-race`; the four removal/replacement regressions also pass with `-race -count=20`. `go vet ./pkg/connection` passes.

ngjaying:
There is still one blocking cleanup issue on the anonymous-connection path.

createConnection now returns (conn, context.Canceled) when its context is canceled. However, detachConnection only closes the connection when conn != nil && err == nil. This means an anonymous connection whose initialization finishes after cancellation can be removed from the pool without calling Close(), leaking the underlying resource.

Please make the anonymous detach path close any non-nil connection regardless of the initialization error, or route anonymous cleanup through the same cancelAndClose lifecycle. A regression test covering cancellation followed by detach would also be useful.

Hanabi9248:
Thanks for catching this, and sorry for the delayed follow-up. Fixed in 3e67344: anonymous detach now uses cancelAndClose, so cleanup waits for initialization and closes a non-nil connection even when initialization returns context.Canceled.

Added cancellation-then-detach regressions for both an in-flight Dial and an already completed initialization. Both fail before the fix. The full connection package passes with failpoints enabled and -race; the named and anonymous regression cases also pass with -race -count=20. go vet passes. These checks ran on Windows; the updated upstream CI is still pending.

ngjaying:
Heads-up on direction: the ConnWrapper lifecycle model this PR patches is being replaced by the Connection Pool ownership redesign in #4177 (Meta-owned lifecycle, explicit removing state, worker join, pool-owned teardown).

The cancellation behavior and late-Dial regression scenarios you identified here (fixing #4130) are carried forward in a stacked follow-up #4179, adapted to the new model: lifecycle-aware initial-Dial backoff plus rewritten named-replacement and anonymous-teardown regression tests. Attribution to your #4135 scenarios is noted in that PR body and in the test file.

Keeping this PR open until the replacement lands; it should be closed as superseded once #4179 merges.

Issue #4095, opened 2026-08-04
[Bug] Graph Redis lookup JOIN followed by pick/filter outputs empty object and panics in JoinTuple.doGetValue


Arisingr:
# [Bug] Graph Redis lookup JOIN followed by pick/filter outputs empty object and panics in JoinTuple.doGetValue

## Environment

- eKuiper version: `v2.4.0-full`
- Deployment: Docker
- OS: Linux
- MQTT source: MQTT
- Lookup source: Redis string
- Rule type: Graph rule

Image details:

```bash
docker inspect ekuiper --format '{{.Config.Image}} {{.Image}}'
docker exec ekuiper /kuiper/bin/kuiper --version
```

## What happened

A Graph rule joins an MQTT stream with a Redis lookup table.

The lookup node receives and emits one record, but the following `pick` node outputs an empty object:

```text
sink result for rule graph_demo_barn_temp: [{}]
```

When a filter reads the projected field, the rule panics and changes to `stopped by error`:

```text
runtime error: index out of range [0] with length 0
github.com/lf-edge/ekuiper/v2/internal/xsql.(*JoinTuple).doGetValue
    internal/xsql/row.go:615
github.com/lf-edge/ekuiper/v2/internal/topo/operator.(*FilterOp).Apply
    internal/topo/operator/filter_operator.go:64
```

## Expected behavior

The lookup result should contain the MQTT fields and the Redis `barn_id` field:

```json
{
  "device_id": "DEV_001",
  "barn_id": "barn_A01",
  "temperature": 32.5,
  "humidity": 65
}
```

A missing field should evaluate to `nil` or produce a normal runtime error. It should not panic and stop the entire rule.

### 1. Create the MQTT stream

```json
{
  "sql": "CREATE STREAM sensorData() WITH (DATASOURCE=\"sensor/test/data\", FORMAT=\"json\", TYPE=\"mqtt\")"
}
```

### 2. Configure the Redis source

Create a Redis source configuration named `demo_redis`:

```json
{
  "addr": "<redis-host>:6379",
  "dataType": "string",
  "password": "<redacted>"
}
```

### 3. Create the Redis lookup table

```json
{
  "sql": "CREATE TABLE deviceBarn() WITH (DATASOURCE=\"0\", FORMAT=\"json\", TYPE=\"redis\", KIND=\"lookup\", CONF_KEY=\"demo_redis\")"
}
```

### 4. Insert lookup data

```bash
[...]

PR #4136, opened 2026-09-10
fix(xsql): guard empty join lookup

Hanabi9248:
A projection that retains only a calculated field empties `JoinTuple.Tuples`. A subsequent unqualified lookup of a discarded field then indexes `tuples[0]` and panics. Return `(nil, false)` when there are no source tuples, while preserving calculated-field lookups.

The regression exercises `Pick` followed by value and metadata lookups, with and without a stream qualifier. It reproduces the panic before the fix. Related to the crash reported in #4095; this does not address the separate Redis lookup/projection result reported there.

Validation on Windows, Go 1.27.1: `go test -race ./internal/xsql -skip '^TestComparison$' -count=1` passes. The full package run fails only `TestComparison` because it expects UTC in formatted errors; the same test fails on the baseline in this machine's CST timezone. `go vet ./internal/xsql` reports the existing copied-lock warning at `row.go:187` on both versions.

### DIFF
--- a/internal/xsql/row.go
+++ b/internal/xsql/row.go
@@ -601,6 +601,9 @@ func (jt *JoinTuple) AddTuples(tuples []Row) {
 
 func (jt *JoinTuple) doGetValue(key, table string, isVal bool) (interface{}, bool) {
 	tuples := jt.Tuples
+	if len(tuples) == 0 {
+		return nil, false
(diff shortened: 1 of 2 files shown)


ngjaying (approved):
LGTM, thanks

PR #4141, opened 2026-09-14
fix(sql): lookup driver capabilities

ngjaying:
## Summary
- Follow-up to #4139: same driver-capability treatment for the lookup path.
- MaxCompute keeps legacy inline literals with zero args (its driver mangles bound args).
- QL maps to numbered $N binds; bare ? is not valid QL.

## Why
- #4139 changed lookup to always pass args with a ? fallback for unmapped drivers, which deterministically breaks MaxCompute (pinned gomaxcompute runs fmt.Sprintf over the query when args are supplied) and leaves QL broken (bare ? is not valid QL syntax).
- Extends the sink-side capability model from #4140 to lookup.

## Changes In This PR
- extensions/impl/sql/lookupSource.go: paramSQLGen gains inline mode (legacy literals, no args) for maxcompute/mc; ql/cznic/cznicql map to dollarBind.
- extensions/impl/sql/lookup_query_test.go: MaxCompute inline snapshots (byte-identical to pre-#4139 output, empty args), QL strict-key rejection, real ql-mem lookup-shaped round-trip.
- Verified: go vet (plain + mysql_integration_test tags), golangci-lint and go test ./extensions/impl/sql/ pass.

## Notes
- Oracle :N binds still verified against driver docs only (carried over from #4139).
- Closes the P1/P2 review findings on #4140 for the lookup path.


ngjaying:
Closing: MaxCompute and QL are not in the supported matrix, so per team decision the driver-specific handling (inline literals for MaxCompute, numbered binds for QL) is reverted rather than shipped. Both drivers intentionally share the generic ? fallback. See #4140 for the final scope.

PR #4142, opened 2026-09-15
fix(sql): reject nil keys in lookup

ngjaying:
## Summary
- Lookup with a nil key now returns an error instead of generating WHERE key IS NULL.

## Why
- Follow-up to #4139: IS NULL on a possibly non-unique nullable key can silently match many rows, while the old code failed loud with broken SQL. The lookup node already skips nil keys before calling, so this guards direct API use.
- Mirrors the same decision in #4140 for sink update/delete keys.

## Changes In This PR
- extensions/impl/sql/lookupSource.go: nil value in buildQuery returns an error.
- extensions/impl/sql/lookup_query_test.go: null tests assert the error; mixed-null snapshot replaced with a plain SELECT * case.
- Verified: go vet (plain + mysql_integration_test tags), golangci-lint and go test ./extensions/impl/sql/ pass.


ngjaying:
Folded into #4140 (cherry-picked as 0daee7135) to avoid a standalone PR for a 16-line change. Closing.

PR #4145, opened 2026-09-17
fix(server): retry init.json persistence errors

ngjaying:
## Summary
- Retry init.json persistence writes up to three times without repeating lookup table provisioning.
- Leave the initialized marker unchanged after retryable startup errors, so the next startup can retry the same init.json.
- Skip definitions blocked by version precedence, and replace corrupt stored definitions from valid init.json content.

## Why
- SF-1024: a transient Stream, Table, or Rule storage error can leave initialization incomplete. Unlike an API request, startup initialization has no caller to retry it.
- A lookup table can connect to an external source before its definition is saved; retrying its entire create operation would repeat that side effect.

## Changes In This PR
- Add structured Applied, Skipped, TerminalError, and RetryableError outcomes for startup initialization.
- Share Stream/Table parsing and the Stream/Table/Rule replace decision with the existing API paths. Version precedence is checked before SHARED compatibility.
- Distinguish storage read errors from readable but corrupt old definitions; the latter can be overwritten by a validated init.json definition.
- Retry only database persistence writes up to three immediate attempts. Unknown runtime errors block marker creation without immediately repeating the operation.
- Preserve the existing marker until the new one is created; cover marker recovery, version skips, corrupt Stream/Table/Rule records, and lookup table write failures with tests.
- Keep REST/RPC import behavior unchanged.

## Notes
- Verified with a real kuiperd process and injected SQLite write failure: three write attempts, old marker retained, then successful initialization on the next startup without changing init.json.
- Targeted tests, the existing TestInitRuleSet in an isolated data directory, processor tests, go vet, golangci-lint, and build passed locally.


ngjaying:
Superseded by #4151. The replacement starts from master and uses a simpler success/version-skip/failure model for init.json.

PR #4155, opened 2026-09-18
fix: bound initial connection retries

Yisaer:
- Bound initial connection retries to 10 seconds.
  - Preserve the configured backoffMaxElapsedDuration for general backoff usage, with a 10-second code fallback.


ngjaying (changes requested):
## Summary

Request changes. This PR partially mitigates Issue #1236 by bounding retry intervals, but it does not yet guarantee that a failed SQL lookup connection releases the lookup-table management path.

## Findings

### P1 - The retry limit does not bound a blocking Dial/Ping

`createConnection` now gives `backoff.Retry` a 10-second `MaxElapsedTime`, but the timeout is only checked between retry attempts. The SQL implementation still calls `database/sql.DB.Ping()` from `Dial()`, not `PingContext()`. If the endpoint is a black-hole/unresponsive address or the driver blocks inside a single dial, `backoff.Retry` never regains control and the lookup lock can still remain held indefinitely.

Please add a driver-level/context-aware connection timeout (for example `PingContext` plus a per-attempt deadline) and add a test with a blocking Dial implementation.

### P1 - The failed lookup connection is not cleaned up

After the new retry timeout, `SqlLookupSource.Connect()` can return with `cw.Wait()` reporting no connection. At that point the source has already called `FetchConnection()` and incremented the reference count, but the error path does not detach that reference or close the partially initialized source. `lookup.CreateInstance()` returns before the source is stored, so its normal `Close()` path is not reached.

This can leave the bad named connection with a permanent reference, make later update/drop operations fail, and cause later lookup tables using the same selector to reuse a failed wrapper. Please define and test an explicit failure-cleanup path; ensure the detach reference ID matches the ID used by `FetchConnection()`.

### P1 - The change bypasses the configured retry duration

The production call in `createConnection()` now always uses `InitialConnectionMaxElapsedTime = 10s`, while the existing `connection.backoffMaxElapsedDuration` configuration defaults to 3 minutes. This changes the behavior for deployments whose database takes more than 10 seconds to become ready, despite the PR description saying the configured duration is preserved for general backoff usage. Please clarify the intended semantics and either make the initial timeout configurable or document/test this breaking behavior.

### P2 - The global lookup lock still covers external network I/O

`lookup.CreateInstance()` still holds the global lookup lock while running `Provision()` and `Connect()`. The PR changes an unbounded block into a block of up to roughly 10 seconds for fast-failing connections, but all lookup create/drop/attach operations can still be serialized behind a failed external connection. A follow-up or this PR should move network I/O outside the global lock, or introduce an explicit loading state.

### P2 - Missing regression coverage

Please add a regression test based on Issue #1236: a failed SQL lookup connection must return, release lookup state, allow a subsequent valid lookup table to be created, and leave no stale connection reference. Also test an actually blocking Dial/Ping path.

## Log/endpoint note

The Issue #1236 log shows `sqlwrong` being attached at 20:35:11 with ref count 1, followed by no `Table sqlwrong is created` or error. This strongly indicates the create path is waiting for the connection. However, the current eKuiper `GET /tables?kind=lookup` handler reads table metadata and does not directly acquire the lookup global lock or connect to SQL. Therefore this PR should not be considered a direct proof that the GET endpoint is fixed; please include an API-level reproduction or goroutine dump, especially for the NeuronEX `/api/ekuiper/...` path.

The separate initial-connection backoff is a reasonable direction, but the above P1 issues need to be addressed before merge.

Yisaer:
> ## Summary
> Request changes. This PR partially mitigates Issue #1236 by bounding retry intervals, but it does not yet guarantee that a failed SQL lookup connection releases the lookup-table management path.
> 
> ## Findings
> ### P1 - The retry limit does not bound a blocking Dial/Ping
> `createConnection` now gives `backoff.Retry` a 10-second `MaxElapsedTime`, but the timeout is only checked between retry attempts. The SQL implementation still calls `database/sql.DB.Ping()` from `Dial()`, not `PingContext()`. If the endpoint is a black-hole/unresponsive address or the driver blocks inside a single dial, `backoff.Retry` never regains control and the lookup lock can still remain held indefinitely.
> 
> Please add a driver-level/context-aware connection timeout (for example `PingContext` plus a per-attempt deadline) and add a test with a blocking Dial implementation.
> 
> ### P1 - The failed lookup connection is not cleaned up
> After the new retry timeout, `SqlLookupSource.Connect()` can return with `cw.Wait()` reporting no connection. At that point the source has already called `FetchConnection()` and incremented the reference count, but the error path does not detach that reference or close the partially initialized source. `lookup.CreateInstance()` returns before the source is stored, so its normal `Close()` path is not reached.
> 
> This can leave the bad named connection with a permanent reference, make later update/drop operations fail, and cause later lookup tables using the same selector to reuse a failed wrapper. Please define and test an explicit failure-cleanup path; ensure the detach reference ID matches the ID used by `FetchConnection()`.
> 
> ### P1 - The change bypasses the configured retry duration
> The production call in `createConnection()` now always uses `InitialConnectionMaxElapsedTime = 10s`, while the existing `connection.backoffMaxElapsedDuration` configuration defaults to 3 minutes. This changes the behavior for deployments whose database takes more than 10 seconds to become ready, despite the PR description saying the configured duration is preserved for general backoff usage. Please clarify the intended semantics and either make the initial timeout configurable or document/test this breaking behavior.
> 
> ### P2 - The global lookup lock still covers external network I/O
> `lookup.CreateInstance()` still holds the global lookup lock while running `Provision()` and `Connect()`. The PR changes an unbounded block into a block of up to roughly 10 seconds for fast-failing connections, but all lookup create/drop/attach operations can still be serialized behind a failed external connection. A follow-up or this PR should move network I/O outside the global lock, or introduce an explicit loading state.
> 
> ### P2 - Missing regression coverage
> Please add a regression test based on Issue #1236: a failed SQL lookup connection must return, release lookup state, allow a subsequent valid lookup table to be created, and leave no stale connection reference. Also test an actually blocking Dial/Ping path.
> 
> ## Log/endpoint note
> The Issue #1236 log shows `sqlwrong` being attached at 20:35:11 with ref count 1, followed by no `Table sqlwrong is created` or error. This strongly indicates the create path is waiting for the connection. However, the current eKuiper `GET /tables?kind=lookup` handler reads table metadata and does not directly acquire the lookup global lock or connect to SQL. Therefore this PR should not be considered a direct proof that the GET endpoint is fixed; please include an API-level reproduction or goroutine dump, especially for the NeuronEX `/api/ekuiper/...` path.
> 
> The separate initial-connection backoff is a reasonable direction, but the above P1 issues need to be addressed before merge.

The issues have been addressed as follows:

  1. Fixed the SQL connection dial/ping timeout handling.
  2. Fixed cleanup of failed lookup connections and ensured the correct reference ID is detached.
  3. The 10-second initial connection timeout is intentional for connection creation.
  4. Holding the lookup lock during connection setup is intentional to prevent concurrent initialization.
  5. Added a unit test covering failed connection cleanup and subsequent successful connection creation.

ngjaying (approved):

PR #4160, opened 2026-09-18
fix(conf): honor schema types for env overrides

Dream95:
## Summary
Env vars overlaid onto YAML configs were parsed by inference. That meant password=123456 could be inferred as an int64 and fail during decode at server/rest.go:92:

~~~log
level=error msg="1 error(s) decoding:\n\n* 'password' expected type 'string', got unconvertible type 'int64', value: '123456'" file="server/rest.go:92"
~~~

Values now use types from the companion JSON schema when available.

This keeps string fields like password, username, and clientid as strings even when the env value looks numeric.

If no schema type is available, fall back to the previous inference behavior.

## Test plan
- [x] Unit tests for getValueType, covering schema types and inference fallback
- [x] MQTT source env overlay keeps password/username/clientid as strings


ngjaying:
Hi @Dream95, thanks for the fix — the reported failure is real, and your diagnosis is spot on: inferring types from value shapes is what breaks numeric-looking passwords.

After reviewing, we've converged on a plan that fixes this at a different layer, and we'd like to share it before deciding how to proceed.

**The architectural point**: when `cast.MapToStruct` (`pkg/cast/cast.go:1134`) decodes the config map, the decoder already knows the target field types — that's the same machinery producing this error. Type conversion should live there and only there. Any overlay-level type resolution (inference or schema lookup) re-answers a question the decoder already holds the answer to, and it can't know the target type when a key is absent from the YAML map.

**Proposed plan**:

1. `cast.MapToStruct` / `MapToStructStrict` gain a bidirectional scalar `DecodeHook`:
   - target `string` + source int/uint/float/bool → `strconv.FormatInt`/`FormatUint`/`FormatFloat('f', -1)`/`FormatBool`
   - target int/uint/float/bool + source string → `ParseInt(..., 10, bits)`/`ParseUint`/`ParseFloat`/`ParseBool`; convert only on success, keep current errors otherwise (empty string stays an error — avoids mapstructure's weak-mode `"" → 0` silent conversion; base 10 avoids the `ParseInt(str, 0, ...)` octal surprise on `"010"`)
2. Env overlay (`process`/`handle`) passes env values through as strings; `getValueType` inference is removed.
3. Net effect: same value ⇒ same result regardless of source (yaml/json/env/API). Notably this also fixes `password: 123456` written unquoted directly in a user yaml conf — today that fails even without env overrides.

**How to proceed**: we're treating this as the real fix rather than an interim patch, so the path forward for this PR is to rework it into the decode-layer hook + overlay pass-through:

- scope is small: ~40 lines in `pkg/cast` + a source×target test grid + deleting `getValueType`
- your `TestMqttNumericPasswordFromEnv` stays as the regression test
- we're happy to review quickly and help with edge cases (bit sizes, empty string, duration hook ordering)

If you'd rather we take the cast-layer part, that's fine too — we'd implement it and credit your analysis in the follow-up; this PR would then be superseded.

Either way, thanks for surfacing this.

PR #4162, opened 2026-09-20
fix(sql): bound lookup connect wait

ngjaying:
## Summary
- Bound SQL lookup table creation wait on the pooled connection to 10s.
- Release the connection reference when the bounded wait fails.
- Fix connection detach to use the original FetchConnection refId.

## Why
- Lookup table creation holds the global lookup lock while calling `ConnWrapper.Wait` with a background context that never cancels.
- With unbounded pool retries, an unreachable database blocks `CreateInstance` forever and hangs all lookup DDL/list/attach paths.
- Bounding the wait at the caller keeps the pooled connection retrying in the background (persistent connection principle) while failing table creation fast.

## Changes In This PR
- `extensions/impl/sql/lookupSource.go`: 10s bounded wait via `WithCancel` + `AfterFunc`; store `refId` and detach by ref on wait failure and in `Close`.
- `pkg/connection/pool.go`: add `DetachConnectionByRef`; `detachConnection` takes the refId explicitly instead of deriving it from the caller context.
- `extensions/impl/sql/client/client.go`: bound single dial ping to 2s with `PingContext` so a blackholed dial cannot outlive the retry budget per attempt.
- `pkg/connection/pool_test.go`: update internal detach calls to pass explicit refIds.

## Notes
- Stacked on #4146 branch (`revise_sql_conn`); only the 4 files above differ from that base.
- `TestConnectionErr` in `pkg/connection` fails on the base too (pre-existing failpoint path issue, unrelated).
- Verified: `go build`, `go vet`, `pkg/connection` tests (except pre-existing failure), `internal/topo/lookup` tests.


ngjaying:
Superseded: the fix was pushed directly into #4161 as requested.

PR #4163, opened 2026-09-20
feat: distinct acc and array positions function

Yisaer:
## Summary

  - Add `acc_distinct_collect` to collect unique non-nil values while preserving insertion order.
  - Add `array_positions(array, value)` to return all matching 0-based indexes.
  - Reuse the existing value comparison logic.
  - Add unit tests for duplicates, missing values, nil inputs, and argument validation.
  - Update English and Chinese documentation.

  ## Examples

  ```text
  acc_distinct_collect(a)
  -- Input: 1, 2, 2, 3
  -- Output: [1, 2, 3]

  array_positions([1, 2, 2, 3, 2], 2)
  -- Output: [1, 2, 4]


ngjaying (changes requested):
A couple of concerns about the current `acc_distinct_collect` implementation:

* **Performance:** this is O(n²) when most accumulated values are unique, because every new value scans the full collected slice with `reflect.DeepEqual`. Since an accumulator may keep growing over a long-running stream, this could become expensive for larger state sizes. Could we keep a membership index alongside the ordered values? For comparable scalar values, a `map[interface{}]struct{}` could provide O(1) membership checks, while non-comparable values such as arrays/maps could fall back to `reflect.DeepEqual`. This would preserve insertion order without repeatedly scanning the entire accumulated result.

* **Consistency with `array_distinct`:** the equality semantics currently differ for composite values. `acc_distinct_collect` uses `reflect.DeepEqual`, so equal arrays/maps are deduplicated, while the existing `array_distinct` intentionally does not deduplicate non-hashable values such as arrays and maps. It would be good to either align the semantics, ideally by sharing a helper, or explicitly document the difference. In either case, please add tests for duplicate arrays/maps so the intended behavior is clear and remains stable.

ngjaying (changes requested):
One more concern about `DistinctIndex`: since it is an exported field of `accStatus`, it will be persisted as part of the checkpoint state by gob.

The index is only an optimization and is fully derivable from `Value` (and the current code already rebuilds it when it is nil), so persisting it means every distinct scalar is effectively stored twice: once in the collected values and once as a key in `DistinctIndex`. For a long-running accumulator, this can unnecessarily increase checkpoint size and serialization cost.

Could we make the index an in-memory-only cache instead, e.g. rename it to an unexported `distinctIndex` field? Gob will then ignore it, and after state restoration the existing lazy rebuild logic can recreate it from `Value`. The newly added rehydration test already covers this behavior nicely.

ngjaying (approved):

PR #4167, opened 2026-09-21
fix(rule): avoid nil panic on concurrent stop

ngjaying:
## Summary
- Fix nil-pointer panic in rule cleanRule when topo auto-exit races explicit stop.
- Serialize ownership check and topo stop in cleanRule under one ruleLock critical section.
- Add regression test pinning stale-cleanup safety.

## Why
- Flaky CI failure (internal/server TestRule): Topo.GetRunId() panicked at topo.go:218 via cleanRule:153 when doStop cleared s.topology concurrently.
- Deeper hazard: Topo.doClose closes shared MergeableTopo sources even after a failed CAS, and CloseSubTopo detaches schema state unconditionally, so a stale run stopping its old topo could detach schema belonging to a replacement run under the same rule ID.
- Keep doStart/doStop lock-free by design (caller holds ruleLock); only the async path changed.

## Changes In This PR
- internal/topo/rule/action.go only: runTopo delegates stop+cleanup to cleanRule; cleanRule checks ownership and stops the topo atomically under write lock, with nil-safe mismatch logging (Debug for already-stopped normal path, Warn for genuine mismatch).
- internal/topo/rule/state_test.go: TestCleanRuleStaleTopo covers replaced-run and stopped-rule stale cleanup plus rapid restarts (no panic, active run undisturbed).

## Notes
- Holding ruleLock across GracefulStop matches the existing doStop path (StopWithLastWill already stops while holding the lock); no new lock ordering is introduced.
- Verified: go build ./internal/..., go test -race ./internal/topo/rule/ pass.


Yisaer (approved):

ngjaying:
Reply to #pullrequestreview-5262861760 (Copilot: test does not deterministically exercise the interleaving) — acknowledged as stated, but not actionable after this fix, and here's why:

The dangerous interleaving lived in the seam *between* the ownership check and the stop. This PR removes that seam: check + stop now happen atomically under one ruleLock critical section in cleanRule. There is no longer a boundary where a deterministic test could park a goroutine, so a deterministic repro of the pre-fix interleaving is structurally out of reach without test-only hooks in the production concurrency path, which we deliberately avoid.

What the test does pin deterministically:
- stale cleanRule (replaced run and stopped rule) is a no-op, never panics, never detaches the active run — this is the exact CI panic line;
- rapid Stop/Start restarts under -race, so any data race in the async cleanup path fails loudly.

One related observation from validating this: asserting Topo IsClosed after an owned cleanup is inherently timing-dependent (cleanRule may run before runTopo calls Open, leaving the topo Initialized by design), which further confirms lifecycle-timing assertions are the wrong tool here. Contract + race detector is the right one.

PR #4169, opened 2026-09-21
feat(conn): explicit connection identity

ngjaying:
## Summary
- Separate connectionKey (which logical connection is shared) from refId (which consumer holds it) in the connection pool.
- Single source of truth for refs (`refMu + map`, `len(refs)`); duplicate refId no longer double-counts, missing DeRef is a no-op.
- New explicit `FetchConnectionWithOptions` API; legacy `FetchConnection` kept as compatibility shim (same connectionKey derivation, stored ref normalized to the consumer identity so legacy Close round-trips). SQL Source/Sink/Lookup migrated; other connectors untouched.
- No runtime reconnect behavior change: reconnect/retry ownership untouched.

## Why
- Source/Sink attached with the DB URL but detached with `rule_op_instance`, causing ref leaks and count drift.
- Two lookup tables sharing one DB shared one refId (map held 1 entry, count showed 2); missing DeRef could drive the count negative.
- `sync.Map + atomic` was a dual source of truth for the same reference set.

## Changes In This PR
- `pkg/connection/conn.go`: ref map single source of truth; `NotifyStatus` snapshots handlers before invoking (invocation stays synchronous).
- `pkg/connection/pool.go`: `FetchOptions{ConnectionKey, RefID, RequireExisting, Type, Props, StatusHandler}`; `fetchInternal` shared by new API + legacy shim; attach-time callback keeps baseline deliver-before-register ordering (serialized subscription is follow-up work); same-key Type conflict rejected on all paths; props cloned on Meta creation.
- `pkg/connection/lookup_ref.go` (new): `WithLookupRefID`/`LookupRefID` helpers taking the real framework context type (no arbitrary-context fallback).
- `extensions/impl/sql`: canonical connectionKey helper (`connkey.go`, non-string selector is a static error); Source/Sink save the consumer refId at Connect and detach verbatim at Close; Lookup requires the framework-injected `lookup:<name>` identity (no DB URL fallback).
- `internal/topo/lookup/table.go`: `CreateInstance` injects `lookup:<name>` into the lookup context.
- Tests: explicit identity / type conflict (incl. failed-fetch never invokes handler) / RequireExisting-missing / legacy shim-detach round-trip / Source+Sink partial-Close release / real dual-lookup shared-DB lifecycle via `CreateInstance`.

## Notes
- One intentional strictness beyond pure compat: same-key Type conflict errors even via the legacy shim (pool invariant validation, not a behavior change; prevents wrong-type assertion downstream).
- `TestConnectionErr` in `pkg/connection` requires a failpoint-instrumented build (pre-existing, fails on a clean tree too).
- Follow-up (stacked): Manager/Meta lifecycle ownership, per-key creation coordination, cancel/done/stopOnce, `Wait` error contract.


ngjaying:
Superseded by #4170: same commits moved to `feat/EK-1238-a1a-identity` in this repo so follow-up work can stack on it. Review history here remains for reference.

